PR #17, opened 2016-03-23
Fetch default dpi from cups

tresf:
**Background:**

Since JVM + Mac suffer a hard crash (https://github.com/qzind/qz-print/issues/155) when the DPI isn't provided.  Like all upstream bugs, this forces us to get creative with a workaround...  

The recommended solution from PDFBOX developers is to provide a density to avoid the hard-crashes, which in turn forces PDFBOX to rasterize the PDF prior to printing.  This works well if we know the printer DPI.... Unfortunately, this DPI comes back as Null on Mac and Linux.

The work-around is to query this value from CUPS. The API we're using in the `lpstat` and `lpoptions` command line utilities.

The advantage of this enhancement is it has other benefits... it allows printer-DPI detection where Java simply didn't provide it before, helping out our Pixel printing logic considerably with the CUPS backends OSs... 

This native work-around uses a shell command to parse attributes from CUPS, namely the default resolution of the printer.

**Exceptions:**

On OS X, this workaround also takes into consideration the delicate renaming problems, as printers identified on Mac default to the CUPS printer `Description:` rather than the `Name:`.  e.g. "List all printers" has a different behavior on Mac versus other platforms as the human readable "Description" is returned by Java whereas Unix/Linux returns the CUPS device name.  This solution works for both scenarios.

**Sample output:**

When this feature is working and a supported printer driver is used, the DPI should show in the console as follows:

Mac:

```properties
[DEBUG] 2016-03-22 22:25:56,442 @ qz.utils.ShellUtilities:?
	Executing: [lpoptions, -p, PDFwriter, -l]
[DEBUG] 2016-03-22 22:25:56,565 @ qz.printer.action.PrintPDF:?
	Parsed density from CUPS: 300 dpi
[INFO] 2016-03-22 22:25:56,583 @ qz.printer.action.PrintPixel:?
	Starting printing (1 copies)
```

Ubuntu:

```properties
[DEBUG] 2016-03-22 22:43:27,433 @ qz.utils.ShellUtilities:?
	Executing: [lpoptions, -p, PDF, -l]
[DEBUG] 2016-03-22 22:43:27,467 @ qz.printer.action.PrintPDF:?
	Parsed density from CUPS: 300 dpi
[INFO] 2016-03-22 22:43:27,523 @ qz.printer.action.PrintPixel:?
	Starting printing (1 copies)
```

### DIFF
--- a/src/qz/printer/action/PrintHTML.java
+++ b/src/qz/printer/action/PrintHTML.java
@@ -19,7 +19,6 @@ import qz.printer.PrintOptions;
 import qz.utils.PrintingUtilities;
 
 import java.awt.print.Printable;
-import java.awt.print.PrinterJob;
 import java.io.IOException;
 
 public class PrintHTML extends PrintImage implements PrintProcessor, Printable {
@@ -36,6 +35,7 @@ public class PrintHTML extends PrintImage implements PrintProcessor, Printable {
     public void parseData(JSONArray printData, PrintOptions options) throws JSONException, UnsupportedOperationException {
         try {
             WebApp.initialize();
+            PrintOptions.Pixel pxlOpts = options.getPixelOptions();
 
             for(int i = 0; i < printData.length(); i++) {
                 JSONObject data = printData.getJSONObject(i);
@@ -43,16 +43,29 @@ public class PrintHTML extends PrintImage implements PrintProcessor, Printable {
 
                 PrintingUtilities.Format format = PrintingUtilities.Format.valueOf(data.optString("format", "FILE").toUpperCase());
 
-                double pageWidth = PrinterJob.getPrinterJob().getPageFormat(null).getWidth();
-                if (!data.isNull("options")) {
-                    pageWidth = data.optJSONObject("options").optDouble("pageWidth", pageWidth);
-                }
-
-                double pageZoom = options.getPixelOptions().getDensity() / 72.0;
+                double pageZoom = pxlOpts.getDensity() / 72.0;
                 if (pageZoom <= 0) { pageZoom = 1; }
 
+                double pageWidth = 0;
+                double pageHeight = 0;
+
+                // web dimension use 96dpi (or equivalent dp/metric)
+                if (options.getDefaultOptions().getPageSize() != null) {
+                    pageWidth = options.getDefaultOptions().getPageSize().getWidth() * (96.0 / 72.0);
+                    //pageHeight = options.getDefaultOptions().getPageSize().getHeight() * (96.0 / 72.0);
+                }
+                if (!data.isNull("options")) {
+                    JSONObject dataOpt = data.getJSONObject("options");
+                    if (!dataOpt.isNull("pageWidth")) {
+                        pageWidth = data.optJSONObject("options").getDouble("pageWidth") * (96.0 / pxlOpts.getUnits().as1Inch());
+                    }
+                    if (!dataOpt.isNull("pageHeight")) {
+                        pageHeight = data.optJSONObject("options").getDouble("pageHeight") * (96.0 / pxlOpts.getUnits().as1Inch());
+                    }
+                }
+
                 try {
-                    images.add(WebApp.capture(source, (format == PrintingUtilities.Format.FILE), pageWidth, pageZoom));
+                    images.add(WebApp.capture(source, (format == PrintingUtilities.Format.FILE), pageWidth, pageHeight, pageZoom));
                 }
                 catch(IOException e) {
                     //JavaFX image loader becomes null if webView is too large, throwing an IllegalArgumentException on screen capture attempt
--- a/src/qz/printer/action/PrintRaw.java
+++ b/src/qz/printer/action/PrintRaw.java
@@ -9,8 +9,7 @@
  */
 package qz.printer.action;
 
-import org.apache.commons.io.Charsets;
-import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.ssl.Base64;
 import org.codehaus.jettison.json.JSONArray;
 import org.codehaus.jettison.json.JSONException;
@@ -25,9 +24,7 @@ import qz.printer.ImageWrapper;
 import qz.printer.LanguageType;
 import qz.printer.PrintOptions;
 import qz.printer.PrintOutput;
-import qz.utils.ByteUtilities;
-import qz.utils.FileUtilities;
-import qz.utils.PrintingUtilities;
+import qz.utils.*;
 
 import javax.imageio.ImageIO;
 import javax.print.*;
@@ -290,25 +287,20 @@ public class PrintRaw implements PrintProcessor {
      */
     public void printToAlternate(PrintService service, byte[] cmds) throws IOException, PrintException {
         File tmp = File.createTempFile("qz_raw_", null);
-        printToFile(tmp, cmds);
+        try {
+            printToFile(tmp, cmds);
+            String[] lpCmd = new String[] {
+                    "lp", "-d", PrintingUtilities.getPrinterId(service.getName()), "-o", "raw", tmp.getAbsolutePath()
+            };
+            boolean success = ShellUtilities.execute(lpCmd);
 
-        String shellCmd = String.format("/usr/bin/lp -d \"%s\" -o raw \"%s\";", service.getName(), tmp.getAbsolutePath());
-        log.trace("Runtime Exec running: {}", shellCmd);
-
-        Process process = Runtime.getRuntime().exec(new String[] {"bash", "-c", shellCmd});
-        try { process.waitFor(); }catch(InterruptedException ignore) {}
-
-        BufferedReader buf = new BufferedReader(new InputStreamReader(process.getInputStream(), Charsets.UTF_8));
-        String output = IOUtils.toString(buf);
-        buf.close();
-
-        log.info("Runtime Exec returned: {}", output);
-        if (process.exitValue() != 0) {
-            throw new PrintException(String.format("Alternate printing returned a non-zero value (%d).%n%s", process.exitValue(), output));
-        }
-
-        if (!tmp.delete()) {
-            tmp.deleteOnExit();
+            if (!success) {
+                throw new PrintException("Alternate printing failed: " + StringUtils.join(lpCmd, ' '));
+            }
+        } finally {
+            if (!tmp.delete()) {
+                tmp.deleteOnExit();
+            }
         }
     }
 
--- a/src/qz/printer/action/WebApp.java
+++ b/src/qz/printer/action/WebApp.java
@@ -41,6 +41,7 @@ public class WebApp extends Application {
     private static Stage stage;
     private static WebView webView;
     private static double pageWidth;
+    private static double pageHeight;
     private static double pageZoom;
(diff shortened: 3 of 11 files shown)


tresf:
@bberenz I think this one is ready for your input.  The utilities have been written and tested on Mac/Linux, but the caching and fetching would occur each time in the current implementation, so a proper home is greatly appreciated. :+1:

bberenz:
I've moved out the ShellUtilities calls to PrintingUtilities. They should now only occur on the first run, or if a printer is missing (as would likely be the case if one was plugged in or renamed after a print).

tresf:
Thanks. 

I reached out to the customer they shared this.... 

Apparently the CUPS commands are Locale Aware meaning the Mac printer reverse-lookup logic won't work so I'm not sure how to proceed here... 

screen shot 2016-03-23 at 11 01 57 am  screen shot 2016-03-23 at 11 08 01 am

tresf:
Turns out Java has support for setting up `export LANG=foo` prior to making the system calls.

@klabarge can you please test on Ubuntu using a non-English locale?
Clone and run the `dpi` branch and you should see something like this in console if a DPI isn't provided:
```
Parsed default density from CUPS Foo "Foo": 300 dpi
```


Testing results, Hoy!

screen shot 2016-03-23 at 3 14 17 pm

screen shot 2016-03-23 at 4 09 14 pm

tresf:
@bberenz so this is working terrific for `dpi` but once a user changes the width/height to `mm` the DPI gets all out of wack and java runs out of heap.

Is there a chance you're not converting density properly?

bberenz:
It's possible, though that usually shows as a `NegativeArraySizeException`, not an out of heap error.
Are you able to post any logs of the print job?

tresf:
> Are you able to post any logs of the print job?

I can later if needed, but if we force `density: 11.81102362205`, the crash goes away, so it's very clearly a bug associated with the new parsing logic.

tresf:
> it's very clearly a bug associated with the new parsing logic

Well, specifically how we handle the dpi after we've parsed it.

tresf:
@bberenz so this should be pretty easy to fix... 

In all other portions of the code, the `density` is relative to the units, but when we use the native density, it's not.

Take the below statement:

```java
Book book = new Book();
for(PDDocument doc : pdfs) {
    book.append(new PDFPrintable(doc, scale, false, (float)(density * pxlOpts.getUnits().as1Inch()), false), page, doc.getNumberOfPages());
}
```

 * When units are inches, the line `density * pxlOpts.getUnits().as1Inch()` resolves to `300.0`.
 * When units are mm, the line `density * pxlOpts.getUnits().as1Inch()` resolves to `7619.999885559082`

tresf:
Perhaps this... 

```java
if (res != null) {
   density = res.getFeedResolution(pxlOpts.getUnits() == PrintOptions.Unit.INCH ? PrinterResolution.DPI : PrinterResolution.DPCM * 10);
}
```

But this yields `304.7999954223633` instead of `300.0`, I assume due to rounding errors?

Also, @bberenz this DPI detection logic needs to be used for all Pixel printing, as well as `PrintPDF`.

tresf:
... or this... 

```java
        // this should be safe as zero * foo is still zero
        double density = pxlOpts.getDensity() * pxlOpts.getUnits().as1Inch();
        if (pxlOpts.getDensity() == 0 && pxlOpts.isRasterize()) {
            PrinterResolution res = (PrinterResolution)output.getPrintService().getDefaultAttributeValue(PrinterResolution.class);

            // Fallback on native workaround
            if (res == null && !SystemUtilities.isWindows()) {
                res = PrintingUtilities.getNativeDensity(output.getPrintService());
            }

            // Already to go in DPI form
            if (res != null) {
                density = res.getFeedResolution(ResolutionSyntax.DPI);
            }
        }

        Book book = new Book();
        for(PDDocument doc : pdfs) {
            book.append(new PDFPrintable(doc, scale, false, (float)density, false), page, doc.getNumberOfPages());
        }
```

src/qz/printer/action/PrintHTML.java
```diff
@@ -43,16 +43,29 @@ public void parseData(JSONArray printData, PrintOptions options) throws JSONExce
...
                 PrintingUtilities.Format format = PrintingUtilities.Format.valueOf(data.optString("format", "FILE").toUpperCase());
 
-                double pageWidth = PrinterJob.getPrinterJob().getPageFormat(null).getWidth();
+                double pageZoom = pxlOpts.getDensity() / 72.0;
+                if (pageZoom <= 0) { pageZoom = 1; }
+
+                double pageWidth = 0;
+                double pageHeight = 0;
+
+                // web dimension use 96dpi (or equivalent dp/metric)
```
tresf: This is very interesting.... Can you shed some light on this information?

src/qz/printer/action/WebApp.java
```diff
@@ -54,12 +55,14 @@ public void changed(ObservableValue<? extends Worker.State> ov, Worker.State old
...
                 try {
                     Reflect.on(webView).call("setZoom", pageZoom);
+                    log.trace("Zooming in by x{} for increased quality", pageZoom);
                 }
                 catch(ReflectException e) {
                     log.warn("Unable zoom, using default quality");
                     pageZoom = 1; //only zoom affects webView scaling
                 }
 
+                log.trace("Setting HTML page width to {}", (pageWidth * pageZoom));
```
tresf: A bit off topic, but how do we get these logs without firing a debug build?  These could be very useful when debugging with clients.

src/qz/printer/action/PrintHTML.java
bberenz: The [MDN on length](https://developer.mozilla.org/en-US/docs/Web/CSS/length#Absolute_length_units) goes over how css dimensions correspond to physical media.

src/qz/printer/action/WebApp.java
bberenz: I thought I set the log dialog to default to trace when shown, but I guess not.
I can make either the log file or dialog (or both) record at the trace level. I just need to know what you'd prefer.

src/qz/utils/PrintingUtilities.java
```diff
@@ -59,16 +82,7 @@ public static PrinterResolution getNativeDensity(PrintService service) {
...
-                CUPS_DESC = ShellUtilities.getCupsPrinters();
-                CUPS_DPI = ShellUtilities.getCupsDensities(CUPS_DESC);
-            }
-
-            for(String name : CUPS_DESC.keySet()) {
-                if (CUPS_DESC.get(name).equals(service.getName())) {
-                    return CUPS_DPI.get(name);
-                }
-            }
+            return CUPS_DPI.get(getPrinterId(service.getName()));
```
tresf: I realize this is confusing (calling a `.get` on a variable that was never initialized) but should be perfectly valid from an order of operations perspective.

src/qz/printer/action/PrintRaw.java
```diff
@@ -290,25 +288,19 @@ public void printJobRequiresAttention(PrintJobEvent printJobEvent) {
...
-        if (!tmp.delete()) {
-            tmp.deleteOnExit();
+        try {
+            printToFile(tmp, cmds);
+            boolean success = ShellUtilities.execute(new String[] {
+                    "lp", "-d", PrintingUtilities.getPrinterId(service.getName()), "-o", "raw", tmp.getAbsolutePath()
+            });
+
+            if (!success) {
+                throw new PrintException("Alternate printing failed.  Please troubleshoot the above commands manually.");
```
bberenz: Isn't this statement going to look out of context when a stacktrace is actually thrown?

src/qz/utils/PrintingUtilities.java
bberenz: Why was `CUPS_DPI` moved out if it is only needed in this method?

tresf: > Why was CUPS_DPI moved out if it is only needed in this method?

It certainly doesn't need to be, but since `CUPS_DPI` and `CUPS_DESC` are so closely coupled, it didn't make sense in my mind to cache one now and then the other later, it just seemed like we'd have strange scenarios where one was updated and the other was not.  Thoughts?

src/qz/printer/action/PrintRaw.java
tresf: > Isn't this statement going to look out of context when a stacktrace is actually thrown?

Yeah, I hadn't thought of that.  I'll fix it.

src/qz/utils/PrintingUtilities.java
tresf: I'm getting a NPE on this line now I believe.  Should we just make the code redundant you think?

bberenz: Well the call to get `CUPS_DPI`'s values relies on `CUPS_DESC`, so I don't think we'd need to worry about a synchronization issue between what's available in each map. And if they are only raw printing, as this other use case for `CUPS_DESC` is, then they shouldn't have to worry about dpi anyway, right?

tresf: > if they are only raw printing, as this other use case for CUPS_DESC is, then they shouldn't have to worry about dpi anyway, right?

Right.

> Well the call to get CUPS_DPI's values relies on CUPS_DESC, so I don't think we'd need to worry about a synchronization issue between what's available in each map.

Right... just seemed wasteful but I'll bring back the somewhat redundant logic.

src/qz/printer/action/PrintHTML.java
tresf: Hmm... Perhaps I'm reading the wrong portion, but the part I see says `px` is relative to the viewing device... 

> px
Relative to the viewing device.
For screen display, typically one device pixel (dot) of the display.
For printers and very high resolution screens one CSS pixel implies multiple device pixels, so that the number of pixel per inch stays around 96.

I guess I'm still confused as to how this was discovered, what impact it would have had on previous builds vs. new builds, etc.

bberenz: This was discovered when I tried to print a 2.25in div on a 2.25in label and it came out with excessive whitespace. Viewing the resulting pixel size of the html element showed 96dpi, a little investigating lead to sources like this stating the same. But the first three paragraphs under the heading are the focus
>Absolute length units represents a __physical measurement and when the physical properties of the output medium are known, such as for print layout__. This is done by anchored one of the unit to a physical unit and to defined the other relatively to it. The anchor is done __differently for low-resolution devices, like screens, and high-resolution devices, like printers__.

>For low-dpi devices, the unit px represents the physical reference pixel and the others are defined relative to it. __Thus, 1in is defined as 96px__ which equals 72pt. The consequence of this definition is that on such devices, length described in inches (in), centimeters (cm), millimeters (mm) doesn't necessary match the length of the physical unit with the same name.

>For high-dpi devices, inches (in), centimeters (cm), millimeters (mm) are defined as their physical counterparts. Therefore the __px unit is defined relative to them (1/96 of 1 inch)__.

For low end devices, it's basically saying the unit on the device won't match a physical inch because of this 96dpi conversion.
You're also reading the px css description, which for someone laying out a page for printing they shouldn't use because they cannot be sure of the printer dpi in all cases. These changes allow them to just use the inch/cm css units and get that same amount on the physical medium.

PR #24, opened 2016-04-06
Add multi-res icon support

tresf:
Not ready for merge... Still working out some issues with Ubuntu...


tresf:
Turns out, this was happening in the old version too, just wasn't as obvious.... 

image

PR #66, opened 2016-06-07
Use English locale for string case conversion

onurguzel:
This commit fixes a bug when the application runs on an environment with Turkish locale. Turkish have two separate "i" character dotted and dotless. Per the Unicode standard, our lowercase "i" becomes "İ" (U+0130 "Latin Capital Letter I With Dot Above") when it moves to uppercase. Similarly, our uppercase "I" becomes "ı" (U+0131 "Latin Small Letter Dotless I") when it moves to lowercase. Using English locale prevents case conversion with an incorrect locale.

For additional info, you can check out the links below:
http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html
https://blog.codinghorror.com/whats-wrong-with-turkey/
http://mattryall.net/blog/2009/02/the-infamous-turkish-locale-bug

### DIFF
--- a/src/qz/printer/PrintOptions.java
+++ b/src/qz/printer/PrintOptions.java
@@ -15,6 +15,7 @@ import java.awt.*;
 import java.awt.print.PageFormat;
 import java.awt.print.PrinterException;
 import java.awt.print.PrinterJob;
+import java.util.Locale;
 
 public class PrintOptions {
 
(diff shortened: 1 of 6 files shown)


tresf:
@onurguzel Thanks for the detailed write up.  This should be safe to merge.

@bberenz any objections here?

bberenz:
Never knew about this quirk. Code looks fine, so no objections here.

tresf:
FYI, Please ignore Travis, it appears to be failing for reasons unrelated to the PR. :+1:

PR #25, opened 2016-04-06
Add new library for HID support

bberenz:
New libraries have been added for HID support.
This is in a working state and ready for testing across devices.

Usb events were given a common interface (DeviceIO) to help reduce duplicate code sitting in the PrintSocketClient. So UsbIO was refactored a little around this, and HidIO of course follows it.

The response on callbacks for streaming (for serial, usb, and hid) has changed to be a complete object, rather than broken into parameters. This is to allow for identifying response types, `RECEIVE` as general data, `ERROR` as failures, and `ACTION` for listeners (currently only HID).

A bit of refactoring on PrintSocketClient happened. Turning some calls into static methods so they could be moved to their respective Utilities classes and keep the client object focused on processing receive/send events on the websocket. This doesn't affect anything for connecting, as they were all private instance calls before.

PR #69, opened 2016-06-14
Add new library for HID support

tresf:
Supersedes #25

--- 

From @bberenz: 

New libraries have been added for HID support.
This is in a working state and ready for testing across devices.

Usb events were given a common interface (DeviceIO) to help reduce duplicate code sitting in the PrintSocketClient. So UsbIO was refactored a little around this, and HidIO of course follows it.

The response on callbacks for streaming (for serial, usb, and hid) has changed to be a complete object, rather than broken into parameters. This is to allow for identifying response types, RECEIVE as general data, ERROR as failures, and ACTION for listeners (currently only HID).

A bit of refactoring on PrintSocketClient happened. Turning some calls into static methods so they could be moved to their respective Utilities classes and keep the client object focused on processing receive/send events on the websocket. This doesn't affect anything for connecting, as they were all private instance calls before.


tresf:
FYI, the exit crash still occurs with the patch on OS X 10.11... https://gist.github.com/tresf/0813e8d20385d5d8eccc6c8634602722

Steps:

1. Leave scale disconnected
1. Open QZ Tray
1. Open Demo Page
1. Switch to HID tab
1. Click "Listen for events"
1. Attach scale
1. From the HID tab click "List Devices"
1. Locate the scale and click "Claim device"
1. Without releasing device and without stop listening, close QZ Tray
1. The above crash log appears

tresf:
@bberenz A second note... like I had stated before, the flickering weight needs to be addressed, even if only in JavaScript.


untitled

tresf:
@bberenz Kyle and I tested extensively and we can confirm that 5e45845 prevents the random crashes from occurring.  We tested the software about 10 times each.
 * Without the patch the software crashes from 40 seconds to 7 minutes.
 * With the patch the software does not crash while running

The remaining items are:
 * Thoughts on the exit crash (this isn't easily reproducible, so this is can wait until after the PR is merged)
 * Thoughts on the flickering of weight

bberenz:
> Thoughts on the flickering of weight

I believe it is caused by the calls trying to get data faster than it's available, did you ever try:
```diff
+   device.setNonBlocking(false);
    this.device = device;
}
```
in the `HidIO` constructor?

---
> Thoughts on the exit crash

From the old PR:

>> The java logs seem to indicate you might have a stream open when closing. It has to be timed just so to hit an error (which would explain the reproduction problems), but does this crash happen only when that condition is true? Or have you had it crash on close without streaming?

Were you ever able to confirm that you _didn't_ need a stream open to cause the crash?
How are you closing QZ by the way? The tray menu, right?
And your steps to reproduce there include activating "Listen for Events", which more or less is what you explicitly turned off in your fix.. We might have to disable that call entirely on OSX to avoid this.

tresf:
> 
I believe it is caused by the calls trying to get data faster than it's available, did you ever try:
```diff
+   device.setNonBlocking(false);
    this.device = device;
}
```
in the `HidIO` constructor?


No, we'll test that tonight and see if it helps.

> Were you ever able to confirm that you _didn't_ need a stream open to cause the crash?

It's too intermittent to know.  We have no way of reliably reproducing this in QZ Tray.

> And your steps to reproduce there include activating "Listen for Events", which more or less is what you explicitly turned off in your fix.. We might have to disable that call entirely on OSX to avoid this.

I know you feel this way, but that's now what we observe.  The fix doesn't turn this off in our testing so thoughts welcome on that.

tresf:
> The java logs seem to indicate you might have a stream open when closing. It has to be timed just so to hit an error

If this hunch is true, is there a way to explicitly close all streams at close so that we can circumvent the crash?

bberenz:
> is there a way to explicitly close all streams at QZ Tray close

`SocketConnection:108` `disconnect()`
And all the SocketConnection objects are handled by PrintSocketClient, which calls this method when the respective websocket closes.

tresf:
Thanks although I'm not sure that addressed the question.  In the event of our Mac crash, we had to explicitly call `service.shutdown();` to prevent a crash however we're not doing that in all cases.  For example here:

tresf:
I can confirm that the List Devices button will crash the software in a similar fashion to what was happening with `findDevice`.  @bberenz do you anticipate any issues if I just wrap the shutdown logic into `getHidDevices` and call that instead of `service.getAttachedHidDevices();` on line 50?

bberenz:
> do you anticipate any issues if I just wrap the shutdown logic into getHidDevices

No, it should be fine.

tresf:
11c1e58 seems to fix the "List Devices" crash. Now onto the flickering...

tresf:
And I can confirm that 11c1e58  finally kills the "Listen for Events" on Mac.

tresf:
> 
I believe it is caused by the calls trying to get data faster than it's available, did you ever try:
```diff
+   device.setNonBlocking(false);
    this.device = device;
}
```
in the `HidIO` constructor?


Just did.... no dice.

untitled

tresf:
After close examination, the data sent from the MT scale that starts with `04` seems to be redundant in USB mode and erroneous in HID mode suggesting optimistically, we can probably toss all messages starting with `04` in the first slot.

Per 494b160 and on the side of caution, I've added a commit to this branch which filters the erroneous `00 00 00 00 00 00` data chunks, as they seem to be quite unlikely in scale data as they omit the precision, units and weight so it should be safe.

@klabarge if you can get your scale to detect on your Mac again, can you try out the latest commit?

tresf:
@bberenz I think this is good to rebase and merge. :+1:

bberenz:
I tried a rebase. Not sure if it performed correctly though. Threw a _lot_ of problems at me.

tresf:
sorry, rebase may have been the wrong terminology in this case.   Fixing the merge conflicts and then merging is what I was aiming for.

bberenz:
I believe I've cleaned up my mess, the `2.0` branch just has a merge commit into the `hid_code` branch now, and should be ready for you to handle from here.

PR #74, opened 2016-07-10
Don't silently disable secure ports

dsanders11:
I ran into this during development, and had to use breakpoints to find why insecure ports kept being attempted even when `usingSecure` was set to true.

I understand the reasoning for the disabling, but it makes development a PITA since in this case I was trying to update the JavaScript code and test it with an existing QZ Tray deployment and wanted to verify the use of secure ports and request signing.

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -617,7 +617,7 @@ var qz = (function() {
                         if (options == undefined) { options = {}; }
 
                         //respect forcing secure ports if it is defined, otherwise disable
-                        if (options.usingSecure === 'undefined') {
+                        if (options.usingSecure === undefined) {
                             _qz.log.trace("Disabling secure ports due to insecure page");
                             options.usingSecure = false;
                         }
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -615,7 +615,12 @@ var qz = (function() {
                     //disable secure ports if page is not secure
                     if (typeof location === 'undefined' || location.protocol !== 'https:') {
                         if (options == undefined) { options = {}; }
-                        options.usingSecure = false;
+
+                        //respect forcing secure ports if it is defined, otherwise disable
+                        if (options.usingSecure === 'undefined') {
+                            _qz.log.trace("Disabling secure ports due to insecure page");
+                            options.usingSecure = false;
+                        }
                     }
 
                     var attempt = function(count) {


js/qz-tray.js
```diff
@@ -615,7 +615,12 @@ var qz = (function() {
                     //disable secure ports if page is not secure
                     if (typeof location === 'undefined' || location.protocol !== 'https:') {
                         if (options == undefined) { options = {}; }
-                        options.usingSecure = false;
+
+                        //respect forcing secure ports if it is defined, otherwise disable
+                        if (options.usingSecure === 'undefined') {
```
tresf: @dsanders11, typo... this should be either `=== undefined` (without single quotes) or `typeof ... === 'undefined'` otherwise the conditional fails.

dsanders11: Interesting, pretty sure I grabbed the exact code I was running that was working fine, but I've also been burning the midnight oil and JavaScript isn't my language of choice, so adjust as needed. 👍

tresf: Ok.  FYI, PRs need to be fixed and merged or rejected so if you don't fix this on your branch I'll have to close this out and do the commit myself. :+1:

js/qz-tray.js
```diff
@@ -617,7 +617,7 @@ var qz = (function() {
                         if (options == undefined) { options = {}; }
 
                         //respect forcing secure ports if it is defined, otherwise disable
-                        if (options.usingSecure === 'undefined') {
+                        if (options.usingSecure === undefined) {
                             _qz.log.trace("Disabling secure ports due to insecure page");
```
tresf: :+1:

js/qz-tray.js
```diff
@@ -615,7 +615,12 @@ var qz = (function() {
                     //disable secure ports if page is not secure
                     if (typeof location === 'undefined' || location.protocol !== 'https:') {
                         if (options == undefined) { options = {}; }
-                        options.usingSecure = false;
+
+                        //respect forcing secure ports if it is defined, otherwise disable
+                        if (options.usingSecure === undefined) {
```
dsanders11: @tresf, changed it. I suggest doing a squash merge to drop the extraneous extra commit.

tresf: Much obliged.  Luckily GitHub offers a button for it now. :+1:

PR #80, opened 2016-07-19
Fix scale weight readings in demo

dsanders11:
Was erroneously pegged to 'lbs' prior due to checking units after splicing the data rather than before like in the example code


tresf:
:+1: Much obliged.

tresf:
Hmm... this should have gone against 2.0. 

Cherry-picked in via f4894cc967a73a07c2d07c3cbc63bcc0ea40760b. :+1:

PR #140, opened 2016-11-21
Build Constants.java from ant properties, get project version from git

lukas-w:
Moved Constants.java to Constants.java.template, which is now being processed by ant's `expandproperties` filter. The project version is read using `git describe` (formatted as e.g. `2.0.1-14` when building from a branch 14 commits ahead of the tag v2.0.1). We may still want to have the current release hard-coded somewhere as a fallback when git is not available.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 /.idea/compiler.xml
 /README.md
 .DS_Store
+src/qz/common/Constants.java
+out
--- a/ant/project.properties
+++ b/ant/project.properties
@@ -5,6 +5,7 @@ vendor.email=support@qz.io
 
 project.name=QZ Tray
 project.filename=qz-tray
+project.version=2.0.2
 
 launch.opts=-Xms512m
 
--- a/build.xml
+++ b/build.xml
@@ -25,7 +25,7 @@
         <delete dir="${out.dir}"/>
     </target>
 
-    <target name="compile-socket" depends="init">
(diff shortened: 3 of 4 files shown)


tresf:
>  We may still want to have the current release hard-coded somewhere as a fallback when git is not available.

This definitely is a concern of mine because we run from IDE (IntelliJ) quite often, which won't run unless the project has first been configured by ant.

Lukas-W:
Added a default version to `project.properties` and modified `build.xml` to not fail when git's unavailable on the system

tresf:
> Added a default version to project.properties and modified build.xml to not fail when git's unavailable on the system

I think this will still fail when run from IDE tho.

Lukas-W:
I guess I'm out of my depth here then. Doesn't IntelliJ use ant's build.xml?

tresf:
> Doesn't IntelliJ use ant's build.xml?

No, it uses some [custom logic](http://stackoverflow.com/a/25066395/3196753).  Note, IntelliJ community edition works great on Linux if you have time to see how it works.

https://qz.io/wiki/Compiling#intellij

tresf:
We still need a way of auto-incrementing the builds but our White-Label portal has this ability already and making IntelliJ and ant both happy is going to take some reconsideration.  I'm closing this PR, but we'll reference it for the future (e.g. if we ever start releasing nightly builds to the general public).

PR #141, opened 2016-12-03
Update sign-message.py

dsanders11:
M2Crypto doesn't support Python 3 and isn't regularly maintained, so it's not a great library to use for example code.  [pyca/cryptography](https://github.com/pyca/cryptography) is a better maintained and modern library to use.

Also added a few minor changes to better match Python style.

### DIFF
--- a/assets/signing/sign-message.py
+++ b/assets/signing/sign-message.py
@@ -22,22 +22,29 @@
 #                                                       #
 #########################################################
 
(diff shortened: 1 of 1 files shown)


tresf:
From https://pypi.python.org/pypi/cryptography:

> It supports Python 2.6-2.7, Python 3.3+, and PyPy 2.6+.

Great, thanks!

PR #146, opened 2016-12-17
Update .gitignore

dsanders11:
Drop mentions of qz-print directory which I don't believe should exist anymore in this repo, and add out directory as ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,4 @@
-/pdf-renderer_0.9.1_qz/build/
-/pdf-renderer_0.9.1_qz/dist/
-/jssc_2.6.0_qz/build/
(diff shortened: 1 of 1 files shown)


tresf:
You might as well go ahead and remove these while your'e at it... They're from the [1.8 branch](https://github.com/qzind/qz-print/tree/1.8).
```diff
-  /pdf-renderer_0.9.1_qz/dist/
-  /jssc_2.6.0_qz/build/
-  /jssc_2.6.0_qz/dist/
```

I'm surprised `/out` wasn't already added as `git` never tries to add that when I do a commit... :\

dsanders11:
Alright, I redid the branch with those removed as well. If it looks good go ahead and merge it.

Issue #60, opened 2016-05-24
Ubuntu 14.04 could not start, a null pointer exception


vkosachev:
Hi guys, experiencing this issue on launch. And so I cannot connect to qz from JS.
What did I miss ?

PS. : By default I have Java 8 installed, but in the example below I tried to use JDK to launch.

```bash
.qz/jdk1.8.0_91/bin/java -jar /opt/qz-tray/qz-tray.jar
Gtk-Message: Failed to load module "pantheon-filechooser-module"
Gtk-Message: Failed to load module "pantheon-filechooser-module"
Gtk-Message: Failed to load module "canberra-gtk-module"
[DEBUG] 2016-05-24 12:03:00,839 @ qz.utils.ShellUtilities:?
	Executing: [uname, -a]
[DEBUG] 2016-05-24 12:03:00,848 @ qz.utils.ShellUtilities:?
	Executing: [gconftool-2, --get, /desktop/gnome/shell/windows/theme]
[DEBUG] 2016-05-24 12:03:00,852 @ qz.utils.ShellUtilities:?
	Executing: [grep, dark_bg_color, /usr/share/themes/Ambiance/gtk-3.0/gtk.css]
[DEBUG] 2016-05-24 12:03:00,853 @ qz.utils.ShellUtilities:?
	Executing: [grep, dark_bg_color, /usr/share/themes/Ambiance/gtk-3.0/gtk-main.css]
[INFO] 2016-05-24 12:03:00,859 @ qz.deploy.DeployUtilities:?
	SSL properties file from /opt/qz-tray/qz-tray.properties
[DEBUG] 2016-05-24 12:03:00,860 @ qz.utils.ShellUtilities:?
	Executing: [certutil, -d, sql:/home/vkosachev/.pki/nssdb, -A, -t, TC, -n, QZ Industries, LLC, -i, /opt/qz-tray/auth/qz-tray.crt]
[DEBUG] 2016-05-24 12:03:01,584 @ qz.auth.Certificate:?
	Using trusted root certificate: CN=qzindustries.com, O=QZ Industries, LLC (8ef2c0b2fe3f75754245b3fb9d54c0caa1b46bb3)
[INFO] 2016-05-24 12:03:01,584 @ qz.auth.CRL:?
	Loading CRL from https://crl.qz.io
[ERROR] 2016-05-24 12:03:01,802 @ qz.ws.PrintSocketServer:?
	Could not start tray manager
java.lang.reflect.InvocationTargetException
[...]

Issue #52, opened 2016-05-08
Hung on launch "java.lang.NoSuchFieldError"


kilbot:
I have previously had success with version 1.9, but when trying to install 2.0.0-rc3 the tray manager won't start. It seems perhaps there is a auth problem with java:

```
$ /Applications/QZ\ Tray.app/Contents/MacOS/QZ\ Tray ; exit;
[ERROR] 2016-05-08 13:34:54,183 @ qz.ws.PrintSocketServer:?
    Could not start tray manager
java.lang.reflect.InvocationTargetException
    at java.awt.EventQueue.invokeAndWait(EventQueue.java:1321)
    at java.awt.EventQueue.invokeAndWait(EventQueue.java:1296)
    at javax.swing.SwingUtilities.invokeAndWait(SwingUtilities.java:1348)
    at qz.ws.PrintSocketServer.main(Unknown Source)
Caused by: java.lang.NoSuchFieldError: CN
    at qz.auth.Certificate.<init>(Unknown Source)
    at qz.auth.Certificate.<clinit>(Unknown Source)
    at qz.common.TrayManager.addMenuItems(Unknown Source)
    at qz.common.TrayManager.<init>(Unknown Source)
    at qz.ws.PrintSocketServer$1.run(Unknown Source)
    at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:301)
    at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
    at java.awt.EventQueue.access$500(EventQueue.java:97)
    at java.awt.EventQueue$3.run(EventQueue.java:709)
    at java.awt.EventQueue$3.run(EventQueue.java:703)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
    at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
    at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
    at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
    at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
    at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
    at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
[...]

PR #149, opened 2016-12-28
Added better support for systemTray

Vzor-:
Some outstanding issues
1. Notifications and Tooltips don't work. They are not yet supported by dorkbox but may be later.
2. Gtk 2vs3 collisions need to be dodged, most are done but https://github.com/Vzor-/tray/blob/2.1/src/qz/ui/LinkLabel.java#L42 eludes me.
3. setLook&feel causes a Gtk collision on elementaryOS (and presumably anything running gtk+) For the moment we just don't run it on those systems https://github.com/Vzor-/tray/blob/2.1/src/qz/common/TrayManager.java#L101
4. SLF4J: Class path contains multiple SLF4J bindings. (one in dorkbox one for log4j) 

Requires testing on every operating system
(Eventually) Fixes https://github.com/qzind/tray/issues/60


src/qz/common/TrayManager.java
```diff
@@ -92,36 +95,33 @@ public TrayManager(boolean isHeadless) {
         shortcutCreator = DeployUtilities.getSystemShortcutCreator();
         shortcutCreator.setShortcutName(Constants.ABOUT_TITLE);
 
-        SystemUtilities.setSystemLookAndFeel();
+        // TODO: DON'T FORGET TO ADDRESS THIS. It causes a gtk2 vs 3 error on elementaryOS
+        // This at least mitigates the problem, still need to fix
+        if (!dorkbox.util.OS.isLinux() || dorkbox.systemTray.jna.linux.Gtk.isGtk2) {
```
Vzor-: This could be done with our isLinux(). I am using the dorkbox one atm just to ensure that its behavior lines up with that of the systemtray itself.

src/qz/common/TrayManager.java
```diff
@@ -92,36 +95,33 @@ public TrayManager(boolean isHeadless) {
...
+                // tray.setToolTip(name); Not supported by Dorkbox (yet)
+            }
+            catch(Exception e ) {
+                log.error("Could not initialize tray, forcing headless mode", e);
+                headless = true;
             }
-            catch(AWTException awt) {
-                log.error("Could not attach tray, forcing headless mode", awt);
+            if (!headless && (tray == null)) {
+                log.error("Unknown tray init error, most likely unsupported os, Forcing headless mode");
```
Vzor-: If the os doesn't support systemtrays, systemtray.get*() is returned null. This should catch every error except for Gtk errors that happen outside of the JVM.

src/qz/common/TrayManager.java
```diff
@@ -156,101 +156,94 @@ public static void main(String args[]) {
...
+        MenuItem aboutItem = new MenuItem("About...", iconCache.getImage(IconCache.Icon.ABOUT_ICON), aboutListener);
+        aboutItem.setShortcut('b');
+        tray.getMenu().add(aboutItem);
+
+        aboutDialog = new AboutDialog(null, iconCache, name);
+        {
+            JMenuItem siteButton = new JMenuItem("Site Manager...", iconCache.getIcon(IconCache.Icon.SAVED_ICON));
+            siteButton.setMnemonic(KeyEvent.VK_M);
+            siteButton.addActionListener(savedListener);
+            sitesDialog = new SiteManagerDialog(siteButton, iconCache);
```
Vzor-: Kind of ugly, but if we need buttons on windows, there is no way around it. This code could be moved into aboutDialog but bleh.

src/qz/common/TrayManager.java
```diff
@@ -156,101 +156,94 @@ public static void main(String args[]) {
...
+            Checkbox j = (Checkbox)e.getSource();
+            prefs.setProperty(Constants.PREFS_NOTIFICATIONS, j.getChecked());
         }
     };
 
     private final ActionListener openListener = new ActionListener() {
         public void actionPerformed(ActionEvent e) {
             try {
-                ShellUtilities.browseDirectory(shortcutCreator.getParentDirectory());
+                boolean avoidGTK2 = dorkbox.util.OS.isLinux() && !dorkbox.systemTray.jna.linux.Gtk.isGtk2;
```
Vzor-: We use this enough a property may be better

src/qz/common/TrayManager.java
```diff
@@ -402,7 +395,7 @@ public boolean showGatewayDialog(final Certificate cert, final String prompt) {
                     log.info("Denied {} to {}", cert.getCommonName(), prompt);
                     if (gatewayDialog.isPersistent()) {
                         if (Certificate.UNKNOWN.equals(cert)) {
-                            anonymousItem.doClick(); // if always block anonymous requests -> flag menu item
+                            anonymousItem.setChecked(true); // if always block anonymous requests -> flag menu item
```
Vzor-: is this correct? I read it as, it toggles only when it is false, so we could just set it to true instead.

src/qz/common/TrayManager.java
```diff
@@ -523,7 +516,10 @@ public void setWarningIcon() {
     /** Thread safe method for setting the specified icon */
     private void setIcon(final IconCache.Icon i) {
         if (tray != null) {
-            SwingUtilities.invokeLater(() -> tray.setImage(iconCache.getImage(i, tray.getSize())));
+            // Gross, if you know a better way, feel free to change this
+            Image blank = new ImageIcon(new byte[1]).getImage();
+            Dimension size = new TrayIcon(blank).getSize();
+            tray.setImage(iconCache.getImage(i, size));
```
Vzor-: This functionality will be hard to verify since systemTray auto resizes icons

src/qz/ui/AboutDialog.java
```diff
@@ -22,7 +22,7 @@
     String name;
 
     public AboutDialog(JMenuItem menuItem, IconCache iconCache, String name) {
-        super(menuItem, iconCache);
+        super(null, "About", iconCache);
```
Vzor-: I mean, when will it not be "About". This lets me pass null as a jMenuItem

src/qz/ui/LinkLabel.java
```diff
@@ -34,22 +34,21 @@ public LinkLabel() {
...
-
-                }
-                catch(Exception ex) {
-                    log.error("", ex);
+        addActionListener(e -> {
+            try {
+                // Sense the action based on the content of the text
+                if (text.contains("@")) {
+                    //TODO This will cause gtk errors on elementary... I think...
+                    Desktop.getDesktop().mail(new URI(text));
```
Vzor-: I don't even know what to do here. Desktop.getDesktop() will immediately cause an uncatchable error if Gtk 3 is running. Is there a generic way to do this on linux?

tresf:
Some notes: 

 * It seems evident that we need a static method or flag for helping with the Gtk workarounds.  Perhaps `SystemUtilities.isGtkConflict()`.
 * In regards to the `AboutDialog` being constructed with a redundant `JMenuItem`, please just switch the constructor to use the dorkbox menu item instead for now.
 * On that last note... I'm not particularly fond of the fact that dorkbox has chosen to reimpliment his own menu items instead of extending that which Swing already offers.  I'd rather not have dorkbox components spread all around the classes simply because he fails to implement `JMenuItem` or `AbstractButton`.
 * The dependency should be all lower-case.  Please rename to `dorkbox-3.0-rc3.jar`.
 * Please revert the `.iml` and `build.xml` changes for now.  We haven't decided how to handle these permanently.  If the issues with JAR search order (https://github.com/qzind/tray/issues/52#issuecomment-269430333) crop up on other workstations we can decide to explicitely add a bouncycastle release.  I think @bberenz can help us out with that.
 * If the native system tray does resizing automatically, please just hard-code a dimension of `32`, which should yield the highest of quality.
 * Please revert `anonymousItem.setChecked`.  `doClick` triggers the action listener, `setChecked` does not.
 
I think that covers everything in the PR so far.

Vzor-:
Alright! The end is near,

> Please don't assume isLinux either, just look for Gtk please.

I just convinced dorkbox to add isGTK3 to the systemtray lib

> please just hard-code a dimension of 32

Should I delete the other image files and logic in iconCache as well?

> doClick triggers the action listener

If I manually fire the callback and set the listener to true, would that be equivalent?

on another note, do you know of a way to do `Desktop.getDesktop().mail(new URI(text));` without using getdesktop (exclusively on systems using Gtk3)?

tresf:
> I just convinced dorkbox to add isGTK3 to the systemtray lib

👍 

> Should I delete the other image files and logic in iconCache as well?

No, we need them for other OSs, just tell IconCache you want the `32 px` version by passing a size of 32x32.

> If I manually fire the callback and set the listener to true, would that be equivalent?

Sure.  I assume there's no `doClick` in his library?

> `Desktop.getDesktop().mail(new URI(text));`

Have you tried [`xdg-email`](https://linux.die.net/man/1/xdg-email)?

Vzor-:
Ok I am preparing a commit, do you want everything squashed or will you just do that when you pull

.idea/tray.iml
```diff
@@ -27,4 +27,3 @@
     </orderEntry>
   </component>
 </module>
-
```
tresf: ^---

tresf:
Testing ElementaryOS 0.3.2 x64:

 * **Hard Crash:**  About dialog URL needs to be patched to use `xdg-open`
   ```
   (java:7885): Gtk-ERROR **: GTK+ 2.x symbols detected. Using GTK+ 2.x and GTK+ 3 in the same process is not supported
   ```
 * **Hard Crash:** HTML (JavaFX) printing is a hard crash as well.
   ```
   (java:8114): GLib-GObject-WARNING **: cannot register existing type 'GdkWindow'

   (java:8114): Gdk-WARNING **: /build/gtk+2.0-3ongxb/gtk+2.0-2.24.23/gdk/x11/gdkdrawable-x11.c:952 drawable is not a pixmap or window

   (java:8114): Gdk-ERROR **: The program 'java' received an X Window System error.
   This probably reflects a bug in the program.
   The error was 'BadWindow (invalid Window parameter)'.
     (Details: serial 131 error_code 3 request_code 20 (core protocol) minor_code 0)
   ```

dorkbox:
@tresf I agree with you most of the time, but a *tiny nit* --- The `SystemTray` doesn't use Swing components for it's API (`JMenu`, `AbstractButton`, or their sub-classes) because there is undefined behavior if they were to be used without a java native peer and there is an expectation otherwise. This is an important distinction because the `SystemTray` menus and entries do not always have java native peers the way those objects expect.  I know that you understand the difference, but a quick google search for "java system tray" shows that there problems going back more than 10+ years on this topic.

Given that a good deal of the menu system and event dispatch had to be re-implemented in a cross-platform way, I felt the best solution was a clean and concise API that doesn't leave room for ambiguity or undefined behavior. For example, if one examines the source of [AbstractButton](http://grepcode.com/file_/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/javax/swing/AbstractButton.java) and compares it to the source for [MenuItem](https://github.com/dorkbox/SystemTray/blob/91c401e389658b10bf580796f8a6adda4575e5ec/src/dorkbox/systemTray/MenuItem.java) (or [Entry](https://github.com/dorkbox/SystemTray/blob/91c401e389658b10bf580796f8a6adda4575e5ec/src/dorkbox/systemTray/Entry.java)), it should make sense.

Admittedly,  it pains me that the `JMenu`, etc components could not be re-used in a safe and clean way - as I wanted the `SystemTray` to be as easy for developers to use as possible, and would have preferred a 100% "drop in" replacement; ultimately the decision was reached to make sure it is absolutely clear that it is neither a Swing nor JavaFX menu. For the majority of the time, a simple change from `JMenu` -> `Menu` is sufficient, as I mirrored as much of the Swing API that was applicable.

Please understand it wasn't a failure or ignoring advice/input, but rather a careful move on the side of caution to prevent undefined behavior and maintain a clean API.

tresf:
@dorkbox I've moved the above conversation here, I hope that's OK. https://github.com/dorkbox/SystemTray/issues/36

dorkbox:
@tresf no problem.

Vzor-:
Two major issues fixed with systemTray, starting final round of testing.
https://github.com/dorkbox/SystemTray/commit/687035c3f1ef882231fa0a3c6813f26b829f3ad8
https://github.com/dorkbox/SystemTray/issues/43

tresf:
Superseded by #163

PR #170, opened 2017-02-23
Fixed "document is not defined" when qz running on node.js

burya4ok:
Added checking in function "absolute" when print image

### DIFF
--- a/assets/signing/sign-message.R
+++ b/assets/signing/sign-message.R
@@ -1,0 +1,39 @@
+#
+# Echoes the signed message and exits
+# usage:  R sign-message.R "test"
+#
+
+#########################################################
+#             WARNING   WARNING   WARNING               #
+#########################################################
+#                                                       #
+# This file is intended for demonstration purposes      #
+# only.                                                 #
+#                                                       #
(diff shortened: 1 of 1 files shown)


tresf:
@burya4ok thanks!

So that this patch works for other non-browser environments, can you please try this instead?

```js
if (typeof window !== 'undefined' && typeof document.createElement === 'function') { ... }
```

burya4ok:
Yes, this patch works for me.

tresf:
@burya4ok thanks for testing.  If you are ok adjusting your pull request, we'd happy accept it.

tresf:
@burya4ok thank you very much.  Unfortunately the pull request now has unnecessary changes to `.idea/codeStyleSettings.xml`.  Would you so kindly `git revert c79efc3` and then we will merge this in?

tresf:
GitHub wouldn't let me fix the xml whitespace through the browser, so I squashed the commits, but must have put the HEAD in a strange state, it recorded zero commits and couldn't resurrect them.

Anyway... cherry-picked into https://github.com/qzind/tray/commit/7091497f42a2cb0a9aeef394ceb990ee4c012f51.

Issue #162, opened 2017-02-06
RTF Document Support


tresf:
This is a gap which was never brought over.

https://github.com/qzindustries/qz-print/pull/1/files#diff-f4216bab2f36aded65b021c0844d7e98R195

```java
JEditorPane rtfEditor = new JEditorPane();
rtfEditor.setBackground(Color.white);
rtfEditor.setVisible(false);
rtfEditor.setContentType("text/rtf");
rtfEditor.setText(data);
```

PR #171, opened 2017-02-24
Set User-Agent for Requests

mrloop:
Currently defaults to Java version for example `Java/1.8.0_121`, instead
use `QZTray/2.0.2 (Linux; 4.4.0-63-generic) Java/1.8.0_121 (Oracle
Corporation)`

Useful for analytics and resolving issues.

see https://tools.ietf.org/html/rfc7231#section-5.5.3


tresf:
@mrloop we've had several requests to add user agent strings to our requests and we've declined all of them since we don't consider our application a conventional browser.  Most requests have come from strange back-end filtering settings (e.g. NetSuite) which block user agents that don't match a particular pattern.

The difference being, your request is feature complete and causes no harm, but if we accept, I have a few requests... 

1. Rebase against 2.1 branch.  We'd prefer to not make this change in 2.0 due to the volume of users which may already be expecting `Java/1.8.0_121`.
2. Wrap it into a utility function, in case other headers are needed, we can benefit from wrapping all requests for new file formats (e.g. #162) as well as future header strings and relocates the static function from CRL.

Once this is completed, we'll test against the major OS's (7, 10, macOS, Ubuntu) and make sure everything matches as close as possible to what a browser would send, then merge.

mrloop:
Closing this PR as against branch `2.0`

PR #175, opened 2017-02-28
Set User-Agent for Requests

mrloop:
Currently defaults to Java version for example `Java/1.8.0_121`, instead
use `QZTray/2.0.2 (Linux; 4.4.0-63-generic) Java/1.8.0_121 (Oracle
Corporation)`

Useful for analytics and resolving issues.

see https://tools.ietf.org/html/rfc7231#section-5.5.3

### DIFF
--- a/src/qz/utils/ConnectionUtilities.java
+++ b/src/qz/utils/ConnectionUtilities.java
@@ -1,0 +1,43 @@
+/**
+ * @author Ewan McDougall
+ *
+ * Copyright (C) 2017 Tres Finocchiaro, QZ Industries, LLC
+ *
+ * LGPL 2.1 This is free software.  This software and source code are released under
+ * the "LGPL 2.1 License".  A copy of this license should be distributed with
+ * this software. http://www.gnu.org/licenses/lgpl-2.1.html
+ */
+package qz.utils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLConnection;
+
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import qz.common.Constants;
+
+public final class ConnectionUtilities {
+
+    private static final String USER_AGENT = String.format("%s/%s (%s; %s) Java/%s (%s)",
+        Constants.ABOUT_TITLE.replaceAll("[^a-zA-Z]", ""),
+        StringUtils.join(ArrayUtils.subarray(StringUtils.split(Constants.VERSION, "."), 0, 2), "."),
+        System.getProperty("os.name"),
+        System.getProperty("os.version"),
+        StringUtils.join(ArrayUtils.subarray(StringUtils.split(System.getProperty("java.version"), "."), 0, 2), "."),
+        System.getProperty("java.vendor"));
+
+    /**
+     * Returns an input stream that reads from the URL.
+     * While setting the underlying URLConnections User-Agent.
+     *
+     * @param url an absolute URL giving location of resource to read.
+     */
+    public static InputStream getInputStream(String urlString) throws IOException {
+        URLConnection urlConn = new URL(urlString).openConnection();
+        urlConn.setRequestProperty("User-Agent", USER_AGENT);
+        return urlConn.getInputStream();
+    }
+}
--- a/src/qz/utils/ConnectionUtilities.java
+++ b/src/qz/utils/ConnectionUtilities.java
@@ -21,13 +21,14 @@ import qz.common.Constants;
 
 public final class ConnectionUtilities {
 
-    private static final String USER_AGENT = String.format("%s/%s (%s; %s) Java/%s (%s)",
+    private static final String USER_AGENT = String.format("%s/%s (%s; %s; %s) Java/%s (%s)",
         Constants.ABOUT_TITLE.replaceAll("[^a-zA-Z]", ""),
         StringUtils.join(ArrayUtils.subarray(StringUtils.split(Constants.VERSION, "."), 0, 2), "."),
         System.getProperty("os.name"),
         System.getProperty("os.version"),
+        System.getProperty("os.arch"),
         StringUtils.join(ArrayUtils.subarray(StringUtils.split(System.getProperty("java.version"), "."), 0, 2), "."),
-        System.getProperty("java.vendor"));
+        System.getProperty("java.vm.name"));
 
     /**
      * Returns an input stream that reads from the URL.
--- a/src/qz/auth/CRL.java
+++ b/src/qz/auth/CRL.java
@@ -3,10 +3,11 @@ package qz.auth;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import qz.utils.ConnectionUtilities;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.net.URL;
 import java.util.ArrayList;
 
 /**
@@ -37,8 +38,7 @@ public class CRL {
                 public void run() {
                     log.info("Loading CRL from {}", CRL_URL);
 
-
-                    try(BufferedReader br = new BufferedReader(new InputStreamReader(new URL(CRL_URL).openStream()))) {
+                    try(BufferedReader br = new BufferedReader(new InputStreamReader(ConnectionUtilities.getInputStream(CRL_URL)))) {
                         String line;
                         while((line = br.readLine()) != null) {
                             //Ignore empty and commented lines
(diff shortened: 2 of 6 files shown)


mrloop:
Reworked #171 and opened PR against 2.1 branch. 2.1 branch isn't building at the moment.

src/qz/utils/ConnectionUtilities.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public static InputStream getInputStream(String urlString) throws IOException {
+        URLConnection urlConn = new URL(urlString).openConnection();
+        urlConn.setRequestProperty("User-Agent", userAgent());
+        return urlConn.getInputStream();
+    }
+
+    public static String userAgent() {
+        String.format("%s/%s (%s; %s) Java/%s (%s)",
```
tresf:
```java
return String.format("%s/%s (%s; %s) Java/%s (%s)",
```

src/qz/utils/ConnectionUtilities.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLConnection;
+
+import qz.common.Constants;
+
+public final class ConnectionUtilities {
+    private ConnectionUtilities() {
+    }
+
```
tresf: Since the function name is identical to that of `URLConnection` please add a comment explaining the purpose.

src/qz/utils/ConnectionUtilities.java
```diff
@@ -0,0 +1,29 @@
...
+    public static InputStream getInputStream(String urlString) throws IOException {
+        URLConnection urlConn = new URL(urlString).openConnection();
+        urlConn.setRequestProperty("User-Agent", userAgent());
+        return urlConn.getInputStream();
+    }
+
+    public static String userAgent() {
```
tresf: This should never change once the JVM has been loaded so it should just be a `private static` variable (which would eliminate the `return` problem).

If you want to keep it in a function for cleanliness purposes, you should still store it in a variable to avoid redundant `replaceAll` calculations, etc and the function should be private since it isn't used anywhere else.

e.g. 

```java
private static USER_AGENT = String.format("%s/%s (%s; %s) Java/%s (%s)",
    Constants.ABOUT_TITLE.replaceAll("[^a-zA-Z]", ""),
    Constants.VERSION,
    System.getProperty("os.name"),
    System.getProperty("os.version"),
    System.getProperty("java.version"),
    System.getProperty("java.vendor"));
```

src/qz/utils/ConnectionUtilities.java
```diff
@@ -0,0 +1,29 @@
...
+        String.format("%s/%s (%s; %s) Java/%s (%s)",
+            Constants.ABOUT_TITLE.replaceAll("[^a-zA-Z]", ""),
+            Constants.VERSION,
```
tresf: Since most user-agent strings use `major.minor`, I would recommend truncated off the `.patch` and just having it `QZTray/2.1`.

mrloop:
@tresf thanks for the feedback, have updated.

tresf:
@mrloop this looks perfect.  We will use this PR to perform our testing and we will have it merged as soon as possible.

@klabarge can you please fill out the following:

|         | Chrome User Agent String | QZ Tray User Agent String | 
|------|-------------------------------|-----------------------------|
| Linux + Oracle |  |  |
| Linux + OpenJDK |  |  |
| Mac |  |  |
| Win |  |  |

Once we have collected this information, we'll have further feedback or merge.

src/qz/utils/ConnectionUtilities.java
```diff
@@ -0,0 +1,43 @@
...
+        System.getProperty("java.vendor"));
+
+    /**
+     * Returns an input stream that reads from the URL.
+     * While setting the underlying URLConnections User-Agent.
+     *
+     * @param url an absolute URL giving location of resource to read.
+     */
+    public static InputStream getInputStream(String urlString) throws IOException {
+        URLConnection urlConn = new URL(urlString).openConnection();
```
tresf: @klabarge @lite1979 for testing purposes, to get the string from QZ Tray, using IntelliJ, just add

```java
System.out.println(">>>>>>>>>> " + USER_AGENT);
```

klabarge:
On Ubuntu 16.04, I cannot get QZ Tray User Agent to return OpenJDK (returns `QZTray/2.1 (Linux; 4.4.0-64-generic) Java/1.8 (Oracle Corporation)`).

`java -version` returns OpenJDK
image

`update-alternatives` is set to OpenJDK
image

`$JAVA_HOME` is also set to OpenJDK
image

tresf:
@klabarge Fedora's OpenJDK reports `Oracle Corporation` as well.  I think it's OK.

```bash
java -XshowSettings:properties -version

[...]
    java.runtime.name = OpenJDK Runtime Environment
    java.runtime.version = 1.8.0_45-b13
    java.specification.name = Java Platform API Specification
    java.specification.vendor = Oracle Corporation
    java.specification.version = 1.8
    java.vendor = Oracle Corporation
[...]
```

mrloop:
Maybe the above fixup commit is better it uses `java.vm.name` and also includes the architecture of the OS `os.arch`?

tresf:
> also includes the architecture of the OS `os.arch`?

32-bit JVM on 64-bit Windows shows the following:

| Prop | Value |
|------|--------|
| ` java.vm.name` | `Java HotSpot(TM) Client VM` |
| `os.arch` | `x86` |

64-bit JVM on 64-bit Windows:

| Prop | Value |
|------|--------|
| ` java.vm.name` | `Java HotSpot(TM) 64-Bit Server VM` |
| `os.arch` | `amd64` |

I do like a way of getting OpenJDK vs Oracle since we only raise support issues through Oracle against the official JRE/JDK however this is not a hard-requirement.

What I'd like to spend some more time nitpicking is the formatting of the other parts of the string.  For example, if most websites parse the user agent string for `amd64 | x86_64`, I'd like to make sure we spell that out as consistently as possible.

mrloop:
@tresf I'm unsure what your suggesting in regards to `amd64 | x86_64` my thinking is the User-Agent string should just contain whatever is reported by `os.arch` and it up to the analytics software or developer looking at it to interpret it however they see fit.
 
Here's some examples of `os.arch` values:
http://lopica.sourceforge.net/os.html
http://www.java-gaming.org/topics/possible-values-for-system-property/14110/msg/112286/view.html#msg112286

tresf:
> @tresf I'm unsure what your suggesting in regards to `amd64 | x86_64` my thinking is the User-Agent string should just contain whatever is reported by `os.arch` and it up to the analytics software or developer looking at it to interpret it however they see fit.

I'm trying to determine if the tools out there take the strings into consideration or not.  We should code directly to the standard or else the value has marginal worth.

For example, if tools like Google Analytics or Amazon statistical reporting is already matching on these values, great however if they're not, we should mimic the web browser.  Investigating this would take a bit of insight into the analytics tools out there.

Another potential problem is the difference in OS definition.  If the browser reports `NT 5.1` instead of Java reporting `Windows 8.1` we need to make a decision on what we send.

tresf:
One component all Linux browsers seem to contain consistently is the `X11` flag, which appears to be best derived from the Graphics Environment that Java is using:

| OS | `java.awt.graphicsenv` |
|-----|----------------------------|
| mac | `sun.awt.CGraphicsEnvironment` (C=Carbon) | 
| win | ` sun.awt.Win32GraphicsEnvironment` |
| linux | `sun.awt.X11GraphicsEnvironment` | 

or optionally, the awt toolkit value:

| OS | `awt.toolkit` |
|-----|---------------|
| mac | `sun.lwawt.macosx.LWCToolkit` |
| win | `sun.awt.windows.WToolkit` |
| linux | `sun.awt.X11.XToolkit` |

... hopefully these are all still accessible in headless mode.

Also, Ubuntu is explicitly mentioned.  I don't see the Linux Kernel version being mentioned in any other strings.  We had to use a [custom script for this](https://github.com/qzind/tray/blob/2.0/src/qz/utils/SystemUtilities.java#L121).  I fear this will get gross quickly. 

Although this all probably sounds overkill for a User Agent string, if we don't it will product [strange results by the analytics engine]((https://cloud.githubusercontent.com/assets/6345473/23637719/4df343da-02ab-11e7-9674-6982386cf48b.png).  If we're going to do this, we should do our best to do it right.
.

tresf:
After looking at several user agents on several OSs, I've decided to follow suit with the `Mozilla/5.0` prefix as well as convert as many strings as possible to the native OS equivalent.  This is being tracked at #177.  Feedback welcome.

Closing as #177 now supersedes this PR.

PR #176, opened 2017-03-05
Tighten up 'processPrintRequest' a little

dsanders11:
`data` is not optional in `processPrintRequest` and `releasePrintProcessor` should always execute no matter what.

### DIFF
--- a/src/qz/utils/PrintingUtilities.java
+++ b/src/qz/utils/PrintingUtilities.java
@@ -155,7 +155,7 @@ public class PrintingUtilities {
             PrintOutput output = new PrintOutput(params.optJSONObject("printer"));
             PrintOptions options = new PrintOptions(params.optJSONObject("options"), output);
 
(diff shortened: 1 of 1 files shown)


tresf:
> `data` is not optional in `processPrintRequest`

True, but that should already be handled logically here: https://github.com/dsanders11/tray/blob/61e089124661f15b624266c6265ca06d26a7aa73/src/qz/utils/PrintingUtilities.java#L151

> `releasePrintProcessor()` should always execute no matter what.

I think it does.  If `data` threw an exception it would have never reached `getPrintProcessor(...)` to begin with, since it's nested it should have thrown it up the stack and never gotten a `PrintProcessor`.  Everything inside the `try` is being caught by the `Exception` clause.

Have you run into scenarios where this logic has helped?  Can you explain the use case?

tresf:
Valid points.  Merging.

PR #179, opened 2017-03-10
ZPL Image width needs to be divisible by 8, enforce this like EPL

robraux:
I investigated the option of performing ZB64 Encoding and Compression, however it only applies to ZPL II, not ZPL. This change is sufficient and is what others do when utilizing ````^GFA```` tag.

This was tested with both a raw PNG, as well as base64 ````data:image/png;base64,```` to ensure there were no discrepancies.

### DIFF
--- a/src/qz/printer/LanguageType.java
+++ b/src/qz/printer/LanguageType.java
@@ -20,8 +20,8 @@ import java.util.List;
  */
 public enum LanguageType {
 
-    ZPLII(false, false, "ZEBRA", "ZPL2"),
-    ZPL(false, false),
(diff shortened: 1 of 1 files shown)


tresf:
@robraux thanks!

robraux:
@tresf thank you for the merge. When will 2.0.4 be tagged and packaged which should include this release? I'm eager to get this deployed on the official channel for some international shipping I'm doing.

tresf:
>  When will 2.0.4 be tagged and packaged

Our release cycles generally have a few months between them due to the regression testing we perform on each release.  If you're a premium subscriber, contact us at support@qz.io and we'll make sure you get a packaged release prior.

PR #190, opened 2017-04-16
R signing example

carlganz:
I have been working on using tray with R on the server, so I figured I would share a basic signing example in R.


tresf:
@carlganz much thanks.  Are indenting differences between lines 35 and 40 intentional?

carlganz:
Yes that works.

I used a code formatter that made some dubious decisions.

PR #202, opened 2017-06-29
Apply fixes to PDF printing and orientation

tresf:
Cherry-picked out of 2.0 branch so we can fast track a 2.0.4 release with Firefox fixes.


bberenz:
Branch `reverse-fix` contains all commits from `pdffix`, which is under review for merging via #255, so this is request is not needed anymore.

PR #203, opened 2017-06-30
Allow exclusion of demo assets using a build property

onurguzel:
I have added a build property to optionally exclude demo files from installers. These files are not needed on most of client computers and excluding them reduces install file size by ~6%, which makes a difference when deploying to lots of computers over slow networks.

The default is to include demo files, so I believe it is an acceptable feature.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -141,7 +141,7 @@
         <property name="build.type" value="-community"/>
     </target>
 
-    <target name="include-assets" depends="init,get-version" if="${dist.include_demo}">
+    <target name="include-assets" depends="init,get-version" unless="dist.minimal">
         <echo>Copying resource files to output</echo>
 
         <copy todir="${dist.dir}/${demo.dir}/${asset.dir}">
--- a/build.xml
+++ b/build.xml
@@ -58,7 +58,6 @@
             <fileset dir="${src.dir}" excludes="**/*.java"/>
         </copy>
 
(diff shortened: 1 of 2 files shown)


tresf:
@onurguzel I would favor `dist.minimal=false` and then set it to true for stripping unwanted assets.  I believe it then could be removed from the properties file as well (this may or may not be desired) and simply added to the compiling documentation.

onurguzel:
> @onurguzel I would favor `dist.minimal=false` and then set it to true for stripping unwanted assets. I believe it then could be removed from the properties file as well (this may or may not be desired) and simply added to the compiling documentation.

If the changes in `build.xml` is OK for you, I can rename the property.

build.xml
```diff
@@ -57,6 +57,11 @@
         <copy todir="${build.project.dir}">
             <fileset dir="${src.dir}" excludes="**/*.java"/>
         </copy>
+
+        <!-- Distribute socket with LICENSE only -->
```
tresf: This comment is left over from the [`qz-print`](https://github.com/qzind/qz-print/blob/1.9/build.xml#L248-L256) split applet/socket builds and can now be removed.

tresf:
> If the changes in build.xml is OK for you, I can rename the property.

Yes, the `LICENSE.TXT` can be moved into the compilation target for now.

onurguzel:
@tresf I've made the changes, you can review again.

Issue #208, opened 2017-07-05
Don't Entirely Replace Properties File When Replacing SSL Cert


dsanders11:
Currently if you have to update the SSL certificate on an existing QZ Tray installation, the process will entirely replace the `qz-tray.properties` file. This is not ideal because any custom settings like `authcert.override` will be lost.

Would it be possible for it to reuse the keystore password for the new keystore? I believe this is the only reason it needs to replace that file, to update the keystore password.

We're using LetsEncrypt certificates so they need to be replaced every 3 months, and I always seem to forget to save off the custom settings before they get lost. :-) It's especially painful because we're running a fork that adds a couple extra settings to the properties file as well.

dsanders11:
@tresf, FYI, some time in the next 3 months (before the next LetsEncrypt cert expiration, heh) I'll probably try to fully integrate the renewal process into QZ Tray itself to make it less painful and reduce the [bus factor](https://en.wikipedia.org/wiki/Bus_factor).

I'm thinking an interface where QZ Tray will check the expiration on the certificate, and a few days before expiration it'll try to renew (and log errors if it has problems) and replace the keystore while running (might be a little tricky).

To get the new certificate I'm thinking a web service (since LetsEncrypt needs public access) running on a server, which QZ Tray will contact (with username and password stored in `qz-tray.properties` for an extra layer of security) and ask the server to renew the certificate for it, providing the certificate and a signature to prove it has the private key. The server will renew the certificate (probably shell out to run `certbot` as they don't seem to have a direct API you can use) and hand the new info back to QZ Tray. Only trickiness there is the process might take a little bit, so need a clean API for QZ Tray to check for the result in case it gets disconnected while waiting and won't restart the whole renewal process.

Any thoughts or pre-requisite work towards that end would be appreciated.

tresf:
This seems like an involved process for something that `certbot` could simply farm off to a post-hook, which kills the Java process and relaunches.

The standard use-case is to for qz-tray simply randomize the private key password when it is converted from `pem` to `p12` and then to `jks`.  We could certainly check for existence of the `authcert.override` when rewriting the properties file on a Trusted SSL install, but making Java responsible for rewing the certificate seems like the wrong place and the password logic should cover 80% of use-cases which is what we aim to meet.

tresf:
[...]

PR #275, opened 2018-02-07
Make reload more friendly: reload properties and keystore

dsanders11:
Pull request as mentioned in #208. To answer the question there, `StatisticsHandler` is used because it is required in order to use `Server.setStopTimeout`, as per [the documentation](https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/server/Server.html#setStopTimeout-long-).

The main change is to move the reload logic to `PrintSocketServer.reloadServer()` so that it can easily be called programmatically from other parts of the code that may need to reload QZ. This may have other potential uses I hadn't thought of, such as if `wss.host` is `0.0.0.0` maybe QZ would automatically reload on a network change so it could bind on any new network interface (not sure if it does that already, or if `0.0.0.0` only binds once on all network interfaces available when it is called).

### DIFF
--- a/src/qz/common/TrayManager.java
+++ b/src/qz/common/TrayManager.java
@@ -332,6 +332,11 @@ public class TrayManager {
 
     public void exit(int returnCode) {
         prefs.save();
+        try {
+            PrintSocketServer.stopServer();
+        } catch (Exception e) {
+            log.error("Error while stopping server", e);
+        }
         System.exit(returnCode);
     }
 
@@ -437,11 +442,9 @@ public class TrayManager {
      * Sets the WebSocket Server instance for displaying port information and restarting the server
      *
      * @param server            The Server instance contain to bind the reload action to
-     * @param running           Object used to notify PrintSocket to reiterate its main while loop
-     * @param securePortIndex   Object used to notify PrintSocket to reset its port array counter
      * @param insecurePortIndex Object used to notify PrintSocket to reset its port array counter
      */
-    public void setServer(final Server server, final AtomicBoolean running, final AtomicInteger securePortIndex, final AtomicInteger insecurePortIndex) {
+    public void setServer(final Server server, final AtomicInteger insecurePortIndex) {
         if (server != null && server.getConnectors().length > 0) {
             singleInstanceCheck(PrintSocketServer.INSECURE_PORTS, insecurePortIndex.get());
 
@@ -454,11 +457,7 @@ public class TrayManager {
                 public void run() {
                     try {
                         setDangerIcon();
-                        running.set(false);
-                        securePortIndex.set(0);
-                        insecurePortIndex.set(0);
-
-                        server.stop();
+                        PrintSocketServer.reloadServer();
                     }
                     catch(Exception e) {
                         displayErrorMessage("Error stopping print socket: " + e.getLocalizedMessage());
--- a/src/qz/ws/PrintSocketServer.java
+++ b/src/qz/ws/PrintSocketServer.java
@@ -17,6 +17,7 @@ import org.apache.log4j.rolling.RollingFileAppender;
 import org.apache.log4j.rolling.SizeBasedTriggeringPolicy;
 import org.eclipse.jetty.http.HttpVersion;
 import org.eclipse.jetty.server.*;
+import org.eclipse.jetty.server.handler.StatisticsHandler;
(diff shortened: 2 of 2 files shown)


src/qz/ws/PrintSocketServer.java
```diff
@@ -49,8 +50,13 @@
     public static final List<Integer> INSECURE_PORTS = Collections.unmodifiableList(Arrays.asList(8182, 8283, 8384, 8485));
 
 
-    private static TrayManager trayManager;
-    private static Properties trayProperties;
+    private static TrayManager trayManager = null;
+    private static Properties trayProperties = null;
+    private static Server server = null;
+
+    private static final AtomicBoolean running = new AtomicBoolean(true);
```
bberenz: Why has `running` changed to be true by default?
The `runServer` loop appears only loop on a `true` state for this value now too (L138), which means it will always loop through every port in the lists when ran. We want this to break on the first successful connection.

src/qz/ws/PrintSocketServer.java
```diff
@@ -154,15 +165,23 @@ public void configure(WebSocketServletFactory factory) {
...
                 server.setStopAtShutdown(true);
                 server.start();
 
                 running.set(true);
                 trayManager.setServer(server, running, securePortIndex, insecurePortIndex);
                 log.info("Server started on port(s) " + TrayManager.getPorts(server));
 
+
                 server.join();
+                while (!server.isStopped()) {
```
bberenz: This added loop doesn't seem necessary, as from what I can gleam from `server.join` it takes care of exactly what this loop has been tasked with doing.

src/qz/common/TrayManager.java
```diff
@@ -452,11 +452,7 @@ public void setServer(final Server server, final AtomicBoolean running, final At
                 public void run() {
                     try {
                         setDangerIcon();
-                        running.set(false);
-                        securePortIndex.set(0);
-                        insecurePortIndex.set(0);
```
bberenz: These are passed in as parameters and only used in this one spot, so if they aren't being used here anymore, we should probably clean up the method's signature too.

src/qz/ws/PrintSocketServer.java
dsanders11: It is necessary. I know, it doesn't quite make sense on face value, but without it you get:

```java
java.lang.IllegalStateException: !STOPPED
	at org.eclipse.jetty.server.handler.HandlerWrapper.destroy(HandlerWrapper.java:131)
	at qz.ws.PrintSocketServer.runServer(Unknown Source)
	at qz.ws.PrintSocketServer.main(Unknown Source)
```

There's also a `server.isStopping()` method. I believe `server.join()` will return the moment you call `server.stop()`, but internally the state is `STOPPING`, not `STOPPED`. So we have to wait until it is fully stopped before calling `server.destroy()`.

src/qz/ws/PrintSocketServer.java
dsanders11: > which means it will always loop through every port in the lists when ran

Actually, it does not, it stops after it finds one, because `server.join()` blocks. If that wasn't true the program would simply exit once `running` was set to true since the loop would stop and it would exit.

> Why has running changed to be true by default?

For reload to work the loop needs to run again when reload is called. Running refers to the overall state of the program, so as long as it's running, if the server is stopped, start it up again.

src/qz/common/TrayManager.java
dsanders11: Addressed this. `insecurePortIndex` was still required due to `singleInstanceCheck`, but some refactoring might work there for it to not be necessary, instead using `getInsecurePortInUse()` and dropping `singleInstanceCheck` entirely.

But, that's outside the scope of this PR.

dsanders11:
@bberenz, I've made some changes, addressed your comments, and based it off of the latest `2.0` branch.

tresf:
Re-milestoning for 2.1 branch.  No need to rebase at this point, we have some fast forwarding to do first but we'll eventually need to rebase against 2.1.

tresf:
Re-milestoning for 2.2 branch.

tresf:
FYI, due to some changes in how macOS Catalina handles SSL certs, this will be re-milestoned for 2.1.  Related #491.  Will close and supercede once the new PR is open.

PR #290, opened 2018-03-26
i18n support

tresf:


liushuyu:
I have marked more strings when something occurred to me:
When you make some actions like access the serial port, it will produce the following log message if locale is `zh_CN`:
```
[INFO] 2018-03-26 16:53:11,635 @ qz.common.TrayManager:?
        Allowed localhost to 访问串口
```
Which may make technical support difficult, since you cannot understand every language in this world. I am thinking maybe you could use the "action code" in the first element of `qz/ws/PrintSocketClient/Method` and the log may become:

```
Allowed localhost to serial.findPorts
```

liushuyu:
New changes:

- Added a way to adjust locale in (`qz.common.I18NLoader`):
```java
public static void changeLocale(Locale locale);
```

- More UI strings (?) are translatable now

tresf:
Superseded by #290.

liushuyu:
>Superseded by #290.

@tresf I think you mean #303 ?

tresf:
:) yes!

PR #178, opened 2017-03-07
File communication interface.

mrloop:
Allow qz-tray to read/write from files.
Writing to a file will either overwrite or append to contents of a file
depending on configuration.
Reading a file involves registering for notifications of file
modifications and recieving file contents in notification of file
modification.

For simplicity of implementation a FileWatcher is created for each File
that modification notifications are wanted for.

PR #306, opened 2018-04-11
File communication interface

tresf:
I'll quote @mrloop:

> Allow qz-tray to read/write from files.
> Writing to a file will either overwrite or append to contents of a file
> depending on configuration.
> Reading a file involves registering for notifications of file
> modifications and receiving file contents in notification of file
> modification.
> 
> For simplicity of implementation a FileWatcher is created for each File
> that modification notifications are wanted for.

Example:

Write:
```js
qz.file.write('test.txt', { data: 'hello world' });
```

Read:
```js
qz.file.read('test.txt').then(function(data) { console.log(data); });
```

By default, data is written to a sandboxed location, based on the certificate `CN` (Common Name) parameter.  This can be overridden by providing `{ sandbox: false }`.

Supersedes #178.


src/qz/utils/FileUtilities.java
```diff
@@ -61,25 +75,177 @@
...
 
+    private static void populateWhiteList() {
+        whiteList = new ArrayList<>();
+        //default sandbox locations. More can be added through the properties file
+        whiteList.add(new Pair<>(Paths.get(SystemUtilities.getDataDirectory()), "|sandbox|"));
+        whiteList.add(new Pair<>(Paths.get(SystemUtilities.getSharedDataDirectory()), "|sandbox|"));
+
+        Properties props = PrintSocketServer.getTrayProperties();
+        if (props != null) {
+            StringBuilder propString = new StringBuilder(props.getProperty("file.whitelist"));
```
tresf: This can throw a NPE and needs a guard.  Probably just a `props.getProperty("file.whitelist", "")` would suffice.

src/qz/ws/PrintSocketClient.java
```diff
@@ -461,6 +475,97 @@ private void processMessage(Session session, JSONObject json, SocketConnection c
...
+            case FILE_LIST: {
+                Path absPath = FileUtilities.getAbsolutePath(params, shownCertificate, true);
+
+                if (Files.exists(absPath)) {
+                    if (Files.isDirectory(absPath)) {
+                        ArrayList<String> files = new ArrayList<>();
+                        Files.list(absPath).forEach(file -> files.add(file.getFileName().toString()));
+                        sendResult(session, UID, new JSONArray(files));
+                    } else {
+                        sendError(session, UID, "Path is not a directory");
```
tresf: The errors aren't logged to the console at all which leaves little debugging information.  We should have something rather verbose exposed to the log file when an error occurs, including the offending `absPath`.

src/qz/ws/PrintSocketClient.java
```diff
@@ -461,6 +475,97 @@ private void processMessage(Session session, JSONObject json, SocketConnection c
...
+                    }
+                } else {
+                    sendError(session, UID, "Path does not exist");
```
tresf: Can we do the equivalent of trying a `mkdir` here and then loop the call back onto itself?  That we aren't throwing errors right off the start.

sample.html
```diff
@@ -827,47 +838,167 @@ <h4 class="panel-title">Options</h4>
...
+                        <div class="btn-toolbar">
+                            <button type="button" class="btn btn-info" onclick="listFiles();">List Files</button>
+                            <div class="btn-group">
+                                <button type="button" class="btn btn-info" onclick="readFile();">Read File</button>
+                                <button type="button" class="btn btn-success" onclick="writeFile();">Write File</button>
+                                <button type="button" class="btn btn-warning" onclick="deleteFile();">Delete File</button>
+                            </div>
+
+                            <div class="btn-group">
+                                <button type="button" class="btn btn-default" onclick="startFileListen()">Listen for Events</button>
```
tresf: Can we make it more clear that this is `Folder` events?

src/qz/utils/FileWatcher.java
```diff
@@ -0,0 +1,132 @@
...
+    private synchronized static String getLines(Path path, FileIO listener) throws IOException {
+        StringBuilder sb = new StringBuilder();
+        String buffer;
+
+        if (listener.isReversed()) {
+            try(ReversedLinesFileReader reader = new ReversedLinesFileReader(path.toFile())) {
+                int count = 0;
+                while((buffer = reader.readLine()) != null && count++ != listener.getLines()) {
+                    // Warning, this will strip "\r" from the data
+                    sb.append(buffer).append("\n");
```
tresf: After testing out `reversed`, I can't see a logical or intuitive reason to return the lines in reverse order.  Thoughts welcome, but I think if you're asking for the last 10 lines of a file, they should still be in the natural order.

For example:

First 10:
```
test1
test2
test3
test4
test5
test6
test7
test8
test9
test10
```

Last 10:
```
test7
test8
test9
test10
test11
test12
test13
test14
test15
test16
```

src/qz/ws/PrintSocketClient.java
bberenz: This call is just listing the files though. Do we really want a write operation on that?

tresf: My concern is that the default folder isn't available and throws a false error and that's non-sane defaults.  Perhaps we swallow the exception if it's the root-most directory and send back a black list of files?  I really don't like someone clicking that "list files" and have the software yell at them about a folder they didn't know should exist.  It seems unintuitive to need to rely on the file write to get the parent folder.  Thoughts are welcome here.

bberenz: You're right, I forgot about the default folder on clean runs.
I've altered the code so that if we are performing a fileio operation, it will make sure at least that folder always exists first

src/qz/ws/PrintSocketClient.java
```diff
@@ -461,6 +475,106 @@ private void processMessage(Session session, JSONObject json, SocketConnection c
...
+                }
+
+                break;
+            }
+            case FILE_WRITE: {
+                FileParams fileParams = new FileParams(params);
+                Path absPath = FileUtilities.getAbsolutePath(params, shownCertificate, false);
+
+                if (Files.isWritable(absPath)) {
```
tresf: Some order of operations issue is happening here... this check fails when I try shared: false, sandbox: true and I can't write a file.

bberenz: I'm an idiot. I was trying to remember why I left this check out, thought I tested it with it in, guess not. Files that don't exist can't pass the "isWritable" check.

tresf:
There seems to be a bug with `stopListening` where it will throw an AccessDenied exception...

```
[DEBUG] 2018-07-10 11:27:29,496 @ qz.ws.PrintSocketClient:176
	Message: {"call":"file.stopListening","promise":{},"params":{"sandbox":true,"shared":false,"path":""},"timestamp":1531236449495,"uid":"8pwba7"}
[WARN] 2018-07-10 11:27:29,496 @ qz.ws.PrintSocketClient:215
	Bad signature on request
[ERROR] 2018-07-10 11:27:29,497 @ qz.ws.PrintSocketClient:230
	FileIO exception occurred
java.nio.file.AccessDeniedException: C:\Users\tresf\AppData\Roaming\qz\sandbox\UNTRUSTED
	at qz.utils.FileUtilities.getAbsolutePath(FileUtilities.java:105)
	at qz.ws.PrintSocketClient.processMessage(PrintSocketClient.java:499)
	at qz.ws.PrintSocketClient.onMessage(PrintSocketClient.java:223)
	at sun.reflect.GeneratedMethodAccessor10.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.jetty.websocket.common.events.annotated.CallableMethod.call(CallableMethod.java:70)
	at org.eclipse.jetty.websocket.common.events.annotated.OptionalSessionCallableMethod.call(OptionalSessionCallableMethod.java:68)
	at org.eclipse.jetty.websocket.common.events.JettyAnnotatedEventDriver$2.run(JettyAnnotatedEventDriver.java:210)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555)
	at java.lang.Thread.run(Thread.java:748)
```

bberenz:
My fault on simplifying the close call path, 2aa134f6ca7f73d5d245fca9abd34a6d2c238825 should take care of it.

PR #315, opened 2018-04-26
Update license file

klabarge:
Removes:
* qz-print
* PDF Renderer

Adds:
* jsemver
* hid4java
* Apache log4j
* MIT License

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -25,6 +25,12 @@ jetty Web Server Library (As-Is, No Modifications)
   http://eclipse.org/jetty/
 
 
+Apache log4j (As-Is, No Modifications)
+  Copyright (C) 1999-2005 The Apache Software Foundation
+  Apache License, Version 2.0 (attached), with Copyright Notice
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -25,22 +25,39 @@ jetty Web Server Library (As-Is, No Modifications)
...
+
+
 Apache PDFBox (As-Is, No Modifications)
   Copyright (C) 2009–2015 The Apache Software Foundation
   Apache License, Version 2.0 (attached), with Copyright Notice
   https://pdfbox.apache.org/
 
 
 jSSC Library (As-Is, No Modifications)
-  Copyright (c) 2010-2013 Alexey Sokolov (scream3r)
```
tresf: This whitespace change is OK, but let's be careful when we're using stuff like Atom text editor, it'll do whitespace cleanup all over the files that we may not have asked for. :)

klabarge: Noted 🤦‍♂️

PR #334, opened 2018-06-12
Add HTTP about page

tresf:
Provides HTTP about page at https://localhost:8181 and JSON formatted page at https://localhost:8181/json.


dsanders11:
@tresf, is there any way to use a filter so that the `/json` endpoint isn't necessary, and instead an `Accept` header with `application/json` would return the JSON? Seems like it would fit more nicely with the main endpoint being usable as a WebSocket or HTML page. But, it would preclude being able to view the JSON in a web browser easily (and takes a little bit of extra config if you want to use something like CURL). Just my 2 cents, since `/json` is a somewhat cryptic endpoint

tresf:
> is there any way to use a filter so that the `/json` endpoint isn't necessary, and instead an Accept header with `application/json` would return the JSON

Sure!  This is a great idea and we hadn't even considered parsing the HTTP headers but it should be doable.  We only have one client live with this, so we'll have to ask them if if they'll need the old endpoint still.

> `/json` is a somewhat cryptic endpoint

I think you're embellishing a bit 😄, but I do agree that specifying `xhr.responseType='json'` should be something we can capture Java-side.

dsanders11:
> I think you're embellishing a bit 😄

Perhaps, but I'm just imagining if the use of HTTP endpoints was expanded in the future, it would feel out of place. For example, there could be `/stats` for things like print count, last print, up time, etc. There could be `/console` which could be the `sample.html` page tailored to just that QZ Tray instance (would need some thinking through, maybe a `qz-tray.properties` key for listing the cert/signing URLs). Could be `/config` for access to a subset of the `qz-tray.properties` keys (would be more useful if plugins land). So, with those kinds of possibilities in mind, a `/json` endpoint would be out of place and kinda cryptic. ;)

tresf:
Can you open up a separate issue for this enhancement?  We'll slate it for 2.0.8 if possible but I don't want to lose track of it.

PR #362, opened 2018-10-11
Send data over serial from url

### LINKED ISSUE #296: Allow URL for serial data
Currently, serial data is sent through a `String` which is ok for small payloads but does not scale well to larger payloads.

The ability to send serial data from a URL or local file path is desired for certain serial devices to allow a large data stream to be sent without too much chatter over the websocket channel.

Vzor-:
Closes #296

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1166,7 +1166,7 @@ var qz = (function() {
              *  @param {string} [properties.stopBits=1]
              *  @param {string} [properties.parity='NONE'] Valid values <code>[NONE| EVEN | ODD | MARK | SPACE]</code>
              *  @param {string} [properties.flowControl='NONE'] Valid values <code>[NONE | XONXOFF_OUT | XONXOFF_IN | RTSCTS_OUT | RTSCTS_IN]</code>
-             *  @param {string} [properties.type='RAW'] Valid values <code>[RAW | URL]</code>
+             *  @param {string} [properties.type='PLAIN'] Valid values <code>[PLAIN | FILE]</code>
              *
              * @returns {Promise<null|Error>}
              *
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1159,14 +1159,15 @@ var qz = (function() {
              * Any responses from the device will be sent to serial callback functions.
              *
              * @param {string} port An open port to send data over.
-             * @param {string} data The data to send to the serial device.
+             * @param {Object} data The object containing data and metadata to send to the serial device.
+             *  @param {string} [data.type='PLAIN'] Valid values <code>[PLAIN | FILE]</code>
+             *  @param {string} data.data The data to send to the serial device.
              * @param {Object} [properties] Properties of data being sent over the serial port.
              *  @param {string} [properties.baudRate=9600]
              *  @param {string} [properties.dataBits=8]
              *  @param {string} [properties.stopBits=1]
              *  @param {string} [properties.parity='NONE'] Valid values <code>[NONE| EVEN | ODD | MARK | SPACE]</code>
              *  @param {string} [properties.flowControl='NONE'] Valid values <code>[NONE | XONXOFF_OUT | XONXOFF_IN | RTSCTS_OUT | RTSCTS_IN]</code>
-             *  @param {string} [properties.type='PLAIN'] Valid values <code>[PLAIN | FILE]</code>
              *
              * @returns {Promise<null|Error>}
              *
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1159,7 +1159,7 @@ var qz = (function() {
              * Any responses from the device will be sent to serial callback functions.
              *
              * @param {string} port An open port to send data over.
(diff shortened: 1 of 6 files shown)


src/qz/communication/SerialProperties.java
```diff
@@ -72,4 +82,8 @@ public int getFlowControl() {
         return flowControl;
     }
 
+    public boolean isFile() {
+        return type == SerialDataType.FILE;
+    }
+
```
bberenz: Everything up to this point seems geared towards allowing more than just the two types, which is fine, but it means this has to be a `getType` to avoid having to refactor later if a new type is introduced. 
If we aren't expecting to add more types though, it would probably be better to refactor the enum & type variable out into a boolean "`fromFile`" value to return here.

sample.html
```diff
@@ -1641,6 +1655,7 @@ <h4 class="panel-title">Options</h4>
     function sendSerialData() {
         var properties = {
             baudRate: $("#serialBaud").val(),
+            type: $("#serialPlainRadio").prop("checked") ? "PLAIN" : "FILE",
```
bberenz: Radio type input can have a `value=""` attribute on them, so we can pull these directly with a `$("input[name='serialType']:checked").val()` call instead of having to use a ternary like this, especially if we need more types later.

tresf:
@bberenz @Vzor- shouldn't the data type be part of the data, not the connection properties?

sample.html
```diff
@@ -1641,12 +1655,12 @@ <h4 class="panel-title">Options</h4>
     function sendSerialData() {
         var properties = {
             baudRate: $("#serialBaud").val(),
+            type: $("input[name='serialType']:checked").val(),
```
tresf: As discussed with @bberenz, this should be part of the data object, not the connection properties.

Vzor-: on it

js/qz-tray.js
```diff
@@ -1159,7 +1159,9 @@ var qz = (function() {
              * Any responses from the device will be sent to serial callback functions.
              *
              * @param {string} port An open port to send data over.
-             * @param {string} data The data to send to the serial device.
+             * @param {Object} data The object containing data and metadata to send to the serial device.
```
bberenz: We still need to allow for existing "string" data types too, so data type documented here needs to be `{string | Object}`

src/qz/ws/PrintSocketClient.java
```diff
@@ -324,7 +324,8 @@ private void processMessage(Session session, JSONObject json, SocketConnection c
                 SerialProperties props = new SerialProperties(params.optJSONObject("properties"));
                 SerialIO serial = connection.getSerialPort(params.optString("port"));
                 if (serial != null) {
-                    serial.sendData(props, params.optString("data"));
+                    JSONObject data = params.optJSONObject("data");
+                    serial.sendData(props, SerialUtilities.getDataTypeJSON(data), data.optString("data"));
```
bberenz: Accounting for string types here - if optJSONObject returns null, we need to pull the command from params[data] as a string (like before) instead of from data[data]

sample.html
```diff
@@ -1646,8 +1660,11 @@ <h4 class="panel-title">Options</h4>
...
             flowControl: $("#serialFlow").val()
         };
-
-        qz.serial.sendData($("#serialPort").val(), $("#serialCmd").val(), properties).catch(displayError);
+        var serialData = {
+            type: $("input[name='serialType']:checked").val(),
+            data: $("#serialCmd").val()
+        };
+        qz.serial.sendData($("#serialPort").val(), serialData, properties).catch(displayError);
     }
```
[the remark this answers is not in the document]
bberenz: this is right, but we still have to maintain backwards compatibility - see new review

bberenz:
No, not really. The problem is that if users running the latest java version visit a site using the older javascript, suddenly it doesn't work anymore. We can encourage use of the newer format in our guides, but we can't expect what's already out there to be updated.

js/qz-tray.js
```diff
@@ -1159,7 +1159,7 @@ var qz = (function() {
              * Any responses from the device will be sent to serial callback functions.
              *
              * @param {string} port An open port to send data over.
-             * @param {Object} data The object containing data and metadata to send to the serial device.
+             * @param {string|Object} data The object containing data and metadata to send to the serial device.
```
tresf: Per @bberenz's comments, the fix should be in Java, but the API illustrating both will stay here in JS.

bberenz:
Latest commit covers the last issue I had, so if it looks good on your end @tresf, I think we can merge this in.

tresf:
Tested:
* ✅ **PASS** String: `var serialData = 'W\n';`
* ✅ **PASS** Object: `var serialData = { type: 'PLAIN', data: 'W\n' };`
* ✅ **PASS** Object: `var serialData = { type: 'FILE', data: 'file:///Users/jdoe/test1.txt' };`

Important note, since 3711308 this API has changed slightly in 2.0 and 2.1's serial support has regressions that need to be resolved with a fast-forward.  The regressions must be addressed prior to the 2.1 release.

PR #374, opened 2018-11-02
Dock Fix

### LINKED ISSUE #367: Hide dock icon on MacOS
Placeholder to hide the Dock icon on MacOS.

Proof of concept:

```bash
sudo sed -i -e 's/-jar/-jar -Dapple.awt.UIElement="true"/g' /Applications/QZ\ Tray.app/Contents/MacOS/QZ\ Tray
```

The issue is that the dialog focus never properly raise it to the foreground which I believe can be worked around using AppleScript: https://stackoverflow.com/a/15426739/3196753

Vzor-:
closes #367

### DIFF
--- a/ant/apple/apple-launcher.sh.in
+++ b/ant/apple/apple-launcher.sh.in
@@ -22,9 +22,9 @@ fallback=$?
 
 # Fallback on Internet Plug-Ins version if needed
 if [ $fallback -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


klabarge:
Tested, this seems to work great on macOS Mojave 10.14 (beta)

All dialogues seems to work as normal when QZ Tray is removed from the dock.

PR #381, opened 2018-11-15
Remove JavaFX dependencies from PrintOptions

### LINKED ISSUE #380: PrintOptions shouldn't rely on JavaFX
For systems without JavaFX installed, `PrintOptions` prevents any printing at all because it uses a class that is only available with JavaFX.

```java
import com.sun.javafx.print.Units;
import javafx.print.PageOrientation;
import javafx.print.PrintColor;
```

Can we find a way to support page orientation and print color without the class errors?  An example is Ubuntu system using `openjdk` but WITHOUT `openjfx` will fail to print.

tresf:
Closes #380

### DIFF
--- a/src/qz/printer/PrintOptions.java
+++ b/src/qz/printer/PrintOptions.java
@@ -1,8 +1,5 @@
 package qz.printer;
 
-import com.sun.javafx.print.Units;
-import javafx.print.PageOrientation;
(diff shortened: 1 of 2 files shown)


tresf:
Done, switched to `Map.Entry`.  @Vzor- can you double-check I didn't break anything with the File IO code?

PR #387, opened 2018-11-28
Use URL instead of constructing link element

dsanders11:
@tresf, just a suggestion, I haven't tested it for cross-browser compatibility on older browsers, but creating a temporary element is a bit icky, so this feels cleaner.


tresf:
IE11 doesn't support `URL()` and the existing logic works fine.  I see no advantage in using `URL()` except that you personally find it "icky". :)

PR #391, opened 2018-12-04
Gracefully handle certificate renewals

lukas-w:
This adds support for reading the renewal information from `digital-certificate.txt` that will be generated by the website. The renewal information contains a signature that is verified using the intermediate certificate's public key. If a certificate is a renewal and the previous certificate is whitelisted, the renewal will be whitelisted automatically. Same goes for blacklists.

Unfortunately, because of the way certificates used to be parsed, the new certificates are not backwards compatible. This means that certificates containing renewal information will be recognised as untrusted by Tray versions that don't contain this PR's patch f66606b. The current implementation is not able to ignore additional data at the end of the file, so parsing the intermediate certificate will fail in this case. I'm not sure about the severity of this. Is a warning on the website's certificate generation page sufficient? Something along the lines of "Be sure to update QZ Tray to version 2.0.9 or later to use any newly generated certificates".

### DIFF
--- a/src/qz/auth/Certificate.java
+++ b/src/qz/auth/Certificate.java
@@ -21,15 +21,13 @@ import javax.security.cert.CertificateParsingException;
 import java.io.*;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.security.Signature;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Properties;
+import java.util.*;
 
 /**
  * Created by Steven on 1/27/2015. Package: qz.auth Project: qz-print
@@ -195,7 +193,17 @@ public class Certificate {
                     log.warn("Failed to retrieve QZ CRL, skipping CRL check");
                 }
             }
+
+            split = split[split.length-1].split(X509Constants.RENEWAL_INFO);
+            if (split.length == 2 && theIntermediateCertificate != null) {
+                try {
+                    readRenewalInfo(split[1], theIntermediateCertificate);
+                } catch (Exception e) {
+                    log.error("Error reading certificate renewal info", e);
+                }
+            }
         }
+
         catch(Exception e) {
             CertificateParsingException certificateParsingException = new CertificateParsingException();
             certificateParsingException.initCause(e);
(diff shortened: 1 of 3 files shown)


tresf:
> If a certificate is a renewal and the previous certificate is whitelisted, the renewal will be whitelisted automatically. Same goes for blacklists.

Clarification... are you saying you blacklist if the previous cert was blacklisted?  I'd like this logic removed.  Sometimes when a client clicks "block" it's because the certificate was renewed but a caching issue is serving up the old cert and people get impatient and click "Always block".  Once available, the new cert shouldn't be blocked as it will cause more support issues.  Raising the dialog is fine, which would mimic the behavior today.

lukas-w:
> are you saying you blacklist if the previous cert was blacklisted? I'd like this logic removed.



> are you saying you blacklist if the previous cert was blacklisted? I'd like this logic removed.

Removed via 305b2c4364501b98fea970fdecc9f217b386579b.

tresf:
> Something along the lines of "Be sure to update QZ Tray to version 2.0.9 or later to use any newly generated certificates".

This would be a huge problem.  Backwards compatibility at renewal time is paramount and the confusion of adding two certificate download is not ideal.

This is going to look very confusing, but how about we do something like follows... 

```diff
  -----BEGIN CERTIFICATE-----
  MIIEyCBD...
 --START INTERMEDIATE CERT--
 MIIFEjCCA...
+  --START INTERMEDIATE CERT--
+  --START RENEWAL INFO--
+  AAABBBB...
+  --START RENEWAL SIGNATURE--
+  AAABBBB...
```

This will allow the `split[0]` and `split[1]` to continue working and then we can phase out the invalid `--START INTERMEDIATE CERT--` with the improved base code.

tresf:
Nevermind.... `if (split.length == 2) {` kills that workaround. :(

tresf:
@lukas-w let's switch back to the idea of storing this information inside the certificate data itself.  It will fix the backwards compatibility issues with certificate parsing.

I found a field that is very liberal about what it allows... `SAN`.  `SubjectAlternateName` is common on SSL certs for the various domains that are allowed.  I would much rather use a private extended attribute out of courtesy of the x509 specification but finding documentation for this outside of the RFC has been very difficult.

Here's an example of a `SAN` URI that can contain the fingerprint.
```
#3: ObjectId: 2.5.29.17 Criticality=false
SubjectAlternativeName [
  URIName: qz:renew/43:51:43:a1:b5:fc:8b:b7:0a:3a:a9:b1:0f:66:73:a8
]
```

I'm not proud of this, but forge should be able to write it and Java should be able to read it without the backwards compatibility issues.  Thoughts welcome.

lukas-w:
Ok, I'll try this and open a new pull request.

PR #396, opened 2018-12-19
Gracefully handle certificate renewals

### LINKED ISSUE #43: Graceful handling of cert renewals
Currently our "Allow" dialog is displayed if a particular signer/host has been trusted by whitelisting the certificate using a certificate fingerprint.

At renewal time, a new certificate is issued to the client, but the client is given the option to use their own CSR or alternatively the public key (saved from 1, 2 or 3 year prior) which allows the private key used for signing to be re-used. (in addition, they always have the option to generate a brand new keypair, but there's no way to make that graceful)
- This certificate fingerprint is specific to the **certificate only**
- There's no way to let QZ Tray know that the new cert replaces the old cert
- The "Allow" dialog will show for every client after a cert has been renewed

The ideal scenario is to allow the whitelist process to whitelist all activity from the signer, but we don't know anything about the signer when the first dialog appears (there's no signature, just a certificate), so our JavaScript API would need to change slightly as well as our QZ Tray websocket API (perhaps add the a task of QZ Protocol versioning) for this to be implemented properly.

The solution must offer a level of backwards compatibility with the 2.0 branch (albeit falling back on the less-graceful 2.0 behavior).

lukas-w:
This is like #391 but the renewal info is stored in the certificate itself as was decided in https://github.com/qzind/tray/pull/391#issuecomment-444286171. The format of `digital-certificate.txt` is not changed, so certificates will remain backwards compatible. I still included eb96fc4c98e200db60910a2e5a616ffb3f7126e2 which ensures that changes to the certificate file will be possible in the future.

Closes #43

### DIFF
--- a/src/qz/auth/Certificate.java
+++ b/src/qz/auth/Certificate.java
@@ -145,16 +145,18 @@ public class Certificate {
 
             //Strip beginning and end
             String[] split = in.split("--START INTERMEDIATE CERT--");
+            byte[] serverCertificate = Base64.decode(split[0].replaceAll(X509Constants.BEGIN_CERT, "").replaceAll(X509Constants.END_CERT, ""));
 
             X509Certificate theIntermediateCertificate;
             if (split.length == 2) {
-                theIntermediateCertificate = (X509Certificate)cf.generateCertificate(new StringBufferInputStream(split[1]));
+                byte[] intermediateCertificate = Base64.decode(split[1].replaceAll(X509Constants.BEGIN_CERT, "").replaceAll(X509Constants.END_CERT, ""));
+                theIntermediateCertificate = (X509Certificate)cf.generateCertificate(new ByteArrayInputStream(intermediateCertificate));
             } else {
                 theIntermediateCertificate = null; //Self-signed
             }
 
             //Generate cert
-            theCertificate = (X509Certificate)cf.generateCertificate(new StringBufferInputStream(split[0]));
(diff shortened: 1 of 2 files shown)


src/qz/auth/Certificate.java
```diff
@@ -147,18 +145,16 @@ public Certificate(String in) throws CertificateParsingException {
 
             //Strip beginning and end
             String[] split = in.split("--START INTERMEDIATE CERT--");
-            byte[] serverCertificate = Base64.decode(split[0].replaceAll(X509Constants.BEGIN_CERT, "").replaceAll(X509Constants.END_CERT, ""));
 
             X509Certificate theIntermediateCertificate;
             if (split.length == 2) {
-                byte[] intermediateCertificate = Base64.decode(split[1].replaceAll(X509Constants.BEGIN_CERT, "").replaceAll(X509Constants.END_CERT, ""));
-                theIntermediateCertificate = (X509Certificate)cf.generateCertificate(new ByteArrayInputStream(intermediateCertificate));
+                theIntermediateCertificate = (X509Certificate)cf.generateCertificate(new StringBufferInputStream(split[1]));
```
tresf: According to Oracle, this should be completely valid and return either a casted `X509Certifcate` or raise a `CertificateException`.  In reality, the library fails to parse the cert without `\r` and or `\n` per http://www.doublecloud.org/2014/03/reading-x-509-certificate-in-java-how-to-handle-format-issue/.  I will be reverting to the Base64 code to avoid any backwards compatibility issues.

PR #401, opened 2019-01-17
Add certificate download to HTTP status page

tresf:
@bberenz opening a PR so we can have a todo list...

### DIFF
--- a/src/qz/ws/HttpAboutServlet.java
+++ b/src/qz/ws/HttpAboutServlet.java
@@ -130,9 +130,9 @@ public class HttpAboutServlet extends DefaultServlet {
 
             Certificate cert = jks.getCertificate("root-ca");
 
-            return "-----BEGIN CERTIFICATE-----\n" +
-                    new String(Base64.encodeBase64(cert.getEncoded()), StandardCharsets.UTF_8) +
-                    "\n-----END CERTIFICATE-----";
+            return "-----BEGIN CERTIFICATE-----" + System.lineSeparator() +
+                    new String(Base64.encodeBase64(cert.getEncoded(), true), StandardCharsets.UTF_8) +
+                    "-----END CERTIFICATE-----";
         } else {
             return null;
         }
--- a/src/qz/ws/HttpAboutServlet.java
+++ b/src/qz/ws/HttpAboutServlet.java
@@ -1,6 +1,5 @@
 package qz.ws;
 
-import org.apache.commons.ssl.Base64;
 import org.codehaus.jettison.json.JSONArray;
 import org.codehaus.jettison.json.JSONException;
 import org.codehaus.jettison.json.JSONObject;
@@ -18,7 +17,6 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
-import java.security.cert.Certificate;
 import java.util.Iterator;
 import java.util.Properties;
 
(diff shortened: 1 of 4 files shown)


src/qz/ws/HttpAboutServlet.java
```diff
@@ -51,6 +78,66 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
...
+        Properties sslProps = DeployUtilities.loadTrayProperties();
+
+        if (sslProps != null) {
+            KeyStore jks = KeyStore.getInstance("jks");
+            jks.load(new FileInputStream(new File(sslProps.getProperty("wss.keystore"))), sslProps.getProperty("wss.storepass").toCharArray());
+
+            Certificate cert = jks.getCertificate("root-ca");
+
+            return "-----BEGIN CERTIFICATE-----" + System.lineSeparator() +
+                    new String(Base64.encodeBase64(cert.getEncoded(), true), StandardCharsets.UTF_8) +
```
tresf: `System.lineSeparator()` can potentially conflict with `org.apache.commons.ssl.Base64` "isChunked" flag because Apache appears to default to `\r\n` on all platforms where as `System.lineSeparator()` defaults to `\n` on Unix, `\r\n` on Windows.

src/qz/ws/HttpAboutServlet.java
```diff
@@ -51,6 +78,66 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
         }
     }
 
+    private void generateJsonResponse(HttpServletRequest request, HttpServletResponse response) {
+        JSONObject aboutData = AboutInfo.gatherAbout(request.getServerName());
+
+        try {
+            String certData = loadCertificate();
```
tresf: Should we place this data in the `ssl:` block?

src/qz/ws/HttpAboutServlet.java
```diff
@@ -38,6 +61,10 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
         }
         display.append("</table>");
 
+        if (!aboutData.isNull("ssl") && aboutData.optJSONArray("ssl").length() > 0) {
+            display.append("<p><a href='/cert'>Download Cert</a></p>");
```
tresf: Similar to the JSON question... should we add this to the `ssl` section?

src/qz/ws/HttpAboutServlet.java
```diff
@@ -51,6 +78,66 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
...
+    }
+
+    private String loadCertificate() throws GeneralSecurityException, IOException {
+        Properties sslProps = DeployUtilities.loadTrayProperties();
+
+        if (sslProps != null) {
+            KeyStore jks = KeyStore.getInstance("jks");
+            jks.load(new FileInputStream(new File(sslProps.getProperty("wss.keystore"))), sslProps.getProperty("wss.storepass").toCharArray());
+
+            Certificate cert = jks.getCertificate("root-ca");
```
tresf: Is there a way to avoid hard-coding `root-ca` into this?  This could break print-server deployments.  For example, is there a way to grab this information from the active `socket` connection instead?

bberenz:
Went over the data source in AboutInfo class to see how I could get the data into the ssl block and realized there was a simpler way to do all of this... Data is in ssl block, no aliases are hard coded now.
I can't seem to find sources on the base64 method, which is why I assumed it was using `lineSeparator` too, that's been correct as well.

src/qz/ws/HttpAboutServlet.java
```diff
@@ -76,6 +88,48 @@ private void generateJsonResponse(HttpServletRequest request, HttpServletRespons
...
+        Properties sslProps = DeployUtilities.loadTrayProperties();
+
+        if (sslProps != null) {
+            KeyStore jks = KeyStore.getInstance("jks");
+            jks.load(new FileInputStream(new File(sslProps.getProperty("wss.keystore"))), sslProps.getProperty("wss.storepass").toCharArray());
+
+            if (jks.isCertificateEntry(alias)) {
+                Certificate cert = jks.getCertificate(alias);
+
+                return "-----BEGIN CERTIFICATE-----" + System.lineSeparator() +
```
tresf: @bberenz did you mean to leave in this `System.lineSeparator()`?

bberenz: Nope, must of missed resetting it when I reverted a few changes.

tresf:
Looks good.  Thanks for patching the cert logic to work with JDK11.  Merging.

PR #418, opened 2019-02-20
Add dithering render support

tresf:
Moving to an open PR so we don't lose track of this one.

@bberenz the offending unit test was a PDF which require the snapshot version from PDFBOX to test (and then the appropriate flags set).  We can re-use this branch for that as well if you'd like.


tresf:
Moved to #429.

Issue #316, opened 2018-05-01
Allow serial baud settings to be optional


tresf:
Currently, serial baud settings will fallback to some sane defaults (e.g. 9600, 8, 1, NONE, NONE) if not provided, however some devices such as the virtual serial port installed from an Arduino will auto-negotiate working settings without any baud settings at all.

This is a nice option so that a developer does not need to make a guess at the baud settings in order for communication to function.

Issue #85, opened 2016-07-27
Support non-UTF8 Serial Encoding


tresf:
Serial communication currently assumes the returned data is in UTF-8 form per https://github.com/qzind/tray/blob/2.0/src/qz/communication/SerialIO.java#L119.

We should consider adding a configuration option to the serial API to allow specifying the encoding format.

PR #442, opened 2019-04-17
Serial Improvements

bberenz:
Improvements to serial communications:

- Covers #85 by tx/rx encoding on options and separate rx encoding override in response format options.
- Covers #316 by explicit "0"/"auto" values.
- Add several new parsing options to responses
  - Raw/unfiltered
  - Variable length + crc
  - Include header/bounds on return
- Better processing visibility via logging.


bberenz:
It's by no means pretty, but here's my scrapped together serial test for an arduino: [serialTest.ino](https://gist.github.com/bberenz/cabd9ded203b3bb684840b88f0bb0528)

tresf:
Tested against 2.1.0-RC7 for backwards compatibility with the Mettler Toledo scale and everything that we used prior to this PR works.

Note, most of the new settings aren't fully tested yet.  They also are not documented in the wiki yet.

js/qz-tray.js
```diff
@@ -1275,28 +1291,20 @@ var qz = (function() {
...
-             *  @param {string} data.data The data to send to the serial device.
-             * @param {Object} [properties] DEPRECATED: Properties of data being sent over the serial port.
-             *  @param {string} [properties.baudRate=9600]
-             *  @param {string} [properties.dataBits=8]
-             *  @param {string} [properties.stopBits=1]
-             *  @param {string} [properties.parity='NONE'] Valid values <code>[NONE| EVEN | ODD | MARK | SPACE]</code>
-             *  @param {string} [properties.flowControl='NONE'] Valid values <code>[NONE | XONXOFF | XONXOFF_OUT | XONXOFF_IN | RTSCTS | RTSCTS_OUT | RTSCTS_IN]</code>
+             * @param {string} port An open serial port to send data.
+             * @param {string|Array<string>|Object} data Data to be sent to the serial device.
+             *  @param {string} [data.type='PLAIN'] Valid values <code>[FILE | PLAIN | HEX?? | BASE64??]</code>
```
pedroAlba: I see that `HEX` and `BASE64` options were added on documentation, but not on  `SerialUtilities#SerialDataType`. Are there any plans on supporting both of those types?

PR #444, opened 2019-04-25
Added Fallback value for CUPS_DPI when output is empty or invalid

MrL0co:
I Added a fallback value for CUPS_DPI when output is empty or invalid. This prevents qz from polling raw printers over and over.

This caused us al lot of trouble because we had 190 printers from which about half are raw label printers.

### DIFF
--- a/src/qz/utils/ShellUtilities.java
+++ b/src/qz/utils/ShellUtilities.java
@@ -240,6 +240,10 @@ public class ShellUtilities {
                     }
                 }
             }
+            if (!densityMap.containsKey(entry.getKey())) {
+                densityMap.put(entry.getKey(), null);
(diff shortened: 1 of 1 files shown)


src/qz/utils/ShellUtilities.java
```diff
@@ -240,6 +240,10 @@ public static String executeRaw(String[] commandArray) {
                     }
```
bberenz: If its going to check for a missing value after the loop, the `NumberFormatException` block should be emptied to avoid the duplicate code.

tresf: @MrL0co once you address @bberenz's comment, we'd happily merge.

MrL0co: Yes, Sorry i didn't have time during work to address this due to deadlines. It is fixed now.

PR #451, opened 2019-05-22
Initial dark theme support

tresf:
Adds initial Dark Mode support for Windows, macOS and Linux.

image

Caveats:
* Due to an upstream bug (https://bugs.openjdk.java.net/browse/JDK-8214131), the title bars are still Light Mode themed.
* Windows 10 has scaling issues, bug report filed against a (more recent) fork here: https://github.com/bobbylight/Darcula/issues/5
* ~A (benign) illegal access exception will show in the logs for Java 9 and higher.~<br>**Edit:** Fixed by switching to bobbylight/Darkula fork.


tresf:
With exceptions of a few color choices, Ubuntu 18.04 seems to handle the dark-mode theme for Java just fine without using Darkula theme.  Will patch the PR to handle this scenario.

image

tresf:
@Vzor- I'm testing with `Adwaita-dark` on Ubuntu and for some reason the `Gdk` JNA code fails and never shows the "Approve" dialog.

If I comment-out `DPI.getDPI();` the issue does NOT occur.

```console
(java:13526): GLib-GObject-WARNING **: 13:57:36.177: cannot register existing type 'GdkDisplayManager'
(java:13526): GLib-CRITICAL **: 13:57:36.177: g_once_init_leave: assertion 'result != 0' failed
(java:13526): GLib-GObject-CRITICAL **: 13:57:36.177: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
```

src/qz/utils/DPI.java
```diff
@@ -0,0 +1,44 @@
+//TODO find a better home and name for me
```
bberenz: I see two options for this - either we throw the class contents inside the `UbuntuUtilities` instead since that's the only location it is used, or we keep it here but call it `DisplayUtilities` and move relevant methods over from the `SystemUtilities` class too.
I like option 2.

tresf: Thanks @bberenz.  I've done a bunch of refactoring.  Can you take a second look?

src/qz/utils/GtkUtilities.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            Class toolkitClass = Class.forName("sun.awt.UNIXToolkit");
+            Method versionMethod = toolkitClass.getDeclaredMethod("getGtkVersion");
+            Enum versionInfo = (Enum)versionMethod.invoke(toolkitClass);
+            Method numberMethod = versionInfo.getClass().getDeclaredMethod("getNumber");
+            int version = ((Integer)numberMethod.invoke(versionInfo)).intValue();
+            log.debug("Found Gtk{}", version);
+            return version;
+        } catch(Throwable t) {
+            log.warn("Could obtain GtkVersion information from UNIXToolkit: {}", t.getMessage());
```
bberenz: "Could not"

src/qz/utils/GtkUtilities.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return 0;
+    }
+
+    private static double getGtk2ScaleFactor(GTK2 gtk2) {
+        Pointer display = gtk2.gdk_display_get_default();
+        log.debug("Gtk 3.22+ detected, calling \"gdk_screen_get_resolution\"");
```
bberenz: Is "3.22" the right version for this method?

src/qz/ui/LinkLabel.java
```diff
@@ -25,14 +28,16 @@
 
     private ArrayList<ActionListener> actionListeners;
 
+    public static Color DEFAULT_COLOR = Color.BLUE;
```
bberenz: Should this point to `Constants.TRUSTED_COLOR` instead?

src/qz/utils/GtkUtilities.java
Vzor-: should be 2.10

tresf:
All comments addressed.

PR #467, opened 2019-07-15
Initial support for mask tray icon

### LINKED ISSUE #464: Black and white tray icons
[mask-icon.zip](https://github.com/qzind/tray/files/3452795/mask-icon.zip)
@budislav is helping us create black/white icons for better macOS and Windows 10 integration.

First draft:

image

image

@klabarge made the following observations:
* QZ's black is darker than the other macOS icons
* QZ's icon appears closer to the edges than the other macOS icons

@tresf made the following observations
* In dark mode, the icon's paper looks smaller.

tresf:
Closes #464


src/qz/common/TrayManager.java
```diff
@@ -144,6 +143,28 @@ public TrayManager(boolean isHeadless) {
...
+                try {
+                    Thread.sleep(1000);
+                } catch(Throwable ignore) {}
+                boolean darkMode = SystemUtilities.isDarkMode();
+                while(true) {
+
+                    if (darkMode != SystemUtilities.isDarkMode(true)) {
+                        darkMode = SystemUtilities.isDarkMode();
+                        iconCache.fixTrayIcons(darkMode);
+                        // FIXME: how to know which icon to set?
```
tresf: @bberenz ideas welcome here.

src/qz/ui/IconCache.java
```diff
@@ -231,15 +235,73 @@ public void setBgColor(Icon i, Color bgColor) {
...
+                    dest = ColorUtilities.invert(dest);
+                }
+                System.out.println("replacing " + id + " with " + id.replaceAll("mask", "default"));
+                images.put(id.replaceAll("mask", "default"), dest);
+                imageIcons.put(id.replaceAll("make", "default"), new ImageIcon(dest));
+            }
+        }
+
+        // Handle undocumented macOS tray icon padding
+        // FIXME: This should only run once on red/yellow tray icons
```
tresf: @bberenz ideas welcome here too. :)

src/qz/ui/IconCache.java
```diff
@@ -231,15 +235,73 @@ public void setBgColor(Icon i, Color bgColor) {
...
+        return dest;
+    }
+
+    public void padIcon(Icon i, int percent) {
         for (String id : i.getIds()) {
-            images.put(id, ColorUtilities.invert(images.get(id)));
-            imageIcons.put(id, new ImageIcon(images.get(id)));
+            // Calculate padding percentage
+            int w = images.get(id).getWidth();
+            int w1 = w;
```
tresf: left over from debug, needs to be removed

src/qz/ui/IconCache.java
tresf: Added a `Icon.padded` attribute so we know whether or not to re-pad.

src/qz/ui/IconCache.java
tresf: Removed

src/qz/ui/ThemeUtilities.java
```diff
@@ -0,0 +1,46 @@
+package qz.ui;
+
+import javax.swing.*;
+import java.awt.*;
+
+public class ThemeUtilities {
```
bberenz: I propose the following changes to the `refreshAll` methods in this class

```java
public static void refreshAll(Container container, Component ... stragglers) {
    // Handle orphaned UI objects (e.g. Component added to a message dialog)
    for(Component straggler : stragglers) {
        /// as is
    }

    refreshAll(ArrayUtils.addAll(c.getComponents(), stragglers));
}

public static void refreshAll(Component ... components) {
    for(Component c : components) {
        /// as is
    }
}
```

Unless there's a reason to refresh the container explicitly before the stragglers, in which case then I think we should get rid of the first overload (with passing the null) and the associated null checks and just let the for-loops iterate over the resulting 0 element arrays.

src/qz/utils/SystemUtilities.java
```diff
@@ -264,10 +263,13 @@ public static boolean isDarkMode(boolean recheck) {
...
-        if (isDarkMode()) {
-            Constants.WARNING_COLOR = Constants.WARNING_COLOR_LIGHTER;
-            Constants.TRUSTED_COLOR = Constants.TRUSTED_COLOR_LIGHTER;
-        }
+        Constants.WARNING_COLOR = isDarkMode() ? Constants.WARNING_COLOR_DARK : Constants.WARNING_COLOR_LITE;
+        Constants.TRUSTED_COLOR = isDarkMode() ? Constants.TRUSTED_COLOR_DARK : Constants.TRUSTED_COLOR_LITE;
+    }
+
+    public static boolean prefersMaskTrayIcon() {
+        // TODO: Test on Windows
```
bberenz: Is this todo something meant to be targeted in this PR?

src/qz/ui/IconCache.java
```diff
@@ -231,15 +236,73 @@ public void setBgColor(Icon i, Color bgColor) {
...
+        if (SystemUtilities.isUbuntu()) {
+            UbuntuUtilities.fixTrayIcons(this);
+        }
+
+        // Handle mask-style tray icons
+        if (SystemUtilities.prefersMaskTrayIcon()) {
+            // Clone the mask icon
+            for (String id : Icon.MASK_ICON.getIds()) {
+                BufferedImage src = images.get(id);
+                BufferedImage dest = clone(src);
```
bberenz: Why not use `images.get(id)` directly if we aren't going to use the `src` reference anywhere else?

src/qz/ui/IconCache.java
```diff
@@ -231,15 +236,73 @@ public void setBgColor(Icon i, Color bgColor) {
...
+    public static BufferedImage clone(BufferedImage src) {
+        Image tmp = src.getScaledInstance(src.getWidth(), src.getHeight(), src.getType());
+        BufferedImage dest = new BufferedImage(tmp.getWidth(null), tmp.getHeight(null), BufferedImage.TYPE_INT_ARGB);
+        Graphics g = dest.createGraphics();
+        g.drawImage(tmp, 0, 0, null);
+        g.dispose();
+        return dest;
+    }
+
+    public void padIcon(Icon i, int percent) {
```
bberenz: For clarity can we call this param `icon` instead? Especially since its being used in a for-loop.

src/qz/ui/IconCache.java
```diff
@@ -231,15 +236,73 @@ public void setBgColor(Icon i, Color bgColor) {
...
         for (String id : i.getIds()) {
-            images.put(id, ColorUtilities.invert(images.get(id)));
-            imageIcons.put(id, new ImageIcon(images.get(id)));
+            // Calculate padding percentage
+            int w = images.get(id).getWidth();
+            int wPad = (int)((percent/100.0) * w);
+            w += wPad;
```
bberenz: Another clarity item, instead of adding to `w` beforehand like this, can we use `w + wPad` in the buffered image arguments? (and for `h` too)

src/qz/utils/SystemUtilities.java
tresf: @bberenz may be left in the codebase.  Windows too has monochrome tray icons, but the theming of the start bar is different (some light mode installs have a dark bar).  I'd like to support this but may need some platform-specific code to know which color that icon should be.  If you can find information on this, I'd be happy to add it to this PR.

src/qz/ui/GatewayDialog.java
```diff
@@ -112,6 +112,10 @@ private void initComponents() {
         setLocationRelativeTo(null);    // center on main display
     }
 
+    public void refresh() {
+        ThemeUtilities.refreshAll(this, new Component[] { certTable });
```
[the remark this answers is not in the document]
tresf: Good catch.  I use varargs so seldom that I was completely unaware.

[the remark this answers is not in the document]
tresf: ... the `refreshAll(Component ...)`picks up calls to `refreshAll(Container)` too so this varargs really makes the codebase cleaner. 👍

src/qz/ui/ThemeUtilities.java
tresf: Done via d39792d.

src/qz/ui/GatewayDialog.java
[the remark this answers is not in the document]
tresf: Done via d39792d.

src/qz/ui/IconCache.java
tresf: Done via bd4937e.

src/qz/ui/IconCache.java
tresf: Done via bd4937e.

src/qz/ui/IconCache.java
tresf: Done via bd4937e.

tresf:
@bberenz all comments addressed.  The only outstanding item is your question about the Windows `TODO:`.  I'm fine with leaving that in the codebase, removing it, or researching it.  I have a feeling it won't be easy to support (due to Window's strange theme decisions), which is why I didn't dive into it just yet.

src/qz/utils/SystemUtilities.java
tresf: On Windows 10 Pro, I can't seem to trigger dark tray icons at all.

Here's me trying to manipulate the taskbar to be white.  Notice when the color is darker than a particular threshold Windows says `This color is not supported`.
image

Rightfully so, here's what white looks like:
image

What puzzles me is I remember my HP Spectre laptop having a light themed start bar on install.  It was Windows 10 Home (not professional) but it's since been reloaded with professional. I may have to setup a VM to fully test this.

tresf: Light Mode for Windows 10 appears to be brand-new with Cumulative Update 1903, which is opt-in currently.  Installing now in a VM to see what it's like.

On a side-note -- looking at Server 2012 -- it appears the Windows icons are white with a dark border/drop-shadow.

We may be able to add this same drop-shadow programmatically.  Pinging @Vzor- for insight. :)

> image

bberenz: It's fine if its not, just wanted to make sure it wasn't overlooked.
I wasn't aware if windows had any icon masking at all. Interesting to know it has better light color support coming, that one has always annoyed me. 
Based on this I think it might be best to come back to it later then when window's has wider support for it released.

tresf:
Update 1903 definitely fixes the Light Mode theme on Windows 10, which we'll need to accommodate if we decide to toggle this on for Windows 10.

Screen Shot 2019-07-24 at 10 09 19 PM

src/qz/utils/SystemUtilities.java
tresf: Added via 860e559, but only for Windows 10 and higher.  Windows 7 seemed to use a stroke around the white icons to avoid some issues with lighter themes.  Preview added below.

tresf:
Mask icon support for Windows 10+ is coded.  If someone has a Windows 10 install older than the `1903` build, I'd like to test that the icon is properly represented since the task bar was inconsistently "dark" up until that release despite whether or not "dark mode" was toggled on.

mask_icon

klabarge:
I tested on Windows 10 1709 (granted, it was an MS dev machine). Unfortunately, I am unable to use the "Personalize" settings because my trial license has expired.

I also tested on Windows 7, and the tray icon is the normal, color icon.

image

The tray icon looks nice...

image

tresf:
@klabarge what zoom percentage is that desktop running?  The icon looks pixelated.

image

Currently, I try to match the icon to 200%, (32x32) but I don't have one that's setup for 300% (48x48).

tresf:
48x48 icons added, tested successfully on the HP Spectre, which defaults to 300%.

image

... also -- although we don't currently use it -- I've added a JDK8-compatible way for detecting HiDPI display by using JNA to call Win32 libraries.

PR #303, opened 2018-04-07
I18nsupport

johann-sonntagbauer:
This pull request combines #290 and the following changes

* gettext extraction Ant Task
* change language UI
* german translations

PR #469, opened 2019-07-19
I18nsupport continuation + JSON

Vzor-:
~~This is a WIP do not merge.~~
This PR is ready for testing.

This is a continuation of pr #303. 

Done
- Translator script reads & writes to JSON files
- gettext reads JSON files

Todo
- Add gettext to missing items
- ~~Utf8 broke at some point but was working~~
    - ~~Dark mode on windows fails to display Chinese symbols. This seems to be an issue with the bobbylight dracula theme~~
- ~~bugfix, some ui items are missing at the moment~~


Vzor-:
This is mostly done, we need to find if any strings are not surrounded with the gettext function that should be. Right now if something has a gettext but no translation, it will automatically add the English version of the string as the translation during the ant build (eg the translation of "Hello" in Chinese would start out as "Hello" until someone when in and replaced it with "你好"). This makes is hard to differentiate an untranslated string from a string with a default translation. 

To fix this for testing, open the language pack (Chinese CN is the most obvious for testing) and replace every English pass through translation with something obvious like ampersands.

tresf:
Sorry, this was auto-closed when `2.1` was moved to `master`.  It had rebase issues, so if you don't mind @Vzor- we'll need it reopened and retargeted.

PR #478, opened 2019-08-08
Update warning dialogs to show signature validity

### LINKED ISSUE #459: Differentiate between bad signature and bad certificate
**Problem**

When the "Gateway Dialog" is shown for an invalid signature, it shows a dialog which suggests that the certificate is invalid.  This is confusing to customers as the issue is often a bad signature.

The proposal is to make the dialog accurate as to the problem as well as add better information to this dialog for troubleshooting purposes.  Mockup attached:

image

In addition, this enhancement will **finally** allow "always allow" for invalid signatures, limited to that particular session/page refresh/websocket connection.

tresf:
Closes #459


src/qz/ui/component/CertificateTable.java
```diff
@@ -0,0 +1,155 @@
...
+    /** Custom cell renderer for JTable to allow colors and styles not directly available in a JTable */
+    private class CertificateTableCellRenderer extends StyledTableCellRenderer {
+
+        @Override
+        public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int col) {
+            JLabel label = (JLabel)super.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, col);
+
+            // First Column
+            if (value instanceof CertificateField) {
+                stylizeLabel(STATUS_NORMAL, label, isSelected);
```
tresf: You don't explicitly return the `label` here.  Was this intended?  It's not consistent.

src/qz/ui/GatewayDialog.java
```diff
@@ -135,14 +138,31 @@ public void actionPerformed(ActionEvent e) {
...
                                              "</html>");
-            certInfoLabel.setText("Certificate information");
-            verifiedLabel.setIcon(iconCache.getIcon(cert.isTrusted()? IconCache.Icon.VERIFIED_ICON:IconCache.Icon.UNVERIFIED_ICON));
+            certInfoLabel.setText("View request details");
+
+            IconCache.Icon trustIcon;
+            Color detailColor = Constants.TRUSTED_COLOR;
+            if (request.isTrusted()) {
+                //cert and signature are good
+                trustIcon = IconCache.Icon.TRUST_VERIFIED_ICON;
```
tresf: These new changes do not persist after dark/light theme changes.

src/qz/ui/component/CertificateTable.java
bberenz: `stylizeLabel` just returns the label object passed in the parameters, but we are setting an icon in this block first and then returning it.

tresf: Sorry, I'm talking about the lack of redundant assignment (`return` was in context to the function being called )

PR #488, opened 2019-09-10
jdk-11 bump + travis fix

Vzor-:
I don't think adding misc.xml to the git ignore actually does anything here, should it be omitted?

Although a bit of a read, this issue https://github.com/OfficeDev/ews-java-api/issues/235 is basically identical to ours. The tldr of it is the only way to get misc.xml to behave how we want would be to have a task create the misc.xml from a template. That however is not really viable unless it was built directly into intellij. This is a know issue and jetbrains has an active issue on it https://youtrack.jetbrains.com/issue/IDEA-90785 (not our exact problem, but the closer issues were marked as dupe of this)


tresf:
This doesn't really address the following files that are always trying to commit:
* `tray.iml`
* `codeStyleConfig.xml`
* `Project.xml`
* `encodings.xml`

From what I'm reading, git's going to track it regardless of whether or not it's in the `.gitignore` because it's still in the repo.

Furthermore, Travic-CI is [still failing](https://travis-ci.org/qzind/tray/builds/583026215?utm_source=github_status&utm_medium=notification).

tresf:
Perhaps this: https://stackoverflow.com/a/13631525/3196753

tresf:
> Furthermore, Travic-CI is still failing.

Nevermind this comment... I restarted the jobs and they're all green now.  The first two were due to gluon's endpoint being too slow.   The third was just a hung build.

However, I'm thinking we should just remove the oraclejdk8 builds.  Oracle is making it harder and harder for projects to use the older versions.

Vzor-:
> Perhaps this: https://stackoverflow.com/a/13631525/3196753

skip-worktree is a better solution than assume-unchanged, but they are both local only changes. I am looking into other solutions upstream.

PR #493, opened 2019-09-28
Certificate Adjustments

bberenz:
Fix some problems that came up with the recent improvements to requests/certs:

- OffsetDateTime MIN/MAX values are actually outside the valid Instant range
- Allowed/Blocked sites couldn't be removed via the manager dialog due to date formatting mismatch
- Checking cert trusted without checking request trusted (renamed to "verified" to help in the future)


tresf:
Tested with JDK8 and JDK11.  Fixes both issues.

PR #525, opened 2019-11-27
Support alternate signing algorithms

### LINKED ISSUE #191: Support signature hashes other than SHA1
Given that [SHA1 collisions](https://shattered.it/) have been discovered it would be nice to be able to allow the server implementing the signing function to use a stronger hash function.

bberenz:
Support for SHA256 and SHA512 on signatures.

Drops the simplersa library in favor of calling the signature methods ourselves, as it appears to hard-code to sha1 only.

**Edit:** Closes #191


js/qz-tray.js
```diff
@@ -1988,10 +2011,44 @@ var qz = (function() {
...
+            *
+            * @param {string} algorithm The algorithm used in signing. Valid values: <code>[SHA1 | SHA256 | SHA512]</code>
+            * @since 2.1.0
+            *
+            * @memberof qz.security
+            */
+            setSignatureAlgorithm: function(algorithm) {
+                //warn for incompatibilities if known
+                if (!_qz.security.isAlgorithmCompatible()) {
+                    return;
```
bberenz: @tresf, I'm not sure if preventing setting the algorithm as well is the right choice or if we should just make the call to print a warning and let it be set anyway.
And if we do want to prevent it, then in cases where the algorithm was set before a connection, should we reset it back to sha1 or just let it be?

tresf: It's informational, so I think a soft warning is fine for now.

tresf: On second thought, maybe we shouldn't be checking if it's compatible here?  We usually setup the API before connection, so this software warning will almost never fire.  Can we refactor this to a location that will fire 100% of the time?

tresf:
Bumped all signing examples to use `sha512` via 49ace3d.

Note, `sha512` was chosen over `sha256` because it has documented performance benefits on 64-bit systems and has a stronger collision resistance over `sha256`.  Quoting a [stackoverflow answer](https://security.stackexchange.com/a/165588/152647):

> Different hardware favors different functions. `SHA-512` is generally faster on 64-bit processors, `SHA-256` faster on 32-bit processors. (Try the command `openssl speed sha256 sha512` on your computer.)

js/qz-tray.js
bberenz: We call `isAlgorithmCompatible` here and [after opening](https://github.com/qzind/tray/pull/525/files#diff-5b383df0ac6d733ed512ffe910a83668R343) a connection

tresf: Perfect.  Merging.

PR #535, opened 2019-12-17
Add jscript signing support

tresf:
Comodo AV blocks our scripts, this is an attempt to white-list them by adding authenticode signing as a signature block at the bottom of the JScript files.   This PR also adds authenticode to the `.exe` launcher.

This patch is only valid for 2.0 versions.  These JScript files have been removed beginning with 2.1.

Thanks to the kind work of @ebourg at https://github.com/ebourg/jsign/issues/69.


tresf:
Checking the release notes for jsign 3.0:
> Jsign now requires Java 8 or higher

Unfortunately, the 2.0 branch can't use Java 8, so we can't use this feature.  That's of little consequence though for the following reasons:
* Comodo AV still blocks our scripts even when they're signed
* The 2.1 branch DOES use Java 8, so it can benefit from the newer jsign version

Slightly unrelated, but this PR does add support for a signed `qz-tray.exe` launcher, which both 2.0 and 2.1 branches could benefit from.  Opening new issue to track that as I close this.

Since [ant doesn't handle blank values well](https://github.com/ebourg/jsign/issues/71) we have target-specific logic to handle authenticode.  I feel this code will become redundant when signing three separate files so I'm asking @bberenz to help write a reusable ant target for this. :D

tresf:
Edit: Sorry, had accidentally posted in the wrong bug report. Corrected.  Closing.

PR #559, opened 2020-01-23
Add whitelisted certificate directory

### LINKED ISSUE #557: Add post-deployment task for whitelisting
Predominantly intended for white-label customers, a certificate should be white-listable through a post-deployment task.

tresf:
Adds the ability to scan and whitelist certificates that are provided along with a Desktop install.

**Ant**:
The file will be automatically placed in the correct location using the `whitelist.use` parameter.

```bash
ant nsis -Dwhitelist.use=path/to/digital-certficate.txt
```

**Manually**:
The file should be located in `<install>/whitelist/`, i.e:

* Mac : `/Applications/QZ Tray.app/whitelist/`
* Linux : `/opt/qz-tray/whitelist/`
* Windows: `C:\Program Files\QZ Tray\whitelist\`

The white-label portal will be enhanced to pre-whitelist a certificate so that users of the white-label version won't need to click  "☑️ Remember this decision".

Closes #557

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -4,7 +4,7 @@
     <property file="ant/project.properties"/>
     <import file="ant/javafx.xml"/>
 
-    <target name="distribute" depends="init,clean,sign-jar,download-javafx,override-authcert,include-assets">
+    <target name="distribute" depends="init,clean,sign-jar,download-javafx,override-authcert,include-assets,whitelist-certs">
         <echo message="Process complete" />
     </target>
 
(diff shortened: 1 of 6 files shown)


klabarge:
All tests pass.

Unsurprisingly (but I thought I'd note it), the certificate gets clobbered if the installer runs and the name of the certificate file is the same. This won't cause issues at renewal, because both fingerprints in the `allowed.dat` file are present. If the cert file name is different, then it won't clobber anything in the whitelist folder.

|                   |Windows |macOS |Ubuntu  |
|------------|---------|--------|---------|
|**Manual** |  ✅          |  ✅       |  ✅         |
|**Ant**       |  ✅          |  ✅       |  ✅          |

PR #562, opened 2020-01-29
FileIO Filtering

### LINKED ISSUE #555: FileIO Add File Filtering Support
Add support for filtering which FileIO files in a folder are listened upon.

Quoting the customer:

> Software [needs] to be able to only read files like:
> 
> `1.txt`
> `2.txt`
> 
> But QZ also sees files like the below (which we don't need access to) and by accessing them, seems to mess up the other software using the same folder.
> 
> `1.it`
> `2.it`

Vzor-:
Closes #555 

Some things to note,

- If `include` is empty or missing, it will include everything.
- `include`/`exclude` are both lists of strings. Wildcard behaviour is documented [here](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#wildcardMatch(java.lang.String,%20java.lang.String,%20org.apache.commons.io.IOCase)).
- The exclusion list overrides the inclusion list.

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1985,7 +1985,7 @@ var qz = (function() {
              */
             startListening: function(path, params) {
                 if (params && typeof params.include !== 'undefined' && !Array.isArray(params.include)) {
-                     params.include = [params.include];
+                    params.include = [params.include];
                 }
                 if (params && typeof params.exclude !== 'undefined' && !Array.isArray(params.exclude)) {
                     params.exclude = [params.exclude];
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1975,7 +1975,7 @@ var qz = (function() {
              *   @param {boolean} [params.listener.reverse] Controls whether data should be returned from the bottom of the file.  Default value is true for line mode and false for byte mode.
              *   @param {string|Array<string>} [params.include] File patterns to match.  Blank values will be ignored.
              *   @param {string|Array<string>} [params.exclude] File patterns to exclude.  Blank values will be ignored.  Takes priority over <code>params.include</code>.
-             *   @param {boolean) [params.ignoreCase=true] Whether <code>params.include</code> or <code>params.exclude</code> are case-sensitive.
+             *   @param {boolean} [params.ignoreCase=true] Whether <code>params.include</code> or <code>params.exclude</code> are case-sensitive.
              * @returns {Promise<null|Error>}
              * @since 2.1.0
              *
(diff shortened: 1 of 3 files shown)


Vzor-:
That html diff went a bit crazy. Is there any way to clean that up?

tresf:
> That html diff went a bit crazy. Is there any way to clean that up?

For now, we'll go back to the `panel-heading` design layout and reduce the number of input fields.

That said, git does have a tendency to make HTML diffs look bad when you're just adding a div.   In this case, I'm more concerned with the real estate that the new feature is taking up and I think two fields and the old layout is a good compromise for now.

tresf:
> `include/exclude` are both lists of strings. Wildcard behaviour is documented here.

Mentioned in chat, but make sure the API is documented and for consistency purposes, allow both String or String Arrays like this: https://github.com/qzind/tray/blob/20e2b556b3b18a819f0083649cfb72ca96aed57e/js/qz-tray.js#L870-L872

tresf:
> Wildcard behaviour is documented here.

Apache to the rescue. :D . This is a great idea as regex would have been harder to document and support.  Nice!

klabarge:
This works great on macOS.

* `exclude: ["*.txt"]`: All files that **do not have** the file extension ".txt" are read.
* `include: ["*.txt"]`: Only files that **have** the file extension ".txt" are read.
 
However, in Windows, I can't seem to get either the `exclude` or `include` to be honored. I also tried not using a wildcard. For example, I added a file called "test.txt" with the config set to`exclude: "test.txt" and the file was still read.


The logs don't show anything after the `file.startListening` call
```
Message: {"call":"file.startListening","promise":{},"params":{"path":"","sandbox":true,"shared":false,"include":[],"exclude":["test.txt"],"ignoreCase":true,"listener":{"reverse":true,"lines":"10"}},"signature":"JManTJ
```

tresf:
>  in Windows, I can't seem to get either the `exclude` or `include` to be honored

We took a look and it was a mistake with the testing environment, was running an older version).

Some updates that were requested but missing:

* Added missing jsdocs
* Added `string|Array<string>` support
* Improved tooltips

@klabarge can you please re-test?

tresf:
@bberenz the code is ready for review as well.

tresf:
Documentation published: https://qz.io/wiki/2.1-File-Communication#filter-results

klabarge:
I re-tested changes on macOS and Windows and this looks good.

tresf:
This code is bad.  It only filters events to the browser, the files are still being opened, causing locks, etc.

PR #563, opened 2020-01-30
Add support for pdf bounding box

### LINKED ISSUE #198: PDF CropBox Support
It would be great to have the ability to pass a cropbox when printing PDFs. [Apache PDFBox supports it](https://pdfbox.apache.org/docs/2.0.2/javadocs/org/apache/pdfbox/pdmodel/PDPage.html#setCropBox(org.apache.pdfbox.pdmodel.common.PDRectangle)). Here's a [StackOverflow post](https://stackoverflow.com/a/20724647) discussing implementation. It looks like it would fit into the QZ source code [somewhere around here](https://github.com/qzind/tray/blob/61a2f1b9757c6aa8bc2725c645f86c0a43f163bf/src/qz/printer/action/PrintPDF.java#L116).

We've got a use case where we'd like to print USPS SCAN forms on a 4"x6" thermal printer to avoid adding a full printer for that sole reason. The form is an 8.5"x11" PDF, which we can send to QZ reasonably, but the barcode is too small for USPS to scan. In QZ's defense it's perfectly legible so that's a technical shortcoming on their end.

A solution to the problem would be to crop out just the barcode when printing since it is a fixed location, and this would allow it to fit unscaled on the 4"x6" label since the true size of the barcode is less than that size.

Cropping would need to play nice with rotation - in this specific case the goal is to crop and print landscape rather than the PDF's natural portrait orientation.

[Here is an example USPS SCAN form](https://www.usps.com/clicknship/_pdf/SCAN_Form_5630_04282008.pdf).

tresf:
Adds cropbox support.
Closes #198

### DIFF
--- a/sample.html
+++ b/sample.html
@@ -2276,8 +2276,8 @@
         $("#pxlSizeActive").prop('checked', false);
         $("#pxlSizeGroup").css('display', 'none');
 
-        $("#pxlBoundX").val('');
-        $("#pxlBoundY").val('');
+        $("#pxlBoundX").val(0);
+        $("#pxlBoundY").val(0);
         $("#pxlBoundWidth").val('');
         $("#pxlBoundHeight").val('');
         $("#pxlBoundsActive").prop('checked', false);
--- a/sample.html
+++ b/sample.html
@@ -486,6 +486,40 @@
                                                     <input type="number" step="any" id="pxlSizeHeight" class="form-control pull-right" />
                                                 </div>
                                             </div>
+
+                                            <div class="form-group">
+                                                <label class="inline">Bounds</label>
+                                                (
+                                                <label for="pxlBoundsActive" class="inline">Enable:</label>
+                                                <input type="checkbox" id="pxlBoundsActive" onclick="checkGroupActive('pxlBoundsActive', 'pxlBoundsGroup');" />
+                                                )
+                                            </div>
+                                            <div class="inline" id="pxlBoundsGroup">
+                                                <div class="form-group form-inline">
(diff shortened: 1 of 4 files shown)


tresf:
@klabarge can you please test this feature?

tresf:
Documentation added https://github.com/qzind/tray/wiki/2.1-Pixel-Printing#pdf-cropbox

sample.html
```diff
@@ -2242,6 +2276,13 @@ <h4 class="panel-title">Options</h4>
         $("#pxlSizeActive").prop('checked', false);
         $("#pxlSizeGroup").css('display', 'none');
 
+        $("#pxlBoundX").val('');
+        $("#pxlBoundY").val('');
```
klabarge: I think these values should be set to 0.

If these values are null when the `config` is sent, you'll get this error.

```
[ERROR] 2020-01-30 19:04:51,758 @ qz.utils.PrintingUtilities:?
	Failed to print
java.lang.NumberFormatException: Character N is neither a decimal digit number, decimal point, nor "e" notation 
```

klabarge:
Other than my comment, this feature looks good. I tested on macOS, Windows, and Ubuntu and the cropping worked.

sample.html
tresf: > I think these values should be set to 0.

We do this for `margins: {}`, but not for `size: {}`.  We can default `x: 0` and `y: 0`, but for consistency in the sample, I don't think we should default the width and height.

tresf: Ok, turns out `optDouble` strangely divides by zero when the value isn't parsable.  Patched via 
1e9af8b, a95bf4b.

Issue #566, opened 2020-02-04
API vagueness


tresf:
A client brought to our attention today that the API definition is a bit vague in regards to licensing, specifically `setSignaturePromise` and `setCertificatePromise`.

The issue arose when the client tried to pass an in-line `async` arrow function as a parameter to `setSignaturePromise`.

However, I believe `setSignaturePromise` strictly requires a synchronous function.  This is confusing because the API doesn't make a clear distinction between the parameter types.

Furthemore, it may make sense to support both.

Quoting the current API:

### setCertificatePromise(promiseCall)
Set promise resolver for calls to acquire the site's certificate.

Name | Type | Description
-- | -- | --
`promiseCall` | `function` | `Function({function} resolve)` called as promise for getting the public certificate. Should call resolve parameter with the result.

### setSignaturePromise(promiseGen)
Set promise creator for calls to sign API calls.

Name | Type | Description
-- | -- | --
`promiseGen` | `function` | `Function({function} toSign)` Should return a function, `Function({function} resolve)`, that will sign the content and resolve the created promise.

bberenz:
They are both expecting the callback function used by a promise to handle the resolve/reject.
```javascript
/*so not this obj->*/new Promise(/*but this bit ->*/function(resolve, reject) {}).then(/*and not this bit ->*/function(result) {})
```
Not sure what would help clear that up though, a few ideas include `set___Handler`, `set___Callback`, `set___Resolver` ?

And looking at how we wrap things, it shouldn't be too hard to accept a callback or a promise directly for either of those and just check which is set when we need to make use of it.

- - -

As for the async arrow function, as far as I can tell that's already supported, as long as the `resolve` parameter gets called, it shouldn't matter:
```javascript
qz.security.setCertificatePromise(async (resolve, reject) => { /*code*/ resolve(..); });
```
But from the sound of this issue, it seems like maybe they thought they were passing a function to be used as the `.then` function, rather than the `resolve/reject` function, maybe? I don't have enough details on that to be sure though.

tresf:
Perhaps my description is off, but the code looked something like this... 

```js
qz.security.setSignaturePromise(async (toSign) => {
    return async (resolve, reject) => {
    // ...
    };
});
```

From my understanding, an asynchronous function returns an implicit promise.

tresf:
@bberenz here's some proof of concept code that allows `await` syntax... 

https://gist.github.com/tresf/c8fe9f70bdb32b62d333bdc75ae34936

[...]

PR #575, opened 2020-02-08
Clarify API naming/docs

bberenz:
For #566
Changes to some of the security API to improve understanding of what the calls are expecting.
Also adds support for directly passing promises/async functions.

Open to feedback on the new names and the altered jsdocs.

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -511,17 +511,29 @@ var qz = (function() {
 
         security: {
             /** Function used to resolve promise when acquiring site's public certificate. */
-            certPromise: function(resolve, reject) { reject(); },
-            /** Called to create new promise (using {@link _qz.security.certPromise}) for certificate retrieval. */
+            certHandler: function(resolve, reject) { reject(); },
+            /** Called to create new promise (using {@link _qz.security.certHandler}) for certificate retrieval. */
             callCert: function() {
-                return _qz.tools.promise(_qz.security.certPromise);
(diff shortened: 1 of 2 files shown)


tresf:
> Open to feedback on the new names and the altered jsdocs.

I like the code snippets and I understand why the function names are proposed to be changed, but... 

1. The old function names much more accurate with the new API.
2. MDN describes a promise as the following:
> The `Promise` object represents the eventual completion (or failure) of an asynchronous operation, and its resulting value.

So I think `setSignaturePromise` and `setCerticatePromise` are still very descriptive

On the other hand, a factory function can return anything, not limited to the concept of "eventual completion", so although it's slightly more accurate for the signature generation, taking #566 into ECMAScript 2017 and combing with `async` and arrow functions, I believe leaves a more desirable API by keeping the word "Promise".

klabarge:
Whoops, I accidentally reported my testing to the original bug report.

Here's my testing:

> I was able to successfully sign using the below functions. I also tested the "old" functions for backwards compat.
> 
> @tresf Is there anything else I should test?
> 
> ```js
> qz.security.setCertificatePromise(async () => {
>     return $.ajax({ url: "assets/signing/digital-certificate.txt", cache: false, dataType: "text" })
> });
> ```
> 
> ```js
> qz.security.setSignaturePromise(async (toSign) => {
>     return $.ajax("assets/signing/sign-message.php?request=" + toSign)
> });
> ```

Issue #32, opened 2016-04-12
Inconsistent HTML printing


klabarge:
I sent 10 identical print jobs to a PDF printer on Ubuntu 14.04.  I used complete defaults from https://demo.qz.io/ except I changed the job name for each print.

I don't see any difference in the console between the three different outcomes, but I have provided the stacktraces anyway.  I have not provided logs for `rasterize : false` but can be provided if needed.

Here are the results:
- (2) sent as 17.9kb
- (3) sent as 36.3kb
- (5) sent as 136.7kb

I then sent an additional 10 jobs with `rasterize : false`
- (8) 19.8kb
- (1) 9.0kb
- (1) 20.5kb
### With Rastorize
1. **17.9kb**
   - Result: 
     ![image](https://cloud.githubusercontent.com/assets/12505463/14450700/e5305ec8-004e-11e6-8775-548816e38d6f.png)
   - Stacktrace:
   
   ```
   [DEBUG] 2016-04-12 01:14:26,074 @ qz.ws.PrintSocketClient:?
   Message: {"call":"print","promise":{},"params":{"printer":{"name":"PDF"},"options":{"colorType":"color","copies":"1","density":"","duplex":false,"interpolation":"","jobName":"9","margins":"0","orientation":"","paperThickness":"","printerTray":"","rasterize":true,"rotation":"0","scaleContent":true,"size":null,"units":"in","altPrinting":false,"encoding":"","endOfDoc":"","perSpool":"1"},"data":[{"type":"html","format":"plain","data":"<html>   <table style=\"font-family: monospace; border: 1px;\">       <tr style=\"height: 6cm;\">           <td valign=\"top\"><h2>*&nbsp; QZ Print Plugin HTML Printing &nbsp;*</h2><span style=\"color: #F00;\">Version:</span> 2.0.0-RC5<br/><span style=\"color: #F00;\">Visit:</span> https://qz.io/</td>           <td valign=\"top\"><img src=\"https://demo.qz.io/assets/img/image_sample.png\"></td>       </tr>   </table></html>"}]},"timestamp":1460438066034,"uid":"fmeu41"}
   [WARN] 2016-04-12 01:14:26,076 @ qz.ws.PrintSocketClient:?
   Bad signature on request
   [INFO] 2016-04-12 01:14:27,320 @ qz.common.TrayManager:?
   Allowed localhost to print to PDF
   [DEBUG] 2016-04-12 01:14:27,322 @ qz.printer.PrintServiceMatcher:?
[...]

Issue #576, opened 2020-02-09
Switch JavaFX to Monocle Implementation


tresf:
Apparently JavaFX 8 introduced formal support for a "headless" back-end for Java FX that they called "Monocle".   I'd like to discuss the possibility of switching QZ Tray over to the "headless" JavaFX "Monocle" Testing framework.
* Faster render times
* No more popups with `{ rasterize: true }`

This technique was discovered while investigating #547, #32.

### Steps:
<details>
<summary>Hidden, see PR #586 instead</summary>

1. Download from Maven Central: https://mvnrepository.com/artifact/org.testfx/openjfx-monocle
2. Copy the jar to `jre/lib/ext` (Is there a cleaner way to do this?)
3. Configure the JVM arguments:
   ```batch
   # Add the following JVM arguments (through IDE or the command line)
   -Dtestfx.robot=glass -Dglass.platform=Monocle -Dmonocle.platform=Headless -Dprism.order=sw
   ```

For more information see this StackOverflow article which explains it much better: https://stackoverflow.com/a/29131967/3196753
</details>

tresf:
Initial tests are fantastic... This lifts the zoom limitations that we hit with our existing implementation... I'll prepare a PR against 2.0.

![image](https://user-images.githubusercontent.com/6345473/74565738-e44ce500-4f3f-11ea-97b8-19fadbc98862.png)

tresf:
> Initial tests are fantastic... 

I spoke too soon... Quoting the JavaFX developer's mailing list... 

> Although first test was positive, when invoking multiple times, I'm getting some internal errors similar to this: https://stackoverflow.com/questions/49388497 and the framework grows slower and slower as it nears its final capture. (I'm using WebView.capture(...))

Fortunately a developer was kind enough to share their results... 

> Hi Tres,
> 
> We also are suffering from this crash when running our TestFX unit tests, particularly on Java 11.
It is due to a concurrency issue between the JavaFX thread and the QuantumRenderer thread and there is an OpenJDK bug here: https://bugs.openjdk.java.net/browse/JDK-8201567
> 
> Quoting from this bug report:
> > “The QuantumRenderer calls the getPixels() method while trying to find a buffer that's not in use, yet in doing so it can inadvertently modify a buffer that's in use."
> 
> There is also a related TestFX Bug: https://github.com/TestFX/Monocle/issues/56
> 
[...]

PR #580, opened 2020-02-11
Fix HTML capture race conditions

### LINKED ISSUE #547: Inconsistent HTML printing
Quoting #531:

> I'm developing a project where I send orders from a APP directly to the kitchen printer in a restaurant, after the payment is confirmed. Everything is working fine but sometimes the "ticket" printed is huge and I can't understand the problem.
> 
> Take a look at this examples:
> On the right you have a regular ticked printed as expected.
> On the left a similar ticket is printed with a huge font.
>
> IMG_6831
> 
> I can reprint the same order multiple times and 1 in 20 tickets is printed in this weird format.
Anyone had a similar issue?
>
> The computer connected with the printer is brand new, running windows 10 and the printer is a TM-T88VI-111

Quoting @klabarge:

> @Read1337 I was able to reproduce your exact issue. I'm looking more into the logs now, and will let you know when we have more information.
>
> My environment:
> 
> **OS:** Windows 10
> **Printer:** Epson TM-T88VI
> **Java:** AdoptOpenJDK 11.0.4 64-bit
> **QZ Tray:** 2.0.11
> 
> I was also able to reproduce this to a virtual printer.
> 
> 1. Setup Signing
> 1. Set Printer Page Size
>    * Right click printer > Printer properties > Prererences > Advanced
>    * Set page size to "PostScript Custom Page Size", make size 80mm x 297mm
> 
> image
> 
> image

tresf:
Creating a PR because initial tests are passing.

Closes #547 #32

### DIFF
--- a/src/qz/printer/action/WebApp.java
+++ b/src/qz/printer/action/WebApp.java
@@ -1,21 +1,18 @@
 package qz.printer.action;
 
-import javafx.animation.PauseTransition;
+import com.sun.javafx.tk.TKPulseListener;
+import com.sun.javafx.tk.Toolkit;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
 import javafx.concurrent.Worker;
 import javafx.embed.swing.SwingFXUtils;
-import javafx.event.ActionEvent;
-import javafx.event.EventHandler;
-import javafx.geometry.Dimension2D;
 import javafx.scene.Scene;
 import javafx.scene.SnapshotParameters;
 import javafx.scene.image.WritableImage;
 import javafx.scene.web.WebView;
 import javafx.stage.Stage;
-import javafx.util.Duration;
 import org.joor.Reflect;
 import org.joor.ReflectException;
 import org.slf4j.Logger;
@@ -27,7 +24,8 @@ import org.w3c.dom.NodeList;
 
 import java.awt.image.BufferedImage;
 import java.io.IOException;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -40,9 +38,6 @@ public class WebApp extends Application {
 
     private static final Logger log = LoggerFactory.getLogger(WebApp.class);
 
-    private static final int SLEEP = 250;
-    private static final int TIMEOUT = 60; //total paused seconds before failing
-
     private static WebApp instance = null;
 
     private static Stage stage;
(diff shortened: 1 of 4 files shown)


src/qz/printer/action/WebApp.java
```diff
@@ -131,28 +141,30 @@ public WebApp() {
...
-            log.trace("Waiting for JavaFX..");
-            try { Thread.sleep(SLEEP); } catch(Exception ignore) {}
-        }
+        if (startupLatch.getCount() > 0) {
+            try {
+                log.trace("Waiting for JavaFX..");
 
-        if (!started.get()) {
-            throw new IOException("JavaFX did not start");
+                if (!startupLatch.await(1000, TimeUnit.MILLISECONDS)) {
```
tresf: I think this should be `60, TimeUnit.SECONDS` to match our existing code of `TIMEOUT = 60; //total paused seconds before failing`

src/qz/printer/action/WebApp.java
```diff
@@ -76,40 +72,43 @@ public void changed(ObservableValue<? extends Worker.State> ov, Worker.State old
...
-                        webView.setMinHeight(pageHeight * pageZoom);
-                        webView.setPrefHeight(pageHeight * pageZoom);
-                        webView.autosize();
-
-                        snap.playFromStart();
+                    public void run() {
+                        log.debug("Attempting image capture");
+
+                        SnapshotParameters sparams = new SnapshotParameters();
+                        sparams.setTransform(new Scale(pageZoom, pageZoom));
```
tresf: On Java 8 and Java 11 using 4x zoom, it appears to just make the picture bigger without improving quality... :(

image

tresf:
@bberenz Tests are looking great so far.

Some requests:
* Avoid the color white as a background color
* Test the bottom right pixel to make sure the render is complete

Here's Java 11 + Gluon on macOS.  I can't find an issue at a glance with any of the renders.

ezgif-2-7ffa84ff649b

tresf:
I just realized that the zoom is still using the `SnapshotParameters` logic, which is not correct. https://github.com/qzind/tray/pull/580#discussion_r377469631 so my test results are not usable.

src/qz/printer/action/WebApp.java
tresf: @bberenz bump as I don't think this snippet was ever resolved.

bberenz: I thought I addressed it at some point, but it must have gotten reverted in one of the rollbacks. Taken care of now.

klabarge:
I sent over 300 print jobs and all print jobs completed. I didn't hit the race condition using this branch

However, I did notice that the HTML rendering is much more noticeable 

Screen Recording 2020-03-24 at 8 48 13 PM

tresf:
> However, I did notice that the HTML rendering is much more noticeable

I tried to fix it by making the stage smaller, but it results in more partial snapshots.  I guess that won't improve until #576 is completed.  🤞

Merging.

Issue #526, opened 2019-11-27
Whitelist qz protocol handler Google Chrome


tresf:
Chrome 77 removed the checkbox that allows whitelisting of a URL handler.  More information here: ~~https://support.google.com/chrome/thread/14194567?msgid=15347344~~
https://support.google.com/chrome/thread/14193532?hl=en

Starting with the `certs-rewrite` branch, we have more granular control over installation steps (they'e done by Java now).

<details>
<summary>Original proposal (changed in comments below)</summary>

Starting with the `certs-rewrite` branch, we have a JSON writer we can use to set this preference as well as a dedicated area for setting up user settings on launch.  We can probably shim this setting in.

**Windows:** `%LOCALAPPDATA%\Google\Chrome\User Data\Default\Preferences`
**Mac:** \<TBD\>
**Linux:** \<TBD\>
```js
"protocol_handler":{"excluded_schemes":{"qz":false}}
```
</details>

tresf:
Found the following recommendation on a [Google Help Forum](https://support.google.com/chrome/thread/14322141?hl=en):


> You can manually whitelist applications/protocol by adding a registry key for it (windows-only)
> 
> Create a txt document with the info below in it and import into the registry. The users will no longer see the prompt. 
> 
> ```reg
> Windows Registry Editor Version 5.00
> [HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Google\Chrome\URLWhitelist]
> "13"="ms-word://*"
> ```

Tested on Windows 10 swapping out `"13"="ms-word://*"`with `"0"="qz://*"` and it's working.  Will work on a patch now.

tresf:
After some digging, found some documentation for Mac (potentially): https://www.chromium.org/administrators/mac-quick-start

```bash
sudo -s
cd /Library/Managed Preferences/<username>
plutil -convert xml1 com.google.Chrome.plist
cat com.google.Chrome.plist
```

tresf:
Support for Windows and MacOS has been added in #582.  Linux will require a bit more investigation.

tresf:
Linux seems to honor the following per: https://github.com/chromium/chromium/blob/b4511255f9592a4aeee1d6d2e10db07e80c157df/chrome/common/chrome_paths.cc#L496-L500

```bash
sudo mkdir -p /etc/opt/chrome/policies/managed
echo '{ "URLWhitelist": ["qz://*"] }' > /etc/opt/chrome/policies/managed/qz-tray.json
[...]

PR #582, opened 2020-02-11
Whitelist qz protocol handler for Google Chrome

tresf:
Per #526, adds `qz:launch` support to Chrome, Chromium by whitelisting the `qz:` URL on Windows, Mac and Linux.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="EntryPointsManager">
+    <entry_points version="2.0" />
     <list size="4">
       <item index="0" class="java.lang.String" itemvalue="org.eclipse.jetty.websocket.api.annotations.OnWebSocketClose" />
       <item index="1" class="java.lang.String" itemvalue="org.eclipse.jetty.websocket.api.annotations.OnWebSocketConnect" />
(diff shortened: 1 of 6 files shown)


klabarge:
This works on macOS and Windows. I found that Chrome does need to be restarted for it to take effect.

This doesn't seem to work on Ubuntu 18.0.4 64-bit

- Chrome Version 80.0.3987.106 (Official Build) (64-bit)
- I installed Chrome using `sudo dpkg -i google-chrome-stable_current_amd64.deb`

I checked both of these directories `/etc/chromium/policies/managed`, `/etc/opt/chrome/policies/managed` and the `qz-tray.json` file does exist

image

tresf:
They should exist, perhaps with the wrong permissions?

I'm using `PosixFilePermissions.fromString("rwxr-xr-x")` but they don't seem to be taking on a fresh machine.  I'll see what's going on...

https://github.com/qzind/tray/blob/f945c255da1ce2d829d330643ba79a19a423088c/src/qz/installer/LinuxInstaller.java#L110-L117

tresf:
Ugh... It's the `umask`.... https://stackoverflow.com/a/44923969/3196753

tresf:
@klabarge ok, I've patched the permissions per 28dd5bc.  Linux is ready for testing.

klabarge:
Tested on Ubuntu and the patch works 👍

PR #589, opened 2020-02-22
Add printer port information

### LINKED ISSUE #588: Get printer port information
A customer has requested that they can obtain the IP address of a printer.

Most of the time this info is available in the following locations:
* Windows: In registry as `Port: 192.168.1.111`
* Mac/Linux: Through `lpoptions` as `device-uri=socket://192.168.1.111`

Note, this information is not very useful when using a WSD/Bonjour port.

tresf:
Closes #588

### DIFF
--- a/sample.html
+++ b/sample.html
@@ -1393,6 +1393,7 @@
                 list += "<li>" + (data[i].default ? "* " : "") + data[i].name + "<ul>" +
                         "<li><strong>Driver:</strong> " + data[i].driver + "</li>" +
                         "<li><strong>Density:</strong> " + data[i].density + "dpi</li>" +
+                        "<li><strong>Connection:</strong> " + data[i].connection + "</li>" +
                         (data[i].trays ? "<li><strong>Trays:</strong> " + data[i].trays + "</li>" : "") +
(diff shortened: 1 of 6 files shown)


tresf:
@klabarge can you do a final review and merge?

klabarge:
@tresf I just tested on Ubuntu 18.0.4

Printer installed through CUPS returns **Connection:** socket://192.168.0.54

I'll test on macOS next.

klabarge:
macOS looks fine.

```
Connection: ipp://192.168.0.54/ipp/print
```

Note that the connection information doesn't look pretty when the printer is not connected via the network

```
EPSON WF-3730 Series
    Connection: dnssd://EPSON%20WF-3730%20Series._ipps._tcp.local./?uuid=cfe92100-67c4-11d4-a45f-9caed3c8434a
```

This is good to merge if we're okay with my last note.

tresf:
> Note that the connection information doesn't look pretty when the printer is not connected via the network

Yeah, the autoconfigure printers won't have a usable address, that's to be expected.

> This is good to merge if we're okay with my last note.

👍

PR #592, opened 2020-02-25
Asymmetric density

bberenz:
Adds support for asymmetric densities and named density values.

### DIFF
--- a/src/qz/printer/info/NativePrinter.java
+++ b/src/qz/printer/info/NativePrinter.java
@@ -152,11 +152,14 @@ public class NativePrinter {
     }
 
     public List<PrinterResolution> getResolutions() {
-        // TODO: Test/Implement supported resolutions for CUPS
-
         PrintService ps = getPrintService().value();
         PrinterResolution[] resSupport = (PrinterResolution[])ps.getSupportedAttributeValues(PrinterResolution.class, ps.getSupportedDocFlavors()[0], null);
         if (resSupport == null || resSupport.length == 0) {
+            NativePrinterMap printerMap = NativePrinterMap.getInstance();
+            // CUPS doesn't report resolutions properly, instead return the values scraped from console
+            if(printerMap instanceof CupsPrinterMap) {
+                return ((CupsPrinterMap)printerMap).getResolutions(this);
+            }
             resSupport = new PrinterResolution[]{ getResolution().value() };
         }
 
--- a/src/qz/printer/info/NativePrinter.java
+++ b/src/qz/printer/info/NativePrinter.java
@@ -3,10 +3,9 @@ package qz.printer.info;
 import qz.utils.SystemUtilities;
 
 import javax.print.PrintService;
-import javax.print.attribute.ResolutionSyntax;
 import javax.print.attribute.standard.PrinterName;
 import javax.print.attribute.standard.PrinterResolution;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 public class NativePrinter {
@@ -152,22 +151,16 @@ public class NativePrinter {
         return resolution;
     }
 
-    public List<Integer> getResolutions() {
+    public List<PrinterResolution> getResolutions() {
         // TODO: Test/Implement supported resolutions for CUPS
-        List<Integer> densities = new ArrayList<>();
 
         PrintService ps = getPrintService().value();
         PrinterResolution[] resSupport = (PrinterResolution[])ps.getSupportedAttributeValues(PrinterResolution.class, ps.getSupportedDocFlavors()[0], null);
         if (resSupport == null || resSupport.length == 0) {
             resSupport = new PrinterResolution[]{ getResolution().value() };
         }
-        for(PrinterResolution res : resSupport) {
-            if (res != null) {
-                densities.add(res.getFeedResolution(ResolutionSyntax.DPI));
-            }
-        }
 
-        return densities;
+        return Arrays.asList(resSupport);
     }
 
     public static void getDriverAttributes(NativePrinter printer) {
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1099,8 +1099,10 @@ var qz = (function() {
              *   @param {number} [options.bounds.height=0] Height of bounding box
              *  @param {string} [options.colorType='color'] Valid values <code>[color | grayscale | blackwhite]</code>
              *  @param {number} [options.copies=1] Number of copies to be printed.
-             *  @param {number|Array<number>} [options.density=0] Pixel density (DPI, DPMM, or DPCM depending on <code>[options.units]</code>).
+             *  @param {number|Array<number>|Object|Array<Object>} [options.density=0] Pixel density (DPI, DPMM, or DPCM depending on <code>[options.units]</code>).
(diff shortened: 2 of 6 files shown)


src/qz/printer/info/NativePrinter.java
```diff
@@ -152,22 +151,16 @@ public void setResolution(PrinterResolution resolution) {
         return resolution;
     }
 
-    public List<Integer> getResolutions() {
+    public List<PrinterResolution> getResolutions() {
         // TODO: Test/Implement supported resolutions for CUPS
```
tresf: Unfortunately, due to this `TODO:` item and lack of density reporting from CUPS, we can't go higher than what's reported in the `.ppd` file as `*DefaultResolution: 300dpi` on my Dymo LabelWriter 450 Turbo on Mac.

The Dymo `.ppd` comes back with the following when parsing for `dpi`:

```diff
+ *DefaultResolution: 300dpi
  *Resolution 300dpi/300 DPI: "<</HWResolution[300 300]>>setpagedevice"
  *Resolution 300x600dpi/300x600 DPI: "<</HWResolution[300 600]>>setpagedevice"
  *UIConstraints: *DymoPrintQuality Text       *Resolution       300x600dpi
  *UIConstraints: *DymoPrintQuality Graphics   *Resolution       300dpi
  *UIConstraints: *Resolution       300x600dpi *DymoPrintQuality Text 
  *UIConstraints: *Resolution       300dpi     *DymoPrintQuality Graphics
```

I wonder if we should include better multi-res parsing as part of this PR.  What do you think @bberenz?  If so, do you have some printers to scrape info from?  @klabarge if you can offer some info too, that would be great.

tresf: The command is:

```bash
cat "$(lpstat -l -p |grep DYMO|grep ppd|cut -d' ' -f2)" |grep dpi
```

Just replace `DYMO` with a unique printer name.

tresf: My Zebra comes back with the following:

```diff
+ *DefaultResolution: 203dpi
  *Resolution 203dpi/203dpi: "<</HWResolution[203 203]/cupsBitsPerColor 1/cupsRowCount 0/cupsRowFeed 0/cupsRowStep 0/cupsColorSpace 3>>setpagedevice"
  *Resolution 300dpi/300dpi: "<</HWResolution[300 300]/cupsBitsPerColor 1/cupsRowCount 0/cupsRowFeed 0/cupsRowStep 0/cupsColorSpace 3>>setpagedevice"
  *Resolution 600dpi/600dpi: "<</HWResolution[600 600]/cupsBitsPerColor 1/cupsRowCount 0/cupsRowFeed 0/cupsRowStep 0/cupsColorSpace 3>>setpagedevice"
```

tresf: My Epson WF4740 Workforce Pro shows the following (only returned one line so grep'd for `600` instead:

```diff
+ *DefaultResolution: 600x600dpi
  *cupsPrintQuality Normal: "<</cupsInteger1 4 /HWResolution[600 600]>>setpagedevice"
  *cupsPrintQuality High: "<</cupsInteger1 5 /HWResolution[600 600]>>setpagedevice"
```

klabarge: > `cat "$(lpstat -l -p |grep DYMO|grep ppd|cut -d' ' -f2)" |grep dpi`

Some of the `ppd` files on my system didn't have `dpi`. I had better luck `grepping` for "Resolution".

#### Epson WF 3730
```diff
+ *DefaultResolution: 600x600dpi
*cupsPrintQuality Normal: "<</cupsInteger1 4 /HWResolution[600 600]>>setpagedevice"
*cupsPrintQuality High: "<</cupsInteger1 5 /HWResolution[600 600]>>setpagedevice"
```

#### HP LaserJet 500 Color MFP M570
This printer didn't have a `DefaultResolution` nor anything with `dpi`

```diff
/HWResolution [600 600] /PreRenderingEnhance false >> setpagedevice"
/HWResolution [600 600] /PreRenderingEnhance false >> setpagedevice"
```

#### HP LaserJet M603
This printer didn't have a `DefaultResolution`
```diff
*HPPrintQuality FastRes1200/FastRes 1200: "<</HWResolution [600 600] /PreRenderingEnhance true>> setpagedevice"
+ *HPPrintQuality 600dpi/600 dpi: "<</HWResolution [600 600] /PreRenderingEnhance false>> setpagedevice"
*HPPrintQuality ProRes1200/ProRes 1200: "<</HWResolution [1200 1200] /PreRenderingEnhance false>> setpagedevice"
```

#### Zebra LP2844
```diff
+ *DefaultResolution: 203dpi
*Resolution 203dpi/203dpi: "<</HWResolution[203 203]/cupsBitsPerColor 1/cupsRowCount 0/cupsRowFeed 0/cupsRowStep 0/cupsColorSpace 3>>setpagedevice"
*Resolution 300dpi/300dpi: "<</HWResolution[300 300]/cupsBitsPerColor 1/cupsRowCount 0/cupsRowFeed 0/cupsRowStep 0/cupsColorSpace 3>>setpagedevice"
*Resolution 600dpi/600dpi: "<</HWResolution[600 600]/cupsBitsPerColor 1/cupsRowCount 0/cupsRowFeed 0/cupsRowStep 0/cupsColorSpace 3>>setpagedevice"
```

tresf:
> Unfortunately, due to this `TODO:` item and lack of density reporting from CUPS, we can't go higher than what's reported in the `.ppd` file as `*DefaultResolution: 300dpi` on my "Dymo LabelWriter 450 Turbo" on Mac.

I've added cross-feed parsing support to CUPS as well as additional resolution support.

@bberenz it appears you assume the first density is the best density e.g. `bestRes = rSupport.get(0);`.  Are you sorting them as you're adding them?  If so, can you help me do the same in my latest code?

tresf:
... also @bberenz the `best` keyword isn't being honored with this new CUPS logic for some reason.

bberenz:
>  it appears you assume the first density is the best density

No, it only grabs the first resolution to ensure not null as it then [loops through](https://github.com/qzind/tray/pull/592/files#diff-a3fee03485d53a539241fc505374f46aR158) the whole list comparing for better (or worse if draft)

tresf:
> No, it only grabs the first resolution to ensure not null as it then loops through the whole list comparing for better (or worse if draft)

* Fixed an order-of-operations issue where the size of the densities array is zero; done by using an empty `ArrayList` to signify whether or not the densities have been fetched.  This may need eventual improvement down the road to address #607 but should fix this PR.
* Changed the logic to use `null` as default and then set on first loop instead; guards against `IndexOutOfBounds` exception in the event CUPS can't parse the densities resulting in nothing at `.get(0)`.

tresf:
Fixed a parsing issue with `/HWResolution` where it assumed metric.  This should be ready to test.  Since many printers don't offer asymmetric densities, I'll be testing this on my Dymo Labelwriter 450, which has a density of 300x600dpi.

tresf:
Testing is positive but also a bit inconclusive, the quality seems to remain the same between 300x300 and 600x300 on this printer, which seems rather impossible.  Perhaps the driver reports a higher resolution than what is possible, but it does not appear to hinder the output, so I'm merging.

PR #608, opened 2020-03-17
Prompt for Java install if missing

### LINKED ISSUE #606: Installer doesn't warn about missing Java
As @lite1979 discovered testing #10, in 2.1.0, the Windows installer doesn't appear to warn when Java is missing, causing confusing to end-users.

tresf:
Closes #606

### DIFF
--- a/ant/apple/apple-preinstall.sh.in
+++ b/ant/apple/apple-preinstall.sh.in
@@ -7,4 +7,11 @@ set -e
 DIR=$(cd "$(dirname "$0")" && pwd)
 pushd "$DIR/payload/Contents/MacOS/"
 
+# Offer to download Java if missing
+if ! ./"${project.name}" --version; then
+    osascript -e "tell app \"Installer\" to display dialog \"Java is required.  Please install Java and try again.\""
+    open "${java.download}"
+    exit 1
+fi
+
 ./"${project.name}" preinstall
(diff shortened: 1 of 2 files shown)


klabarge:
On macOS, after clicking "OK", the AdoptOpenJDK installer page opened.

However, it opened the webpage in Safari, despite my default browser being Firefox

image

klabarge:
Windows passed testing.

tresf:
> However, it opened the webpage in Safari, despite my default browser being Firefox

Looks like it's using system defaults since `root` doesn't have one.  Fixed it to spawn as the current user per `93933aa`, can you try it again>

tresf:
Hmm... testing failed, it still uses Safari.  Sorry, @klabarge we'll just have to live with this behavior for now.  There are some [ugly ways to detect the user](https://scriptingosx.com/2019/09/get-current-user-in-shell-scripts-on-macos/), but I'm not comfortable with adding that much logic for a small nuance like this.

tresf:
Ok, I think I got macOS using the proper browser now using 0fb61c2.   @klabarge can you test again?  Also, can you double-check that Windows isn't launching the browser as Administrator?  That seems like it might open a can of worms.

klabarge:
>  think I got macOS using the proper browser now using 0fb61c2

Works 👍 

> can you double-check that Windows isn't launching the browser as Administrator?

I don't think so, because it opened up in my existing browser session as a new tab

PR #615, opened 2020-04-11
Separate system tray and window theming logic on Windows

### LINKED ISSUE #597: Tray icon wrong color Windows 10
I think we have some of our logic backwards, the tray icon on Windows 10 sometimes doesn't match the theme.  I believe it's caused on a default Windows 10 install because when I click the theme options it seems to be corrected.

tresf:
Closes #597


tresf:
Tested, pass:
* Windows 10 1909 (with Light theme support)
* Windows 10 1909 (with old bugged registry keys)
* Windows 10 1507 (without Light theme support)
* Windows 7 x64
* Windows XP

PR #633, opened 2020-05-08
Bump Apple cert

tresf:
Apple Installer cert is slated to expire in June.  This PR adds a new 2025 cert to the build system, ~~removing the old one by sh256 hash if present.~~

This needs to be refactored upon 2.0 as well.


tresf:
Note, anyone creating signed packages will need to execute the following command to force the build system to prefer the 2025 cert:

```bash
security delete-certificate -Z "A69020D49B47383064ADD5779911822850235953"
```

I've removed this from the codebase because it only needs to be executed once, then `ant pkgbuild` the HEAD of `master` or `2.0` branch.

PR #643, opened 2020-05-21
Initial draft of Sec-CH-UA* hints

### LINKED ISSUE #552: Add Client Hints
Future versions of Chrome aim to block User Agent Strings and leverage Client Hints instead.

The "Device Hints" are especially important for OS metrics as well as our downloads page.

https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/client-hints
https://developer.mozilla.org/en-US/docs/Glossary/Client_hints
https://www.keycdn.com/blog/client-hints

tresf:
Must be resolved prior to merge:
* [x] The current draft says `x86` is allowed,but `x86_64` is not.  I've asked why in the (resolved) comments of https://github.com/WICG/ua-client-hints/pull/106, previously asked here https://github.com/WICG/ua-client-hints/issues/105
* [x] I'm not sure where to put the Java version information.  I've asked about this too here: https://github.com/WICG/ua-client-hints/issues/111

Closes #552.

### DIFF
--- a/src/qz/utils/ConnectionUtilities.java
+++ b/src/qz/utils/ConnectionUtilities.java
@@ -69,7 +69,11 @@ public final class ConnectionUtilities {
             requestProps.put("Sec-CH-UA-Platform-Version", getPlatformVersion());
             requestProps.put("Sec-CH-UA-Arch", getArch());
             requestProps.put("Sec-CH-UA-Full-Version", Constants.VERSION.toString());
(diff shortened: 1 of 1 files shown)


tresf:
Well, `ua-client-hints` isn't being very receptive of the need for distinction between `x86` and `x86_64` with the WebKit developer outright refusing to provide hardware platform at all.  Since I'm not interested in sparking any controversy (we're just trying to leverage and honor the new standard, not influence it) I'm going to leave the architecture as-is and hope analytics platforms are OK with it.  🍻

Issue #646, opened 2020-05-28
HTML scaling using version 2.1 with Monocle patch


klabarge:
`scaleContent: true` isn't scaling HTML content correctly in 2.1+59. Test results are the same whether using Monocle or not. To reproduce, use `size: {width: 4, height: 6}`

**QZ Tray:** [2.1+59](https://github.com/tresf/tray/releases/tag/v2.1.1)
**Java:** AdoptOpenJDK 11.0.3 and Oracle Java 8u211
**OS:** Tested using macOS and Windows.

`rasterize: true`
<img width="356" alt="image" src="https://user-images.githubusercontent.com/12505463/83087158-94ada780-a05e-11ea-8b69-7ec3ec48be3c.png">

`rasterize: false`
<img width="356" alt="image" src="https://user-images.githubusercontent.com/12505463/83087453-5e245c80-a05f-11ea-958e-bf75b4cdedbe.png">

**Expected results, using the official 2.1 release**
<img width="356" alt="image" src="https://user-images.githubusercontent.com/12505463/83088118-e0f9e700-a060-11ea-8ca5-e7368e331a37.png">

tresf:
@bberenz I think we may have factored-out `WEB_SCALE` in a few areas where it's still needed.  Could that be the cause?

PR #649, opened 2020-05-29
2.1 html scaling

bberenz:
Fixes for the problems outlined in #646


tresf:
@bberenz thanks for the quick turn-around.  Unfortunately with the vector unit-tests something's causing the pages to become larger than 8.5x11:

Logs here: https://gist.github.com/tresf/56eafe254b86f7d129553af174c3087d

image

PR #661, opened 2020-06-05
Add backwards compatible for base64 URLs

### LINKED ISSUE #660: Cannot parse (PLAIN)data:image/png;base64,iVBOR...
It appears we don't have an upgrade in 2.1 to handle 2.0's base64 raw images.  @bberenz can you look into this?  Basically, the 2.0 `data:image/png;base64` hack won't upgrade to 2.1's desktop client.  We do have specific upgrades and downgrades in place in the JavaScript file, but I think Java should be able to handle this with the dated JS as well.

tresf:
Closes #660


tresf:
* ✅ PASS: Tested 2.0.12's `sample.html`
* ✅ PASS: Regression tested 2.1.0's `sample.html`

PR #669, opened 2020-06-09
Add backwards compatibility for new HID, USB, Serial API

tresf:


tresf:
Testing looks good with this patch.  Merging.

image

PR #676, opened 2020-06-23
Raw HTML density support

### LINKED ISSUE #675: Raw Raster HTML Broken
Since the the merge of #625, raw raster html is broken.

Steps to reproduce:

1. Open sample.html
2. From the raw tab try the "HTML" button in the "Raster Print" section
   > 

3. Observe the logs, a `1x1` image is printed
    >```
    >[INFO] 2020-06-17 13:19:06,983 @ qz.printer.ImageWrapper:95
    >    Loading BufferedImage
    >[INFO] 2020-06-17 13:19:06,983 @ qz.printer.ImageWrapper:96
    >    Dimensions: 1x1
    >```

The problem seems to be related to the zoom:

```java
pageZoom = model.getZoom();
System.out.println(pageZoom); // Prints "NaN"
```

The following patch produces an image, albeit not the right quality:
```diff
  pageZoom = model.getZoom();
+ if(Double.isNaN(pageZoom)) {
+     pageZoom = 1;
+ }
```

tresf:
Closes #675

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1264,7 +1264,7 @@ var qz = (function() {
              *
              * @param {Object} options Default options used by printer configs if not overridden.
              *
-             *  @param {Object} [option.bounds=null] Bounding box rectangle.
+             *  @param {Object} [options.bounds=null] Bounding box rectangle.
              *   @param {number} [options.bounds.x=0] Distance from left for bounding box starting corner
              *   @param {number} [options.bounds.y=0] Distance from top for bounding box starting corner
              *   @param {number} [options.bounds.width=0] Width of bounding box
--- a/src/qz/printer/LanguageType.java
+++ b/src/qz/printer/LanguageType.java
@@ -20,21 +20,23 @@ import java.util.List;
  */
 public enum LanguageType {
 
-    ZPL(false, true, "ZPL", "ZPL2", "ZPLII", "ZEBRA"),
-    EPL(true, true, "EPL", "EPL2", "EPLII"),
-    CPCL(false, true),
-    ESCP(false, false, "ESCP", "ESCP2", "ESCPOS", "ESC", "ESC/P", "ESC/P2", "ESCP/P2", "ESC/POS", "ESC\\P", "EPSON"),
-    EVOLIS(false, false),
-    UNKNOWN(false, false);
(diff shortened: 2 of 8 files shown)


tresf:
@bberenz:

If I have a 203 DPI printer and want to print a full web page to it, what dimensions should I provide?
I assume: 
1. 203 will be calculated automatically by our new API
2. Web dpi is assumed 72dpi, will be converted on render
3. If it's a 4x6, dimensions should be 4*72, 6*72, no?

Because when I test raw-zoom using those values using the EPL2 HTML button, it's too big for a 4x6 label.

bberenz:
There was some issue with scaling after adding in the zoom and the adjusted dpi's that should now be accounted for.
As for which dimension to use, for now its going to be based on the dpi used by the printer. So for a 203dpi printer, you would need to supply 812 x 1218.
At some point it will probably be updated to match pixel printing where the values would be in the selected units, but as you mentioned there's some backwards compatibility that needs to be considered.

lite1979:
This looks to be working correctly with respect to the desired 203dpi ZPL printer results. 

As was previously observed, only the first "page", or 4x6 inch label, is printed.

In thermal receipt tests, the media is 80mm wide with a "limitless" height due to the continuous roll. 

I used 3.14x180dpi for a width of 565, which may be just outside the printable area of the Epson TM-T88V, as can be seen by the right side missing a little bit of content. 

The same width of 565, applied to a Star TSP 143, which has a 203 dpi printhead, shows the same result with the right side missing some content. 

Changing the width to 520 allows the entire width of the page to be printed on the media, though if we're defaulting to 180dpi for all ESC/POS printers, the image will be shrunk proportionately for printers that use 203 dpi, such as the Star, Citizen CBM series, Epson T20, and other clones.

The same logic should apply to Zebra printers or clones that happen to be 300dpi instead of 203 dpi, though the difference will be much easier to spot with the naked eye than 203 vs. 180. Can these dpi values for HTML to raw image be customized by our clients?

IMG_20200709_185353543
IMG_20200709_183545312
IMG_20200709_183224141

tresf:
> Can these dpi values for HTML to raw image be customized by our clients?

Yes, you can provide/override density as a value -- as you would for Pixel -- and it will be honored per:

https://github.com/qzind/tray/blob/ce6f873316f5e8100ebbca261b1f8e1eab2322ce/src/qz/printer/action/PrintHTML.java#L83

PR #686, opened 2020-07-02
Tray list checks by instance

### LINKED ISSUE #679: Trays aren't properly listing
QZ Tray currently filters for the word "Tray" when fetching printer trays:

https://github.com/qzind/tray/blob/1781b104314484a83ea1a027b5c65d2d717d8d94/src/qz/printer/PrintServiceMatcher.java#L134-L136

This has serious limitations.  For starters, Dutch calls "Tray 1" "Lade 1" in localized printer drivers.

Second, some printers refer to Trays as a "Cartridge" or even "Auto Select".

QZ Tray should be able to list all trays, regardless of their name.  Instead of filtering by "Tray", we may be able to filter by "MediaTray" type:

```java
for(Media m : (Media[])ps.getSupportedAttributeValues(Media.class, null, null)) { 
     if (m instanceof MediaTray) { jsonService.accumulate("trays", m.toString()); } 
 } 
```

tresf:
Closes #679

### DIFF
--- a/src/qz/printer/PrintServiceMatcher.java
+++ b/src/qz/printer/PrintServiceMatcher.java
@@ -23,6 +23,7 @@ import javax.print.PrintService;
 import javax.print.PrintServiceLookup;
 import javax.print.attribute.ResolutionSyntax;
 import javax.print.attribute.standard.Media;
+import javax.print.attribute.standard.MediaTray;
 import javax.print.attribute.standard.PrinterName;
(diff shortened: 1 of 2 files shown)


tresf:
@lite1979 tested this today:

> When I set tray to "Manual" it works. If the front panel is closed, the printer prompts me to open it for Manual feed.

Also lists items on my printer such as "Auto Select" now, which is an improvement over previous versions.  Merging.

PR #689, opened 2020-07-14
Printer status special character support

### LINKED ISSUE #668: Printer status should send human-friendly printer name
Quoting:

> When using the various commands in the qz tray javascript library against the latest QZ Tray 2.1 version (v2.1.0) application on a 2015 MacBook Pro (macOS Mojave 10.14.6 (18G4032)), I am seeing some inconsistency with how printer names are being returned. For example, the `qz.printers.find()` command returns a list of printers that looks like this on my machine:

> ```json
> [
>    "Canon ABC123/456",
>    "Canon ABC123/456 - Fax"
> ]
> ```
>
> (as you can see, it includes special characters and whitespace)
>
> However, when receiving events via `qz.printers.setPrinterCallbacks()`, the event objects look like this:
> 
>```json
> {
>    "printerName": "Canon_ABC123_456",
>    "statusCode": 0,
>    "statusText": "OK",
>    "severity": "INFO",
>    "cupsString": "none",
>    "message": "OK: Level INFO, StatusCode 0, From Canon_ABC123_456, CUPS string none",
>    "type": "ACTION"
>}
>```
>```json
> {
>    "printerName": "Canon_ABC123_456___Fax",
>    "statusCode": 0,
>    "statusText": "OK",
>    "severity": "INFO",
>    "cupsString": "none",
>    "message": "OK: Level INFO, StatusCode 0, From Canon_ABC123_456___Fax, CUPS string none",
>    "type": "ACTION"
>}
>```
>
> (notice how the `printerName` value has whitespace and special characters replaced with underscores)
>
>
> To fix this, I had just been using regular expressions to match the same replacement behavior (convert all non-word characters to underscores) in order to do comparisons as needed. However, a colleague running on Windows noticed that it seems like the event object's `printerName` values do include special characters, so my comparison logic was breaking in that environment.

Vzor-:
fixes #668 

Ready for merge.

Fixes:
- Mac os printer names reported by status listeners now agree with the names given by 'find all printers'.
- Fixed an issue where '&' was breaking the status reporting on Ubuntu (and probably mac)
- Some erroneous html escaping in sample.html

Known Issues: 
- \ ` ^ +  and probably some other symbols break the jfx PrintServiceLookup, this issue also affects "find all printers" 
- ~~Changing printer names on mac doesn't affect the names returned by status until qz restarts. This is due to there being no warning that a printer's description was changed, and renaming a printer on mac only changes the description.~~

Todo:
- ~~Investigate solution to mac renaming issue~~
- ~~Test changes on windows~~

### DIFF
--- a/src/qz/printer/status/CupsStatusHandler.java
+++ b/src/qz/printer/status/CupsStatusHandler.java
@@ -49,6 +49,8 @@ public class CupsStatusHandler extends AbstractHandler {
 
         while(eventReader.hasNext() && running) {
             XMLEvent event = eventReader.nextEvent();
+            //Todo Remove this debugging log
+            log.warn(event.toString());
             switch(event.getEventType()) {
                 case XMLStreamConstants.START_ELEMENT: {
                     StartElement startElement = event.asStartElement();
--- a/src/qz/printer/status/CupsStatusHandler.java
+++ b/src/qz/printer/status/CupsStatusHandler.java
@@ -45,7 +45,7 @@ public class CupsStatusHandler extends AbstractHandler {
     private void parseXML(XMLEventReader eventReader) throws XMLStreamException {
         boolean isDescription = false, isGuid = false, isFirstGuid = true, running = true;
         String firstGuid = "";
-        String description = "";
+        String eventDescription = "";
 
         while(eventReader.hasNext() && running) {
             XMLEvent event = eventReader.nextEvent();
@@ -53,10 +53,9 @@ public class CupsStatusHandler extends AbstractHandler {
                 case XMLStreamConstants.START_ELEMENT:
                     StartElement startElement = event.asStartElement();
                     String qName = startElement.getName().getLocalPart();
-                    //This is the description of the rss message, NOT the description of the printer
                     if ("description".equalsIgnoreCase(startElement.getName().getLocalPart())) {
                         isDescription = true;
-                        description = "";
+                        eventDescription = "";
                     }
                     if ("guid".equalsIgnoreCase(qName)) {
                         isGuid = true;
@@ -71,7 +70,7 @@ public class CupsStatusHandler extends AbstractHandler {
                 case XMLStreamConstants.CHARACTERS:
                     Characters characters = event.asCharacters();
                     if (isDescription) {
-                        description += characters.getData();
+                        eventDescription += characters.getData();
                     }
                     if (isGuid) {
                         String guid = characters.getData();
@@ -83,7 +82,7 @@ public class CupsStatusHandler extends AbstractHandler {
                             running = false;
                             break;
                         } else {
-                            String printerName = StringUtils.substringBeforeLast(description, "\"");
+                            String printerName = StringUtils.substringBeforeLast(eventDescription, "\"");
                             printerName = StringUtils.substringAfter(printerName, "\"");
                             printerName = StringEscapeUtils.unescapeXml(printerName);
                             if (!printerName.isEmpty() && StatusMonitor.isListeningTo(printerName)) {
--- a/src/qz/printer/status/CupsStatusHandler.java
+++ b/src/qz/printer/status/CupsStatusHandler.java
@@ -43,7 +43,7 @@ public class CupsStatusHandler extends AbstractHandler {
     }
 
     private void parseXML(XMLEventReader eventReader) throws XMLStreamException {
-        boolean isDescription = false, isGuid = false, isFirstGuid = true, running = true;
+        boolean isEventDescription = false, isGuid = false, isFirstGuid = true, running = true;
         String firstGuid = "";
         String eventDescription = "";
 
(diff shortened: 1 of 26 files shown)


src/qz/printer/status/CupsStatusHandler.java
```diff
@@ -45,21 +50,30 @@ private void parseXML(XMLEventReader eventReader) throws XMLStreamException {
...
+                        description = "";
                     }
                     if ("guid".equalsIgnoreCase(qName)) {
                         isGuid = true;
                     }
                     break;
+                }
+                case XMLStreamConstants.END_ELEMENT: {
+                    EndElement endElement = event.asEndElement();
+                    String qName = endElement.getName().getLocalPart();
```
tresf, line 67: If you avoid double `qName` declaration, you can remove the `{}` brackets:

```java
String startElement = event.asEndElement().getName().getLocalPart();
```

```java
String endElement = event.asEndElement().getName().getLocalPart();
```

or even simpler:

```java
if ("description".equalsIgnoreCase(event.asEndElement().getName().getLocalPart())) {
    isDescription = false;
}
```

src/qz/printer/status/CupsStatusHandler.java
```diff
@@ -73,6 +87,7 @@ private void parseXML(XMLEventReader eventReader) throws XMLStreamException {
                         } else {
                             String printerName = StringUtils.substringBeforeLast(description, "\"");
                             printerName = StringUtils.substringAfter(printerName, "\"");
+                            printerName = StringEscapeUtils.unescapeXml(printerName);
```
tresf, line 90: These lines can be combined:
```java
printerName = StringEscapeUtils.unescapeXml(StringUtils.substringAfter(printerName, "\""));
```

src/qz/printer/status/PrinterStatus.java
```diff
@@ -54,7 +59,7 @@ public static PrinterStatus getFromCupsString(String reason, String issuingPrint
     }
 
     public String toString() {
-        String returnString = type.getName() + ": Level " + type.getSeverity() + ", StatusCode " + type.getCode() + ", From " + issuingPrinterName;
+        String returnString = type.getName() + ": Level " + type.getSeverity() + ", StatusCode " + type.getCode() + ", From " + issuingPrinterDescription;
```
tresf, line 62: To avoid confusion please keep `issuingPrinterName` in the logs.  If description is desired, add it as an additional field.

src/qz/printer/status/StatusMonitor.java
```diff
@@ -73,27 +77,39 @@ public synchronized static void closeNotificationThreads() {
     }
 
     public synchronized static boolean startListening(SocketConnection connection, JSONArray printerNames) throws JSONException {
+
+        if (SystemUtilities.isMac()) {
```
tresf, line 81: Can't this be removed, since we check it later?

src/qz/printer/status/StatusMonitor.java
```diff
@@ -165,6 +181,17 @@ public synchronized static boolean isListeningTo(String PrinterName) {
     public synchronized static void statusChanged(PrinterStatus[] statuses) {
         HashSet<SocketConnection> connections = new HashSet<>();
         for(PrinterStatus ps : statuses) {
+            if (isMac()) {
+                NativePrinter nativePrinter = PrintServiceMatcher.matchPrinter(ps.issuingPrinterName);
```
tresf, line 185: Please comment this well.

src/qz/printer/status/StatusSession.java
```diff
@@ -21,7 +21,7 @@ public void statusChanged (PrinterStatus printerStatus) {
 
     private StreamEvent createStatusStream(PrinterStatus status) {
         return new StreamEvent(StreamEvent.Stream.PRINTER, StreamEvent.Type.ACTION)
-                .withData("printerName", status.issuingPrinterName)
```
tresf, line 24: Please keep `issuingPrinterName` and add `issuingPrinterDescription` as an additional field.

src/qz/printer/status/CupsStatusHandler.java
Vzor-: Wouldn't it have to be `StringUtils.substringBeforeLast(StringEscapeUtils.unescapeXml(StringUtils.substringAfter(printerName, "\"")), "\"");`?

src/qz/printer/status/PrinterStatus.java
Vzor-: The reason this changed is because this is the status' toString, and I wanted it to reflect exactly what the status will report to the client. Should I still go ahead with this change?

src/qz/printer/status/CupsStatusHandler.java
tresf: Yeah, that's ufly.  Perhaps `substringBetween` instead?

Vzor-: I think the matches from left to right, we need to do an outer match.

tresf: > I think the matches from left to right, we need to do an outer match.

You're right, and MacOS allows double quotes.  We should leave it as-is.

PR #707, opened 2020-09-08
Initial support for --help option

tresf:
Preview:

```
Usage: java -jar qz-tray.jar (command)

INFORMATION
  --help, -h                  Display help information and exit.
  --version, -v               Display version information and exit.
  --bundleid, -i              Display Apple bundle identifier and exit.
  --libinfo, -l               Display detailed library version information and exit.

ACTION
  --allow, --whitelist, -a    Add the specified certificate to allowed.dat.
                                java -jar qz-tray.jar --allow cert.pem
  --block, --blacklist, -b    Add the specified certificate to blocked.dat.
                                java -jar qz-tray.jar --block cert.pem

OPTION
  --honorautostart, -A        Read and honor any autostart preferences before launching.
  --headless                  Force startup "headless" without graphical interface or interactive components.

INSTALLER
  preinstall                  Perform critical pre-installation steps: Stop instances, all other special considerations.
  install                     Copy to the specified destination and preforms platform-specific registration.
                                java -jar qz-tray.jar install --dest /my/install/location [--silent]
  certgen                     Performs certificate generation and registration for proper HTTPS support.
                                java -jar qz-tray.jar certgen [--key key.pem --cert cert.pem] [--pfx cert.pfx --pass 12345] [--host "list;of;hosts"]
  uninstall                   Perform all uninstall tasks: Stop instances, delete files, unregister settings.
  spawn                       Spawn an instance of the specified program as the logged-in user, avoiding starting as the root user if possible.
                                java -jar qz-tray.jar spawn [program params ...]

For help on a specific command:
  Usage: java -jar qz-tray.jar --help (command)
    --help install
    --help certgen

Process finished with exit code 2
```

### DIFF
--- a/src/qz/utils/ArgValue.java
+++ b/src/qz/utils/ArgValue.java
@@ -1,0 +1,85 @@
+package qz.utils;
+
+import qz.common.Constants;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static qz.utils.ArgValue.ArgType.*;
+
+public enum ArgValue {
+    // Informational
+    HELP(INFORMATION, "--help", "-h", "Display help information and exit.", null),
+    VERSION(INFORMATION, "--version", "-v", "Display version information and exit.", null),
+    BUNDLEID(INFORMATION, "--bundleid", "-i", "Display Apple bundle identifier and exit.", null),
+    LIBINFO(INFORMATION, "--libinfo", "-l", "Display detailed library version information and exit.", null),
+
+    // Actions
+    ALLOW(ACTION,"--allow", "--whitelist", "-a", String.format("Add the specified certificate to %s.dat.", Constants.ALLOW_FILE), "--allow cert.pem"),
+    BLOCK(ACTION,"--block", "--blacklist", "-b", String.format("Add the specified certificate to %s.dat.", Constants.BLOCK_FILE), "--block cert.pem"),
+
+    // Options
+    AUTOSTART(OPTION,"--honorautostart", "-A", "Read and honor any autostart preferences before launching.", null),
+
+    // Installer stubs
+    PREINSTALL(INSTALLER, "preinstall", "Perform critical pre-installation steps: Stop instances, special considerations.", null),
+    INSTALL(INSTALLER, "install", "Copy to the specified destination and preforms platform-specific registration.", "install --dest /my/install/location [--silent]"),
+    CERTGEN(INSTALLER, "certgen", "Performs certificate generation and registration for proper HTTPS support.", "certgen [--key key.pem --cert cert.pem] [--pfx cert.pfx --pass 12345] [--host \"list;of;hosts\"]"),
+    UNINSTALL(INSTALLER, "uninstall", "Perform all uninstall tasks: Stop instances, delete files, unregister settings.", null),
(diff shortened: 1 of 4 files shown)


tresf:
Added command-specific help support via 0c8a440.

```
For help on a specific command:
  Usage: java -jar qz-tray.jar --help (command)
    --help install
    --help certgen
```

e.g.:

```bash
java -jar /opt/qz-tray/qz-tray.jar --help certgen
```

```
Usage: java -jar qz-tray.jar (command)
  certgen                     Performs certificate generation and registration for proper HTTPS support.
                                java -jar qz-tray.jar certgen [--key key.pem --cert cert.pem] [--pfx cert.pfx --pass 12345] [--host "list;of;hosts"]
OPTIONS
  --host, --hosts             Semicolon-delimited hostnames and/or IP addresses to generate the HTTPS certificate for.
  --cert, -c                  Path to a stand-alone HTTPS certificate
  --key, -k                   Path to a stand-alone HTTPS private key
  --pfx, --pkcs12             Path to a paired HTTPS private key and certificate in PKCS#12 format.
  --pass, -p                  Password for decoding private key.
```

Updated the output preview above with the latest.

This PR is ready for QA, and code review.

src/qz/utils/ArgValue.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+    ArgValue(ArgType argType, String match1, String match2, String match3, String description, String usage) {
+        this(argType, new String[] { match1, match2, match3 }, description, usage);
+    }
+    ArgValue(ArgType argType, String match1, String match2, String description, String usage) {
+        this(argType, new String[] { match1, match2 }, description, usage);
+    }
+    ArgValue(ArgType argType, String match, String description, String usage) {
+        this(argType, new String[] { match }, description, usage);
+    }
```
bberenz, line 51: Why not `ArgValue(ArgType argType, String description, String usage, String... matches) {` and avoid multiple constructors like this?

tresf: Because the compiler can't handle the difference between indeterminate strings, I'd have to inject a non-string in the middle. 😕

bberenz: I'm not sure I follow - are you saying it can't have a vararg as the last argument if the preceding argument is the same type?

tresf: Exactly.

tresf: It allows it, but it picks the wrong function declaration unless an array is forced as the last parameter, negating it's usefulness.

tresf: A bit of clarification... the reason matches isn't the last parameter is mostly for readability purposes, since the `--foo` is the most important part of this class, I want it easy to find and read.  The parameter ambiguity can be circumvented using dedicated `null` placeholders in the constructors.

tresf: Done via 8d242b7.

tresf:
This is testing well.  The one change I'd like -- which is proving a bit too difficult for Java -- is to make the help text resize based on console size.  This is difficult because there's no API for Java to query this information resulting in a lot of hacks -- none of which are compatible with the IDE I'd be authoring these with,so merging as-is.

Issue #704, opened 2020-09-03
Nearest-neighbor on macOS 2.1


klabarge:
`nearest-neighbor` is not honored on the 2.1 branch on macOS. Windows and Linux work as expected

**OS:** macOS Catalina 10.15.6
**Java:** AOJDK 11.0.4

<img width="947" alt="image" src="https://user-images.githubusercontent.com/12505463/92065859-7991b880-ed6e-11ea-9bfb-a497de0b2141.png">

PR #711, opened 2020-09-11
Re-add missing dpiscale setter

bberenz:
Fix to scaling interpolation for #704

Looks like a mistake/lost line from a rebase? As the history timeline around [this commit](https://github.com/qzind/tray/commit/25a19840d5237bd10bd4f856fb2c22a9f5534012) is weird.
I didn't see any ill effects to rasterized HTML re-adding this line in, though it would probably be a good idea to test around that as well, since the linked commit was for adding that capability in.

### DIFF
--- a/src/qz/printer/action/PrintImage.java
+++ b/src/qz/printer/action/PrintImage.java
@@ -69,6 +69,8 @@ public class PrintImage extends PrintPixel implements PrintProcessor, Printable 
 
     @Override
     public void parseData(JSONArray printData, PrintOptions options) throws JSONException, UnsupportedOperationException {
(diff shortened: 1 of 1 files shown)


tresf:
> I didn't see any ill effects to rasterized HTML re-adding this line in, though it would probably be a good idea to test around that as well, since the linked commit was for adding that capability in.

I think because `parseData`, the function where the fix lands is overridden, no?

https://github.com/qzind/tray/blob/c96a7dd6e1a1574e4d042addf4a2316a49cd31e9/src/qz/printer/action/PrintHTML.java#L64-L66

PR #724, opened 2020-10-19
Add docker support

klarkc:
Adds docker build file

Build:

```bash
docker build -t tray .
```

Usage:

```bash
docker run -p 8182:8182 -p 8181:8181 -it tray
```

To use a certificate in `cert.pem` file:

```bash
docker run --name tray -v $(pwd)/cert.pem:/tmp/cert.pem -p 8182:8182 -p 8181:8181 tray
docker exec tray /opt/qz-tray/qz-tray --whitelist /tmp/cert.pem
```


tresf:
@klarkc thanks for the contribution!

PR #728, opened 2020-11-04
Allow spool limits for pdf prints

### LINKED ISSUE #727: Pixel "perSpool" option
Currently raw printing allows a "perSpool" option to split large batches into smaller, more reasonably sized batches.

Certain Pixel jobs may benefit from this option as well.  For example, a client is printing > 1,000 pages, but they must wait for the entire job to spool before printing starts.

The proposal is to spool documents in smaller batches (e.g. 10, 100, 500) so that the spooler can begin printing while the next jobs begins the spooling process.

bberenz:
**Reference** 

| Before | After |
|-------|-------|
| `qz.configs.create('Zebra',` **` { perSpool: 10 }`** | `qz.configs.create('Zebra',` **`{ spool: { size: 10 } } )`** |

**Raw only**
| Before | After |
|-------|-------|
| `qz.configs.create('Zebra',` **` { endOfDoc: '^XZ'  }`** | `qz.configs.create('Zebra',` **`{ spool: { end: '^XZ' } } )`** |

Compatibility:
* 2.1.3 and higher only.  QZ Tray will continue to honor the old, deprecated API for [raw only](https://github.com/qzind/tray/pull/728#issuecomment-724852777).
* qz-tray.js will automatically convert the new API to the old format if an older QZ Tray version is detected.

---

perSpool option added for pdf pixel prints, closes #727


js/qz-tray.js
```diff
@@ -1306,7 +1306,7 @@ var qz = (function() {
              *  @param {boolean} [options.altPrinting=false] Print the specified file using CUPS command line arguments.  Has no effect on Windows.
              *  @param {string} [options.encoding=null] Character set
              *  @param {string} [options.endOfDoc=null]
-             *  @param {number} [options.perSpool=1] Number of pages per spool.
+             *  @param {number} [options.perSpool=null] Number of pages per spool. Defaults to <code>1</code> for raw print jobs.
```
tresf, line 1309: @bberenz should we keep this null and document the `1` in the `endOfDoc` instead?  The `1` is ignored without it.

tresf:
@bberenz I believe we need to add some form of backwards compatibility routine or else QZ Tray 2.1.3+ will take the old `perSpool: 1` from older qz-tray.js versions, and run with it.

tresf:
> @bberenz I believe we need to add some form of backwards compatibility routine or else QZ Tray 2.1.3+ will take the old `perSpool: 1` from older qz-tray.js versions, and run with it, resulting in e.g. 1,000 separately spooled documents for instances like the original bug report.

Perhaps a sane way to provide this is to check for `endOfDoc` on the pixel side, since it was previously ONLY required for raw, ignored otherwise.  Then, if `endOfDoc` is there, assume `perSpool` wasn't really intended over on the pixel-side.

bberenz:
Should we just put the pixel option under a different property?
Something like "splitSpool:<int>"?

tresf:
> Should we just put the pixel option under a different property?

I quite prefer keeping raw and pixel using similar APIs.  I was trying to find an API indicator suggesting that `1` was a relic safe to ignore, but I couldn't find one.

My proposal to yours is similar but uses an object.  I'm not sure if this is overkill, but would make home to more spool control parameters in the future.  For example, if we find a way to control winspool's behavior such as "keep spooled documents", or "send directly to the printer", etc although these may be better suited in a non-`print(...)` API since they're sweeping.  Other future possibilities include page-handling, such as sending page 7 to a different tray.  These types of advanced printing capabilities are missing from the PDF spec and may have value, thus a place in the API namespace.  Thoughts welcome.

bberenz:
That'd probably be the way to go besides just blanket ignoring a value of 1 for pixel spools then.
I'd probably go with something besides `config.spool.per`, maybe `spool.sendPer` or `spool.size`?

tresf:
Implemented object notation via 4f1b58b (untested).  Feedback welcome.

tresf:
:white_check_mark: **PASS:** QZ Tray `2.1.3-RC1`
:white_check_mark: **PASS:** QZ Tray `2.1.2`
:white_check_mark: **PASS:** Using deprecated syntax `{ perSpool: 1, endOfDoc: '^XZ' }`

PR #732, opened 2020-11-19
Update JNA

bberenz:
+ libraries dependent on


lite1979:
On MacOS (Catalina for now, I'll test on Big Sur after I update later today) and Ubuntu 20.04, I get the error:

Error: HID device could not be found

When trying to "Claim Device" from sample.html

I got the same error the first time I tried to connect in Windows 10, but I haven't gotten it since, and HID streaming weight from the scale has worked as expected.

lite1979:
HID communication on MacOS Big Sur, Windows 10 (recent update), and Ubuntu 20.04 is now working as expected with the Mettler Toledo PS60. :)

Issue #739, opened 2020-12-11
 TypeError: null is not an object (evaluating 'document.body.offsetHeight')


tresf:
@lite1979 observed while testing #731 that volvo.com's website throws the following error and prints a blank page.  Documenting here, so we can identify the cause and attempt to fix it.  Note, `document.body.offsetHeight` is code that's specific to QZ Tray.

```
	Done: 30.000000000000004 > 100.0
[TRACE] 2020-12-11 17:26:40,389 @ qz.printer.action.WebApp:?
	New state: RUNNING > SUCCEEDED
Exception in thread "JavaFX Application Thread" netscape.javascript.JSException: TypeError: null is not an object (evaluating 'document.body.offsetHeight')
	at com.sun.webkit.dom.JSObject.fwkMakeException(JSObject.java:157)
	at com.sun.webkit.WebPage.twkExecuteScript(Native Method)
	at com.sun.webkit.WebPage.executeScript(WebPage.java:1514)
	at javafx.scene.web.WebEngine.executeScript(WebEngine.java:978)
	at qz.printer.action.WebApp.findHeight(Unknown Source)
	at qz.printer.action.WebApp.lambda$static$1(Unknown Source)
	at com.sun.javafx.binding.ExpressionHelper$SingleChange.fireValueChangedEvent(ExpressionHelper.java:181)
	at com.sun.javafx.binding.ExpressionHelper.fireValueChangedEvent(ExpressionHelper.java:80)
	at javafx.beans.property.ReadOnlyObjectPropertyBase.fireValueChangedEvent(ReadOnlyObjectPropertyBase.java:74)
	at javafx.beans.property.ReadOnlyObjectWrapper.fireValueChangedEvent(ReadOnlyObjectWrapper.java:102)
	at javafx.beans.property.ObjectPropertyBase.markInvalid(ObjectPropertyBase.java:113)
	at javafx.beans.property.ObjectPropertyBase.set(ObjectPropertyBase.java:147)
	at javafx.scene.web.WebEngine$LoadWorker.updateState(WebEngine.java:1257)
	at javafx.scene.web.WebEngine$LoadWorker.dispatchLoadEvent(WebEngine.java:1372)
	at javafx.scene.web.WebEngine$PageLoadListener.dispatchLoadEvent(WebEngine.java:1237)
	at com.sun.webkit.WebPage.fireLoadEvent(WebPage.java:2514)
[...]

PR #743, opened 2020-12-22
Attempt to detect redirected pages

bberenz:
Potential fix for #739 
Issue is caused by page requested being redirected so there is no document body. Error will throw for the js call and then the blank page printed while the actual content gets loaded and thrown away in the background.

The JFX webengine however does not seem to have a listener for http status codes. So this solution uses a null check on the js document.body to assume a redirect is occurring. I'm not sure if this will catch all forms of redirects or if there are instances where body could be null without a redirect, causing the print to be endlessly stuck.


tresf:
@lite1979, requesting your review.

PR #747, opened 2021-01-21
Job Status Support for CUPS

Vzor-:
Ready for review.

~~Currently have an issue with redundant job statuses. Should we track the last status for every job?~~

Windows + Mac Tested

Still needs testing on ubuntu


src/qz/printer/status/printer/CupsPrinterStatusMap.java
```diff
@@ -44,11 +46,811 @@
...
+    WRAPPER_TURNED_OFF(NativePrinterStatus.UNKNOWN_STATUS), // wrapper-turned-off
+    WRAPPER_TURNED_ON(NativePrinterStatus.UNKNOWN_STATUS), // wrapper-turned-on
+    WRAPPER_UNDER_TEMPERATURE(NativePrinterStatus.UNKNOWN_STATUS), // wrapper-under-temperature
+    WRAPPER_UNRECOVERABLE_FAILURE(NativePrinterStatus.UNKNOWN_STATUS), // wrapper-unrecoverable-failure
+    WRAPPER_UNRECOVERABLE_STORAGE_ERROR(NativePrinterStatus.UNKNOWN_STATUS), // wrapper-unrecoverable-storage-error
+    WRAPPER_WARMING_UP(NativePrinterStatus.UNKNOWN_STATUS), // wrapper-warming-up
+
+    //printer-state
+
+    IDLE(NativePrinterStatus.OK), // idle
```
tresf, line 851: Does this belong higher? e.g. "CUPS defined states (defined by CUPS, but not part of the IPP specification" ?

src/qz/printer/status/printer/CupsPrinterStatusMap.java
```diff
@@ -57,17 +859,18 @@
         this.parent = parent;
     }
 
-    public static NativePrinterStatus match(String code) {
+    public static NativeStatus match(String code, String state) {
```
tresf, line 862: What's the purpose of changing the return type to the parent type?

src/qz/printer/status/NativeStatus.java
```diff
@@ -95,13 +80,17 @@
         return fromWmi(bitwiseCode, printer, nativeType, -1, null);
     }
 
-    static Status fromCups(String reason, String printer, NativeType nativeType) {
-        if (reason == null) { return null; }
 
+    static Status fromCups(String reason, String state, String printer, int jobId, NativeType nativeType) {
```
tresf, line 84: There's no point in passing in `nativeType` if you're going to assume Job/Printer based on function signature.  I also feel this diverges it from Wmi, we should probably settle on something that makes them similar.

src/qz/printer/status/NativeStatus.java
```diff
@@ -95,13 +80,17 @@
...
+        return new Status(CupsJobStatusMap.match(reason, state), printer, reason, jobId, printer + jobId);
```
tresf, line 85: Is `printer + jobId` supposed to be the job name?  It's not.

src/qz/printer/status/CupsUtils.java
```diff
@@ -150,6 +196,28 @@ static void startSubscription(int rssPort) {
         cups.ippDelete(response);
     }
 
+    static void startJobSubscription(int rssPort) {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> freeIppObjs()));
```
tresf, line 200: Is this a secondary thread?  Is it needed?

src/qz/printer/status/NativeStatus.java
```diff
@@ -10,6 +12,8 @@
 import java.util.Locale;
 
 public interface NativeStatus {
+    Logger log = LoggerFactory.getLogger(NativeStatus.class);
```
tresf, line 15: Is this needed here?  Can it be a static variable?

src/qz/printer/status/Cups.java
```diff
@@ -22,11 +19,14 @@
         public static int TAG_NAME = INSTANCE.ippTagValue("Name");
         public static int TAG_INTEGER = INSTANCE.ippTagValue("Integer");
         public static int TAG_KEYWORD = INSTANCE.ippTagValue("keyword");
+        public static int TAG_ENUM = INSTANCE.ippTagValue("enum");
```
aimiebell, line 22: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Use of identifier that is a keyword in later versions of Java (NM_FUTURE_KEYWORD_USED_AS_IDENTIFIER)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-use-of-identifier-that-is-a-keyword-in-later-versions-of-java-nm-future-keyword-used-as-identifier).
The identifier is a word that is reserved as a keyword in later versions of Java, and your code will need to be changed in order to compile it in later versions of Java.

src/qz/printer/status/printer/CupsPrinterStatusMap.java
Vzor-: It brings it into line with the other match functions

tresf:
@Vzor- this looks to be in great shape.  From what I understand, CUPS is firing some duplicate events.  I think once that's sorted, we can merge.

tresf:
So it turns out the timing of things is causing hurdles.  Specifically, the JNA calls are too late and don't have the status at the exact time of the RSS notification.  We'll need to rework this.

PR #766, opened 2021-02-26
Raw images to honor orientation 

### LINKED ISSUE #762: Add orientation flag to rawHTML printing
Users want to be able to manipulate the raster graphic when sending rawHTML data to printers using ZPL. 

This is currently not possible with QZ-Tray.

This feature could also enhance abilities for EPL and ESC/POS if we can make it work with ZPL.

My first thought is to study the ZPL manual for an option to adjust orientation for the raster image.

tresf:
Adds the ability for a raw job to handle rotate or orientation normally reserved for pixel jobs.

This will cause backwards compat issues for customer unknowingly passing orientation to a raw job.

Closes #762

### DIFF
--- a/src/qz/printer/action/PrintRaw.java
+++ b/src/qz/printer/action/PrintRaw.java
@@ -243,7 +243,7 @@ public class PrintRaw implements PrintProcessor {
 
     private ImageWrapper getWrapper(BufferedImage img, JSONObject opt, PrintOptions.Pixel pxlOpts) {
         // Rotate image using orientation or rotation before sending to ImageWrapper
-        if(pxlOpts.getOrientation() != PrintOptions.Orientation.PORTRAIT) {
+        if(pxlOpts.getOrientation() != null && pxlOpts.getOrientation() != PrintOptions.Orientation.PORTRAIT) {
             img = PrintImage.rotate(img, pxlOpts.getOrientation().getDegreesRot(), pxlOpts.getDithering(), pxlOpts.getInterpolation());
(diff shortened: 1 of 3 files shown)


src/qz/printer/action/PrintRaw.java
```diff
@@ -146,9 +148,10 @@ public void parseData(JSONArray printData, PrintOptions options) throws JSONExce
         }
     }
 
-    private ImageWrapper getImageWrapper(String data, JSONObject opt, boolean fromFile) throws IOException {
+    private ImageWrapper getImageWrapper(String data, JSONObject opt, PrintingUtilities.Flavor flavor, PrintOptions.Pixel pxlOpts) throws IOException {
         BufferedImage bi;
 
+        boolean fromFile = flavor != PrintingUtilities.Flavor.BASE64;
```
bberenz, line 154: Why does this still use a fromFile variable instead of directly checking the flavor in the if-statement like getpdfwrapper now does?

tresf: Because of the legacy variable which is only used for images.  Do you have a better recommendation?

bberenz: explicitly set the flavor to base64 in the compat if-block?

tresf: Done.

tresf:
All comments addressed.

bberenz (approved):

tresf:
Merging into `html-base64` so we have a centralized branch to complete our testing.

PR #777, opened 2021-03-29
removed print rate and media darkness commands from ZPL sample

tresf:


lite1979:
Interesting. I only used VS code and terminal to change zpl_sample.txt, but I also noticed that three files were changed. Does IntelliJ's presence on my system automatically do this?

tresf:
> Does IntelliJ's presence on my system automatically do this?

Unfortunately, yes.  If the project is opened in IntelliJ once it will start touching these files.  It's something @Vzor- and I have tried and failed at changing the default behavior of.  We usually commit using IntelliJ directly, and then we uncheck these files.

PR #781, opened 2021-04-01
Allow re-encoding on non-String parameters

### LINKED ISSUE #774: Support encoding for non-String formats
Currently, encoding conversions [only apply to `PLAIN` formats](https://github.com/qzind/tray/blob/c2bf27b13c86c675efd6da811c2987f843888109/src/qz/printer/action/PrintRaw.java#L139).

In some occasions, data may be provided in `BASE64`, `FILE` (or even `HEX`) in the wrong encoding and a programmer may wish to leverage Java to do encoding conversion.

One potential downside of this is, it assumes quite a few things:
* It assumes the data being sent can easily be converted to String
* It assumes the data being sent can be guessed as UTF8<br>---- OR ----
* It would require informing Java about the source format as to make no assumptions

I'm thinking we can use something like:

```js
{ encoding: 'Cp1252', sourceEncoding: 'UTF8' }
``` 

Or perhaps... 
```js
{ encoding: 'UTF8:Cp1252' }
```

... although this may be undesired, since a colon is a valid delimiter in some encodings [aliases](https://docs.oracle.com/javase/8/docs/technotes/guides/intl/encoding.doc.html) like `ISO_8859-1:1987`.

Or perhaps... 
```js
{ encoding: { from: 'UTF8', to: 'Cp1252' } }
```

We should also add a warning for all raw formats that have provided `encoding` and it's being ignored.  It can then hint to supply `sourceEncoding`|`from:` to remedy the issue.

tresf:
Closes #774

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -1343,7 +1343,9 @@ var qz = (function() {
              *  @param {string} [options.units='in'] Page units, applies to paper size, margins, and density. Valid value <code>[in | cm | mm]</code>
              *
              *  @param {boolean} [options.altPrinting=false] Print the specified file using CUPS command line arguments.  Has no effect on Windows.
-             *  @param {string} [options.encoding=null] Character set
+             *  @param {string|Object} [options.encoding=null] Character set for commands. Can be provided as an object for converting encoding types for RAW types.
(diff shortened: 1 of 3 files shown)


tresf:
Some notes before merging... 

* The base64 converter I was using defaults to `ISO-8859-1`, and it seems to have been silently converting my data to `Cp1252`.  Switching it to UTF-8 reproduced the originating issue.
* The commands will generally still need whatever proprietary commands are required for encoding selection.

PR #794, opened 2021-04-30
Fix blank HTML printing on sample.html

klabarge:
When size is checked on the `sample.html` page, but the height and width fields are `null`, HTML prints are blank.

This adds an additional check and only updates the `width` and `height` if there's data in the respective fields.

image

### DIFF
--- a/sample.html
+++ b/sample.html
@@ -2926,7 +2926,7 @@
 
     function updateConfig(cleanConditions) {
         var pxlSize = null;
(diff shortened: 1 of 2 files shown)


bberenz:
Would probably be helpful to users outside the demo page if we could add a warning about this when we parse the size values on the [java side](https://github.com/qzind/tray/blob/master/src/qz/printer/PrintOptions.java#L328) too.

PR #797, opened 2021-05-04
Idle detector

### LINKED ISSUE #791: Improve performance of first print
Often QZ Tray is started but it's not actually used until seconds or minutes later.

This offers an opportunity to fix some perceived performance issues... Specifically... 
* [ ] Slow to find printers the first time if a lot of printers are installed
* [ ] Slow to start JavaFX the first time
* Ability to easily add any other slow-to-start tasks

If QZ Tray has an idle detector, we could run these as background tasks so that QZ Tray is perceivably faster on first use.

bberenz:
Add idle detector to perform some 'slow the first time' actions as ~~described in~~ Fixes #791

### DIFF
--- a/src/qz/common/TrayManager.java
+++ b/src/qz/common/TrayManager.java
@@ -204,7 +204,7 @@ public class TrayManager {
         // Slow to find printers the first time if a lot of printers are installed
         performIfIdle((int)TimeUnit.SECONDS.toMillis(10), evt -> {
             log.debug("IDLE: Performing first run of find printers");
-            PrintServiceMatcher.getNativePrinterList(true);
+            PrintServiceMatcher.getNativePrinterList(false, true);
         });
         // Slow to start JavaFX the first time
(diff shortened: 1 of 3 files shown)


src/qz/common/TrayManager.java
```diff
@@ -182,14 +188,34 @@ public TrayManager(boolean isHeadless) {
...
             addMenuItems();
         }
+
+        // Initialize idle actions
+        idleTimers = new ArrayList<>();
+
+        // Slow to find printers the first time if a lot of printers are installed
+        performIfIdle((int)TimeUnit.SECONDS.toMillis(10), evt -> {
+            log.debug("IDLE: Performing first run of find printers");
+            PrintServiceMatcher.getNativePrinterList(true);
```
tresf, line 207: This is only half the battle... what takes a bit is the registry/shell stuff that gets called per-printer.

tresf: Added proactive fetching of all the driver details per 3f133fa

Issue #793, opened 2021-04-30
Exception in thread "JavaFX Application Thread"


klabarge:
I get this error when sending an invalid html print config (for example, a 4x6in label with 2in margins)

`size:{width: 4, height: 6}`
`margins: 2`

The print error is probably sane, however QZ Tray never recovers. I am unable to print using QZ Tray until I force close QZ Tray.

**OS:** macOS 11.2.2 (20D80)
**Java:** AdoptOpenJDK (build 11.0.10+9)

```
Exception in thread "JavaFX Application Thread" java.lang.IllegalArgumentException: 0 or negative value argument
	at java.desktop/javax.print.attribute.standard.MediaPrintableArea.<init>(MediaPrintableArea.java:127)
	at com.sun.prism.j2d.print.J2DPrinterJob.syncPageLayout(J2DPrinterJob.java:691)
	at com.sun.prism.j2d.print.J2DPrinterJob.syncSettingsToAttributes(J2DPrinterJob.java:596)
	at com.sun.prism.j2d.print.J2DPrinterJob.print(J2DPrinterJob.java:822)
	at javafx.print.PrinterJob.printPage(PrinterJob.java:399)
	at javafx.print.PrinterJob.printPage(PrinterJob.java:415)
	at qz.printer.action.WebApp.lambda$print$5(Unknown Source)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
```

Here you can see 8 minutes go by from sending a print job and having to close QZ Tray.

```
[DEBUG] 2021-04-29 19:46:08,073 @ qz.utils.PrintingUtilities:?
	Using qz.printer.action.PrintHTML to print
[DEBUG] 2021-04-29 19:46:08,074 @ qz.printer.PrintServiceMatcher:?
	Searching for PrintService matching PDFwriter
[DEBUG] 2021-04-29 19:46:08,199 @ qz.printer.PrintServiceMatcher:?
	Found 6 printers
[DEBUG] 2021-04-29 19:46:08,300 @ qz.printer.PrintServiceMatcher:?
	Found match: PDFwriter
[...]

PR #798, opened 2021-05-04
Catch and unlatch html if it fails during print call

bberenz:
For #793


tresf:
@klabarge can you test?  You likely need to use `git revert 561bdea` to make sure the `<= 0` guards are disabled.

klabarge (approved):
I tested this and I am able to print jobs after getting the `JavaFX Application Thread` error 👍

>~You likely need to use git revert 561bdea to make sure the <= 0 guards are disabled.~
Edit: No, you won't, this isn't rebased on master yet.

I used the test case of sending an HTML print job with `margins: 2` to a `4x6in` paper size, so I don't think the `<=0` guards come into play.

Issue #468, opened 2019-07-18
Bundled Java Runtime


tresf:
Placeholder to track the progress of bundling the Java Runtime into the installers.

* [x] Download and extract the `target` JDK:<br>**Note:** `ant` makes this tricky.  See [`download-javafx`](https://github.com/qzind/tray/blob/2.1/ant/javafx.xml#L2) as an example.
   * `ant nsis` - Download Windows 64-bit JDK
   * `ant pkgbuild` - Download MacOS 64-bit JDK
   * `ant makeself` - Download Linux 64-bit JDK
* [x] Locate required JRE components using `jlink`, bundle into `<INSTALL_DIR>\jre` folder.
* [x] Ensure all installers use same JRE detection technique (see the NSIS [`FindJRE` function](https://github.com/qzind/tray/blob/2.1/ant/windows/windows-launcher.nsi.in#L55)) (done via #504)
* [x] Installer to use the bundled `java`, `javaw`, `keytool` when available (covered elsewhere)
* [ ] Work through deployment nuances
* [ ] Pass macOS notarization

tresf:
Initial proof of concept available https://github.com/qzind/tray/tree/jlink

tresf:
A small caveat... if building using a newer Java runtime, `jlink` will fail.  More details here: https://bugs.openjdk.java.net/browse/JDK-8204256

The workaround is to make sure your `jlink` version **is the same major version** between that being downloaded (e.g. AdoptOpenJDK11) and that being used to build (e.g. Homebrew's `openjdk@11`).  So on a system which defaults to openjdk14, you have to roll it back to openjdk11:

```bash
brew uninstall --ignore-dependencies openjdk
brew install openjdk@11
ln -s /usr/local/opt/openjdk@11 /usr/local/opt/openjdk
```

tresf:
The bundled Java Runtime is now functional on  the`jlink` branch for Windows and macOS.  The installers succeed without asking to Download Java.

I'm still working through some serious startup issues on MacOS, I assume to be permissions related...

```
java.io.IOException: Cannot run program "defaults": error=316, posix_spawn failed
java.io.IOException: Cannot run program "pgrep": error=2, No such file or directory
java.io.IOException: Cannot run program "whoami": error=0, posix_spawn failed
java.io.IOException: Cannot run program "security": error=2, No such file or directory
java.io.IOException: Cannot run program "system_profiler": error=2, No such file or directory
[...]

PR #823, opened 2021-06-10
Add jdk16 support

### LINKED ISSUE #822: MacOS Big Sur cannot open QZ Tray if OpenJDK16 is installed
Hi,

I have to have OpenJDK16 installed, because it's installed as a dependency for a few other things.

I also have AdaptOpenJDK11 installed, setting the JAVA_HOME environment variable doesn't seem to work, QZ Tray ignores it.

When i run ./QZ\ Tray from the app bundle, it runs the command `java_home -v 1.8+` and the plus appears to just seek out the latest version of Java installed, maybe there is a way to skip the java_home command from running, but i cannot figure it out?

I've checked a number of other issues and can't seem to find anyone with the same problem.

Is there a way i can get this running, without removing OpenJDK16?

Thanks

tresf:
1. Fix tray icon (https://github.com/dyorgio/macos-tray-icon-fixer/issues/9)
2. Fix dark mode (TBD: https://github.com/bobbylight/Darcula/issues/8)

Closes #822

### DIFF
--- a/ant/project.properties
+++ b/ant/project.properties
@@ -9,6 +9,7 @@ project.datadir=qz
 
 install.opts=-Djna.nosys=true
 launch.opts=-Xms512m ${install.opts}
+launch.jigsaw=--add-opens java.desktop/sun.lwawt.macosx=ALL-UNNAMED --add-opens java.desktop/java.awt=ALL-UNNAMED
 launch.overrides=QZ_OPTS
 
 src.dir=${basedir}/src
--- a/ant/unix/unix-launcher.sh.in
+++ b/ant/unix/unix-launcher.sh.in
@@ -87,6 +87,14 @@ else
     echo -e "$SUCCESS Java $curver was detected"
 fi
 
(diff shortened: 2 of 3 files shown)


ant/project.properties
```diff
@@ -9,6 +9,7 @@ project.datadir=qz
 
 install.opts=-Djna.nosys=true
 launch.opts=-Xms512m ${install.opts}
+launch.jigsaw=--add-opens java.desktop/sun.lwawt.macosx=ALL-UNNAMED --add-opens java.desktop/java.awt=ALL-UNNAMED
```
dyorgio, line 12: Maybe you can use AddOpens MANIFEST.MF entries or module-info.java, not sure 🤓

tresf: I would much prefere this however I believe forcing `module-info.java` will break compilation with Java < 8 and require a special ant target.  We will probably do this in the near future when we finish bundling the runtime #468.

PR #828, opened 2021-06-22
Add qz-tray-console.exe for debugging purposes

tresf:
image

Supersedes #801


tresf:
@klabarge installer available [here](https://github.com/tresf/tray/releases/download/TEMP/qz-tray-2.1.3+win-console.exe).

Note, this was cross-compiled using Ubuntu 20.04 ARM64 edition and the new build flag `os.arch` via:

```bash
ant nsis -Dos.arch=x86
```

klabarge:
When I call the console from `cmd` (`"C:\Program Files\QZ Tray\qz-tray-console.exe"`), it launches a new window.

This seems to work as expected. I tried printing some HTML with `rasterize: false` to test the memory shelf and I didn't run into any memory issues.

Is there anything specific I should test before approving?


image

tresf:
> When I call the console from `cmd` (`"C:\Program Files\QZ Tray\qz-tray-console.exe"`), it launches a new window.

Good point.  Yeah, I hadn't thought about that.  If you look at the commit, it's just a small trick that replaces `javaw.exe` with `java.exe` and re-builds the launcher.

The memory shelf should be identical to the normal launcher, because it's a copy of it (clones it at compile time, replaces the `javaw.exe` with `java.exe`).

The new console can be avoided using some NSIS tricks but it's a hack, so I opted for the simplest solution, which was to just leverage the console version of Java.  The reason this happens is because in [Windows, you can't have a GUI app and a console app in the same exe](https://stackoverflow.com/a/574918/3196753) and NSIS is a GUI app only [unless the NSIS source is recompiled](](https://nsis-dev.github.io/NSIS-Forums/html/t-227000.html). 

The hacks include attaching to a console and writing output, but it would be harder to re-use code and may be less reliable.

> Is there anything specific I should test before approving?

I can't think of any.

klabarge:
The only problem I see with this behavior is the console closes automatically when the app is no longer running. So if there's a hard crash to QZ Tray, the stack trace containing the error will be lost.

If there's a hard-crash we're trying to debug on Windows, we'll have to use the old method for launching QZ Tray via the console.

tresf:
> If there's a hard-crash we're trying to debug on Windows, we'll have to use the old method for launching QZ Tray via the console.

Right, and the old method is growing increasingly difficult.  I'll see what I can whip up.

tresf:
@klabarge I've introduced a new technique in 00ff4bd which launches an instance of `cmd.exe` that will remain open even after a crash.  

Installer available [here](https://github.com/tresf/tray/releases/download/TEMP/qz-tray-2.1.3+win-console2.exe).

klabarge:
>which launches an instance of cmd.exe that will remain open even after a crash.

I tested this and it works 👍. Approving.

klabarge (approved):

Issue #835, opened 2021-07-12
Don't suppress setCertificatePromise rejection


tresf:
When a certificate promise rejects, it's hard to know... 

https://github.com/qzind/tray/blob/f39e9d48d8a3e43348b29f108b548db85b92b7fe/js/qz-tray.js#L376

I believe we should `warn` or `allay` when this occurs.

tresf:
Closed via #836

PR #836, opened 2021-07-15
Unsuppress promise rejection

bberenz:
Fix for #835, now prints warning when retrieving certificate

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -373,7 +373,12 @@ var qz = (function() {
                         });
                     }
 
-                    _qz.security.callCert().then(sendCert).catch(sendCert);
+                    function catchAndSendNull(error) {
+                        _qz.log.warn("Failed to get certificate:", error);
+                        sendCert(null);
+                    }
+
+                    _qz.security.callCert().then(sendCert).catch(catchAndSendNull);
                 },
 
                 /** Generate unique ID used to map a response to a call. */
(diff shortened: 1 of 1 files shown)


js/qz-tray.js
```diff
@@ -373,7 +373,12 @@ var qz = (function() {
                         });
                     }
 
-                    _qz.security.callCert().then(sendCert).catch(sendCert);
+                    function catchAndSendNull(error) {
+                        _qz.log.warn("Failed to get certificate:", error);
+                        sendCert(null);
+                    }
+
+                    _qz.security.callCert().then(sendCert).catch(catchAndSendNull);
```
tresf, line 381: It's rare when we use `.then(...)` for us to declare a function like this above when it's small (two lines) and only used in one area.  Of course, we're using a `.catch(...)` here, which doesn't occur often in the API... 

So...  is this code design decision intentened to make debugging easier when exceptions are raised?  If that's the case, I feel the name should be something illustrative to the problem, not the action, e.g. `callCertFailed`, since the `null` value will be obvious based on the logs.  Alternately, the `null` could be part of the warning message.

I believe a named error does improve debugging, so perhaps an inline solution like follows?

```suggestion
                    _qz.security.callCert().then(sendCert).catch(function certFailed(error) {
                        _qz.log.warn("Failed to get certificate:", error);
                        sendCert(null);
                    });
```

bberenz: An anonymous function inline should work just fine, we already know where the warning is coming from so getting extra names in the stacktrace won't do too much anyways.
(And the design decision was more unintentionally mixing styles used in other projects, sorry.)

PR #849, opened 2021-08-10
Add qz:steal support

### LINKED ISSUE #596: qz:steal support
Ability to modify the `qz:launch` behavior to gracefully shut down the OTHER instance of QZ Tray.

* Currently if QZ Tray is started, it will shut down the CURRENT instance of QZ Tray if another instance is running
* It may be beneficial to allow the new instance to "steal" the process and shut down the other
* This can have undefined results when multiple users are working simultaneously, it should only ever be deployed in environments  where workstations are shared (not in a Terminal Services or Citrix style environment when they are using the workstations simultaneously) 

The new feature should be usable through Command Line or HTML/URL:

* Command Line:
   ```bash
   java -jar qz-tray.jar --steal
   ```

* HTML/URL:
   ```html
   <a href="qz:steal">Start QZ Tray, stealing it from another user if possible</a>
   ```

tresf:
Closes #596 

Documentation available:  https://qz.io/wiki/multiple-instances

### DIFF
--- a/src/qz/installer/MacInstaller.java
+++ b/src/qz/installer/MacInstaller.java
@@ -12,7 +12,6 @@ package qz.installer;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import qz.common.Constants;
 import qz.utils.FileUtilities;
 import qz.utils.ShellUtilities;
 import qz.utils.SystemUtilities;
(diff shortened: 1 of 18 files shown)


src/qz/installer/MacInstaller.java
```diff
@@ -12,6 +12,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import qz.common.Constants;
```
bberenz, line 15: Aren't we already static importing this?

bberenz (approved):

src/qz/installer/MacInstaller.java
tresf: Yes, thanks.

Issue #847, opened 2021-08-07
Add SBPL image support


tresf:
QZ Tray can be modified to leverage the SBPL (SATO) programming language for printers such as the GL408e / GL412e.

The image format is very close to some other formats we use:

<img src="https://user-images.githubusercontent.com/6345473/128583999-81aa2035-55f1-47cb-a71e-fdd4189ae664.png" width="500px"/>
<img src="https://user-images.githubusercontent.com/6345473/128583920-cc200937-993f-4b32-91c6-92346b3dd0eb.png" width="500px"/>

[Graphics_Sample.zip](https://github.com/qzind/tray/files/6965578/Graphics_Sample.zip)

tresf:
Closed via b9da58b.

PR #851, opened 2021-08-12
Add SBPL image support

bberenz:
For #847


tresf:
One-off built sent to SATO for testing.  Once they get back with a successful print, we'll merge. Thanks @bberenz!!

PR #876, opened 2021-10-13
Apple installer detects arch

### LINKED ISSUE #875: Warn when running on wrong architecture
The QZ Tray installer should warn when you're running it on the wrong architecture, so something like https://github.com/qzind/tray/issues/873 does not happen.

Vzor-:
Does what it says on the tin

**Edit:** Closes #875

### DIFF
--- a/ant/apple/apple-preinstall.sh.in
+++ b/ant/apple/apple-preinstall.sh.in
@@ -3,11 +3,6 @@
 # Halt on first error
 set -e
 
-if [ $(uname -m) != ${target.arch} ]; then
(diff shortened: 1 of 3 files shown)


tresf:
@Vzor- we're getting some users also complaining that 32-bit Windows doesn't offer an intuitive message (we don't prepare installers for this either).

Vzor-:
image
Yea, that's not great.

tresf:
> Yea, that's not great.

I've moved this to https://github.com/qzind/tray/issues/1027.

tresf:
The additional arch info in the installer knows to install Rosetta now.  That's cool.

image

PR #880, opened 2021-10-21
Add C# signing example using Bouncy Castle library

pasqualirb:


assets/signing/sign-message.bouncycastle.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+
+        // Public method for returning the certificate
+        // ===========================================
+
+        // QZTray.crt is a PEM-encoded certificate signed by the CA (Certificate Authority).
+        //
+        //     Note: It is provided by "QZ Industries, LLC" as "digital-certificate.txt".
+        private string _cert = "QZTray.crt";
```
tresf, line 66: Although there are advantages to using a controller for this, we don't provide this for our back-end examples due to the simplicity of just using AJAX/jquery/fetch.  For this reason, we won't accept this into our examples as it sets a precedent that we do not plan on continuing for all languages.

assets/signing/sign-message.bouncycastle.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+
+        // Public method for signing the input string with the private key
+        // ===============================================================
+
+        // QZTray.key is a PEM-encoded private key of "QZTray.crt".
+        //
+        //     I.e. It's the key of the CSR (Certificate Signing Request), which is the
+        //          original certificate before being signed by the CA key.
+        private string _key = "QZTray.key";
```
tresf, line 38: We do not offer a namespace as it implies our code is a drop-in, which is is not.  Our examples all use `private-key.pem`, this consistency matches our wiki and youtube and cannot change. Minor but we also do not use `_underscore` variable notation as part of our examples.  Next, although the CSR statement is correct (and we support this through our portal) we no longer make mention of CSRs in our signing.  Finally, "PEM encoded" is a bit vague, we would recommend the comment is updated to specify PKCS#8, assuming that format works.  This is intentional on our end as PKCS#1 is a similar format, but breaks many of our examples and enforcing consistency ensures we can continue to support these snippets for years to come.

tresf:
@pasqualirb thanks kindly for the code snippet.  There's a few areas I'd like changed to closer match our examples.  I've made some comments.  If not addressed, that's OK, we would supercede them in a separate PR.

pasqualirb:
> @pasqualirb thanks kindly for the code snippet. There's a few areas I'd like changed to closer match our examples. I've made some comments. If not addressed, that's OK, we would supercede them in a separate PR.

I'll make the changes and as soon as possible I rebase this PR.

assets/signing/sign-message.bouncycastle.cs
pasqualirb: > Finally, "PEM encoded" is a bit vague, we would recommend the comment is updated to specify PKCS#8, assuming that format works.

I had to replace `PemReader` class from `Org.BouncyCastle.Utilities.IO.Pem` with `PemReader` class from `Org.BouncyCastle.OpenSsl` as the latter complies with `PKCS #8` standard.

pasqualirb:
I rebased the PR with the requested changes. Please tell me if you find a problem or if you would like any other change.

tresf:
This looks great, thank you!  I think we can eventually fix our examples to cache the private key on each sign request, but that problem already exists in our other .NET examples so that can be tackled at another time.  Thanks!!

PR #882, opened 2021-11-08
Fix: Support AsymmetricKeyParameter as return value of ReadPrivateKey()

pasqualirb:
I've also fixed indentation of the file with b7c550f21fada264efce13250698ca4db817ab94

### DIFF
--- a/assets/signing/sign-message.bouncycastle.cs
+++ b/assets/signing/sign-message.bouncycastle.cs
@@ -44,8 +44,10 @@ string SignMessage(string msg)
 }
 
 AsymmetricKeyParameter getPrivateKey() {
-    using (var reader = System.IO.File.OpenText(privateKey))
-        return (new PemReader(reader, null).ReadObject() as AsymmetricCipherKeyPair).Private;
+    using (var reader = System.IO.File.OpenText(privateKey)) {
+        var pem = new PemReader(reader).ReadObject();
+        return pem as AsymmetricKeyParameter ?? (pem as AsymmetricCipherKeyPair).Private;
+    }
 }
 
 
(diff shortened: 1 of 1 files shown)


pasqualirb:
References to Bouncy Castle source code:

[When `ReadObject()` returns the result of `ReadPrivateKey()`](https://github.com/bcgit/bc-csharp/blob/release-1.9.0/crypto/src/openssl/PEMReader.cs#L97):
``` cs
public object ReadObject()
{
    PemObject obj = ReadPemObject();

    if (obj == null)
        return null;

    // TODO Follow Java build and map to parser objects?
    //			if (parsers.Contains(obj.Type))
    //				return ((PemObjectParser)parsers[obj.Type]).ParseObject(obj);

    if (Platform.EndsWith(obj.Type, "PRIVATE KEY"))
        return ReadPrivateKey(obj);
```

[When `ReadPrivateKey()` gets the prefix of "PRIVATE KEY"](https://github.com/bcgit/bc-csharp/blob/release-1.9.0/crypto/src/openssl/PEMReader.cs#L238):
``` cs
private object ReadPrivateKey(PemObject pemObject)
{
    //
    // extract the key
    //
    Debug.Assert(Platform.EndsWith(pemObject.Type, "PRIVATE KEY"));

    string type = pemObject.Type.Substring(0, pemObject.Type.Length - "PRIVATE KEY".Length).Trim();
```

When `ReadPrivateKey()` returns [`AsymmetricKeyParameter`](https://github.com/bcgit/bc-csharp/blob/release-1.9.0/crypto/src/openssl/PEMReader.cs#L352) and [`AsymmetricCipherKeyPair`](https://github.com/bcgit/bc-csharp/blob/release-1.9.0/crypto/src/openssl/PEMReader.cs#L359):
``` cs
    case "":
    {
        return PrivateKeyFactory.CreateKey(PrivateKeyInfo.GetInstance(seq));
    }

    default:
        throw new ArgumentException("Unknown key type: " + type, "type");
}

return new AsymmetricCipherKeyPair(pubSpec, privSpec);
```

Issue #889, opened 2021-12-03
HTTPS issue on Firefox Developer Edition on macOS


sheck:
I think this might be a possible regression, similar to #200 

When using:
- macOS 11.6.1
- Firefox Developer Edition 95.0b12 (64-bit)
- QZ tray 2.1.4 (official and branded builds)

I experience consistent HTTPS errors when visiting `https://localhost:8181` and inconsistent failures using qz tray (via websockets).

I have confirmed that on my machine Firefox 94.0.2 and Chrome 96 do not have these issues (see screenshots below). Several co-workers are able to reproduce this behavior as well.

<img width="1536" alt="CleanShot 2021-12-03 at 12 23 19@2x" src="https://user-images.githubusercontent.com/3035506/144669880-ae2e4b61-1eb4-4d63-b0f8-bdc1a4952ac6.png">

<img width="1536" alt="CleanShot 2021-12-03 at 12 22 40@2x" src="https://user-images.githubusercontent.com/3035506/144669910-36c069b6-c058-407f-8842-dea0a157b51c.png">

tresf:
On Mac, the Firefox property `security.enterprise_roots.enabled` is not toggled on for Firefox Developer Edition.

This is something we can fix by expliciting searching for `Firefox Developer Edition` at install time and shimming this policy.

As a temporary workaround, please visit `about:config` and toggle this policy on manually.

tresf:
Closed via #891.  This patch will be available in QZ Tray 2.2.1 (and if released before EOL, QZ Tray 2.1.5).

PR #891, opened 2021-12-09
Add support for experimental Firefox releases

### LINKED ISSUE #217: Support for Developer and "Nightly" editions of Firefox
In addition to the stable version of Firefox, there are "Firefox Developer Edition" and "Firefox Nightly" versions: https://www.mozilla.org/en-US/firefox/channel/desktop/

QZ Tray does not work over wss with these versions. Since these versions are not stable, and unlikely to rolled out to production users, is it worth adding support for these versions?

One pro, is that we will be able to fully test QZ Tray + SSL before the features even hit the beta release.

tresf:
Closes #217, #889

As @sheck points out in #889, Firefox Developer Edition is not properly supported on MacOS.  This PR adds support for detecting these browsers for Policy File installation.

### DIFF
--- a/src/qz/installer/certificate/firefox/locator/AppAlias.java
+++ b/src/qz/installer/certificate/firefox/locator/AppAlias.java
@@ -5,7 +5,9 @@ import java.util.Locale;
 public enum AppAlias {
     // Tor Browser intentionally excluded; Tor's proxy blocks localhost connections
     FIREFOX(
-            new Alias("Mozilla", "Mozilla Firefox", "org.mozilla.firefox", true), // Windows
+            new Alias("Mozilla", "Mozilla Firefox", "org.mozilla.firefox", true),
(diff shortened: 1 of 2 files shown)


tresf:
Firefox nightly uses a version (e.g. `97.0a1`) which breaks semantic versioning.  This means we can't compare the version and check if policies were supported.  For now, I've shimmed a workaround in ab32b9c to change this to `97.0.0-unknown`.  I can't think of any adverse side-effects except that if we're to test nightly versions for a particular feature, detecting that exact version (e.g. `97.0a1` versus `97.0` will not be impossible).

tresf:
Merging.  This feature will be part of QZ Tray 2.2.1 (as well as QZ Tray 2.1.5 if a release is made before EOL).

tresf:
Cherry-picked to 2.1 branch via f4e8dba.

Issue #440, opened 2019-04-10
Dual trust support for perpetual license model


tresf:
As part of our ongoing effort to better serve our clients, we recognize a need to support a perpetual license model.  The following requirements must be met:
* Perpetual licenses (keypair that's trusted and doesn't expire) must not be part of the QZ Industries CA Root chain of trust.
* The license must be able to coexist with websites that do not use a perpetual license.
   * For example, if a website offers a build of QZ Tray with a custom keypair/perpetual license baked in, there must be a way for machines to still be able to use websites that are on a subscription model.
   * The user should be offered an override in the gui (and on 2.1, through CLI) to allow QZ Industries CA Root chain to be toggled on.
* This feature can potentially improve the "community" version as well, offering a path for community builds to still work with a "premium" website.  The logic would be identical.
* As a practice, clients that choose to leverage a perpetual license should white-label our logo off of the installer to help distinguish a QZ installer versus a custom installer.
* In the event of more than one perpetual licenses, offer a technique for side-loading a 2nd, 3rd keypair in addition to the one provided with the custom installer.

tresf:
Mockup: 
![image](https://user-images.githubusercontent.com/6345473/55902326-40651100-5b99-11e9-82a6-336293c2c751.png)

tresf:
The UI components have been written on the `perpetual` branch.  I've moved the checkbox to the Site Manager dialog.

The ability to "Trust QZ Certificates" has been renamed to "Strict certificate mode".  Strict certificate mode will **BLOCK** any QZ-issued certs from being trusted.

The default behavior will be to **OFF** and will allow perpetual certs and QZ certs to live in harmony with each other.

![image](https://user-images.githubusercontent.com/6345473/92059988-a5a63d00-ed60-11ea-86bd-411df14f99db.png)

PR #899, opened 2021-12-17
fix(osx): bundle override.crt file in Contents/Resources

krzaczek:
Hello,

When building package for osx with  `-Dauthcert.use= ` override.crt is not copied from `dist` into `Contents/Resources`.


tresf:
@krzaczek thank you, I'm surprised we missed this.  @Vzor-, FYI.

Related #440 #708.

tresf:
Duplicate: https://stackoverflow.com/q/70198982/3196753

PR #901, opened 2021-12-21
Fix forceRaw / altPrinting upgrade routine

### LINKED ISSUE #898: demo.qz.io uses "Force Raw" instead of "altPrinting", which doesn't work with 2.1.5 and lower
I noticed this on a client's machine a few days ago and confirmed today. 

The newer "Force Raw" flag is available on our test page https://demo.qz.io , but if the client machine is not using 2.2.0+, the printer will print the commands instead of parsing them.

tresf:
Closes #898

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -788,8 +788,8 @@ var qz = (function() {
                         config.rasterize = true;
                     }
                 }
-                if(_qz.tools.versionCompare(2, 1, 4) < 0) {
-                    if(config.forceRaw) {
+                if(_qz.tools.versionCompare(2, 2) < 0) {
+                    if(config.forceRaw !== 'undefined') {
                         config.altPrinting = config.forceRaw;
                         delete config.forceRaw;
                     }


klabarge (approved):
I tested this change locally on macOS using QZ Tray 2.1.5 and it works

tresf:
> I tested this change locally on macOS using QZ Tray 2.1.5 and it works

Awesome.  We also need to regression test this against the known-working QZ Tray 2.1.3 as well as make sure it causes no regressions with newer versions such as QZ Tray 2.2.1.

tresf:
> Awesome. We also need to regression test this against the known-working QZ Tray 2.1.3 as well as make sure it causes no regressions with newer versions such as QZ Tray 2.2.1.

Clarification: The new 2.2.1 version should be confirmed via the logs to ensure that `forceRaw` comes across still, not `altPrintint`.

klabarge:
I've confirmed that 2.2.1 uses `forceRaw` 👍

Issue #914, opened 2022-02-02
qz.file.write should allow data flavors such as base64


tresf:
Currently, [`file.write`](https://qz.io/wiki/file#write-data) assumes all data is in plain text/binary, but often formats such as base64 are preferred.  The API should allow a data `flavor: ` parameter to allow URL, base64, etc.

PR #917, opened 2022-02-04
Support for data flavors in file writes

bberenz:
for #914


tresf:
Thanks @bberenz!  `BASE64` and `FILE` are working perfectly.  Can we also add `HEX` to that listing before merging?

Issue #924, opened 2022-03-01
Windows Server 2012 Crash when Java FX starts


tresf:
Note, this issue is not reproducible on a fresh install, but was present on a Xeon CPU running on an AHV Nutanix virtualized environment with Adoptium and AdoptOpenJDK with QZ Tray 2.1.5.

```
Faulting application name: javaw.exe, version: 11.0.11.0, time stamp: 0x607f471a
Faulting module name: ucrtbase.DLL, version: 10.0.17134.12, time stamp: 0x587decd7
Exception code: 0xc0000409
Fault offset: 0x000000000006e75e
Faulting process id: 0x2b2c0
Faulting application start time: 0x01d82d92b3f17d62
Faulting application path: C:\Program Files\AdoptOpenJDK\jdk-11.0.11.9-hotspot\bin\javaw.exe
Faulting module path: C:\Program Files\AdoptOpenJDK\jdk-11.0.11.9-hotspot\bin\ucrtbase.DLL
Report Id: 8296f45a-9986-11ec-838a-e2e70c4793f1
Faulting package full name: 
Faulting package-relative application ID: 
```

A temporary stop-gap for this hard-crash is to provide a system property/override to not initialize JavaFX on startup.

@bberenz can you add the property to 2.1 (we can later cherry-pick to 2.2).

tresf:
We've tested QZ Tray 2.1.5 with the Java FX task on the following:

✅  Windows Server 2012 R2 Build 9600 in VirtualBox
✅  Windows Server 2012 R2 Build 9600 on Amazon EC2
✅  Windows Server 2012 R2 Build 9600 on bare metal
✅  Windows Server 2012 R2 Build 9600 on Nutanix

All environments used Xeon processors, none crashed.

I'm closing this as invalid until we have steps to reproduce.

tresf:
Reopening... 

I was given an export of all printer mappings on the crashing server, imported it into Windows Server 2012 and the issue is 100% reproducible.

It seems the issue occurs when the printer searching takes a very long time (e.g. 5 minutes) and JavaFX then crashes during its capture.

tresf:
This issue is quite odd.  As it turns out, something with fetching `PrintResolution` for 1,400 printers (specific ones imported from the client environment) followed by starting JavaFX crashes the runtime.

✅ If we start JavaFX first, no crash.
✅ If we fetch these `PrintResolution` for a percentage (e.g. 300), no crash.
[...]

PR #928, opened 2022-03-02
Allow idle tasks to be flagged as disabled

### LINKED ISSUE #926: Provide system property to disable startup tasks
Issue #924 describes a crash which could be delayed by disabling the JavaFX startup task.

Currently, QZ Tray has two startup tasks:
* Printer details
* JavaFX

More may be added in the future, but granularity over these tasks may be needed as stop-gap solutions to problems.

QZ should provide a way to disable each task through a system property.

bberenz:
For #924

Closes #926 

Idle tasks can be disabled by adding `tray.idle.printers=false` or `tray.idle.javafx=false` to the prefs.properties file.

### DIFF
--- a/src/qz/common/Constants.java
+++ b/src/qz/common/Constants.java
@@ -61,6 +61,8 @@ public class Constants {
     public static final String PREFS_HEADLESS = "tray.headless";
     public static final String PREFS_MONOCLE = "tray.monocle";
     public static final String PREFS_STRICT_MODE = "tray.strictmode";
+    public static final String PREFS_IDLE_PRINTERS = "tray.idle.printers";
+    public static final String PREFS_IDLE_JFX = "tray.idle.javafx";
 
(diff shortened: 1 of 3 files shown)


tresf:
U

> Thanks, I've reached out to the client to see if user preferences is suitable, or if it's better as a `System.getProperty(...)`, or environment variable, etc.

User says user preference is fine, but opened PR #930 anyway as -- in my experience -- it's much easier to control these through system-wide properties.

Note, a TODO item on `master` is to move preferences into a list so we can spit them out somewhere into documentation, e.g. https://qz.io/wiki/command-line#command-line-usage.

tresf:
This is testing well, will merge to 2.1.  @klabarge if you have some time to do some additional testing (e.g. Monocle regression testing), please do before we make a new 2.1 release.

PR #929, opened 2022-03-02
Fix performance issues when finding all printers

### LINKED ISSUE #925: Background task for printer fetching is slower than expected
Today, a client with ~1,200 printers was able to illustrate that with QZ Tray 2.1.5, the startup task which fetches the printer details does not properly cache them.

This can be observed as follows:

✅  Initial background task time: ~3 minutes
🚫 First search for printers after background task is completed: ~1 minute
✅ Second search for printers: 2 seconds

@bberenz can you see if we have something awry with our printer searching logic?

tresf:
Closes #925


tresf:
5734879 helps see this issue in real time, it shows a message when the Trays are being crawled.   I'm keeping the debug in there, I think it will help us in the future.

PR #933, opened 2022-03-09
idle hang (2.1)

bberenz:
address UI hanging issue mentioned in #925 by using util printer to avoid calling idle actions on ui threads


tresf:
Merged to 2.1, needs to be be ported to master.

PR #941, opened 2022-03-17
Dedicated class for WebSocket Errors

### LINKED ISSUE #937: Proposal for errors to have unique IDs
Throughout QZ's life cycle, it can throw quite a few errors to the browser.

https://github.com/qzind/tray/search?q=sendError

However, translating these to the front end (e.g. for non-English) can be tricky as it results in trial and error, or scraping the code.

Goals:
* Using unique IDs for the errors will give translator a reliable identifier for translating.
* It should be possible to use code to prevent an ID from being reused
* It should be possible to procedurally generate a list of errors

tresf:
Closes #937 

Still in draft form, the class is just a proof of concept.

TODO:

* [ ] Return error numbers to JavaScript
* [ ] Migrate existing errors to only allow new class
* [ ] Add query support for getting an exhaustive listing

### DIFF
--- a/src/qz/exception/WebsocketError.java
+++ b/src/qz/exception/WebsocketError.java
@@ -1,0 +1,57 @@
+package qz.exception;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import static qz.exception.WebsocketError.WebsocketErrorType.*;
+
+public enum WebsocketError {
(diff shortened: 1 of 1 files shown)


tresf:
@bberenz can you help with the way we're throwing the `DeviceException` up the stack?  The error clearly states that the "USB Device cannot be claimed", which should be reasonable to catch, but instead it throws it all the way up to the socket enum, causing a general error.

Issue #877, opened 2021-10-13
DEMO: Serial should unescape hex


tresf:
Currently the demo only unescapes a few characters like `\n`.

https://github.com/qzind/tray/blob/b5ed5856df37217ad317139b15d257c5a9b198d8/sample.html#L1999-L2000

PR #949, opened 2022-03-29
Unescape hex/unicode for sample serial data

bberenz:
For #877

### DIFF
--- a/sample.html
+++ b/sample.html
@@ -2021,8 +2021,9 @@
             type: $("input[name='serialType']:checked").val(),
             data: $("#serialCmd").val()
         };
-        //allow some escape characters
-        serialData.data = serialData.data.replace(/\\n/g, "\n").replace(/\\r/g, "\r").replace(/\\t/g, "\t");
+        //allow some escape characters (newlines, tabs, hex/unicode)
+        serialData.data = serialData.data.replace(/\\n/g, "\n").replace(/\\r/g, "\r").replace(/\\t/g, "\t")
+            .replace(/\\[xu]([\dA-Fa-f]+)/g, (m, s1) => String.fromCharCode(parseInt(s1, 16)));
 
         qz.serial.sendData($("#serialPort").val(), serialData, options).catch(displayError);
     }
--- a/sample.html
+++ b/sample.html
@@ -2021,9 +2021,14 @@
             type: $("input[name='serialType']:checked").val(),
             data: $("#serialCmd").val()
         };
+
+        var fromHex = function(m, s1) {
+            return String.fromCharCode(parseInt(s1, 16));
+        };
+
         //allow some escape characters (newlines, tabs, hex/unicode)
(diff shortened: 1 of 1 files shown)


sample.html
```diff
@@ -2021,8 +2021,9 @@ <h4 class="panel-title">Options</h4>
             type: $("input[name='serialType']:checked").val(),
             data: $("#serialCmd").val()
         };
-        //allow some escape characters
-        serialData.data = serialData.data.replace(/\\n/g, "\n").replace(/\\r/g, "\r").replace(/\\t/g, "\t");
+        //allow some escape characters (newlines, tabs, hex/unicode)
+        serialData.data = serialData.data.replace(/\\n/g, "\n").replace(/\\r/g, "\r").replace(/\\t/g, "\t")
+            .replace(/\\[xu]([\dA-Fa-f]+)/g, (m, s1) => String.fromCharCode(parseInt(s1, 16)));
```
klabarge, line 2026: This arrow function breaks IE11 support

klabarge (changes requested):

sample.html
klabarge: Can this be rewritten like so?
```js
        serialData.data = serialData.data.replace(/\\n/g, "\n").replace(/\\r/g, "\r").replace(/\\t/g, "\t")
            .replace(/\\[xu]([\dA-Fa-f]+)/g, (m, s1).then(function() {
                String.fromCharCode(parseInt(s1, 16))
            })
        );
```

bberenz: yes, but i found a case in retesting where the \x codes can try to collect too many values to convert, so I pushed up a change where the function is a variable to be used for both

klabarge:
What data should I be testing?

I'm kind of arbitrarily sending stuff so not sure if it's valid.
#### Test 1
**send:** `hi, serial port\n`
**logs:**`"data":"hi, serial port\n"`

#### Test 2
**send:** `hi, serial port\u`
**logs:**`"data":"hi, serial port\\u"`
**error:** `java.lang.IllegalArgumentException: Less than 4 hex digits in unicode value: '\u' due to end of CharSequence`

tresf:
> What data should I be testing?
> 
> I'm kind of arbitrarily sending stuff so not sure if it's valid.
> 
> #### Test 1
> **send:** `hi, serial port\n` **logs:**`"data":"hi, serial port\n"`
> 
> #### Test 2
> **send:** `hi, serial port\u` **logs:**`"data":"hi, serial port\\u"` **error:** `java.lang.IllegalArgumentException: Less than 4 hex digits in unicode value: '\u' due to end of CharSequence`

Those are pretty good tests.  Some known hex would be nice (newline, tab, a-z stuff, but in hex format).

klabarge:
I sent  hex with new lines and tabs and it seems ok.

```
68 69 0A 09 73 65 72 69 61 6C 0A 70 6F 72 74
```

tresf:
```
thanks
	kyle
merging
```

Issue #665, opened 2020-06-08
java.lang.IndexOutOfBoundsException: Bound exceeds available space: size=3, offset=39


tresf:
Similar symptomatically to #619, if the latest API is being used against an older QZ Tray version, the data is reformatted `{ data: { data: ... } }` since https://github.com/qzind/tray/commit/1148e5b1fa53022e8b107f981b94bf1559cf89d0#diff-5b383df0ac6d733ed512ffe910a83668R1777-R1782 and is imcompatible with older QZ tray versions.

Although this new data format works with 2.1.1+, for older versions it is a serious regression preventing HID and USB calls from succeeding.

PR #951, opened 2022-03-29
Additional backwards conversion for newer data formats

bberenz:
for #665

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -608,9 +608,9 @@ var qz = (function() {
 
             /** Stub for rejecting with an Error from withing a Promise */
             reject: function(error) {
-                return _qz.tools.promise(function(resolve, reject)) {
+                return _qz.tools.promise(function(resolve, reject) {
                     reject(error);
-                }
+                })
             },
 
             stringify: function(object) {
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -610,7 +610,7 @@ var qz = (function() {
             reject: function(error) {
                 return _qz.tools.promise(function(resolve, reject) {
                     reject(error);
-                })
+                });
             },
 
             stringify: function(object) {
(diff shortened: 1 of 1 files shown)


klabarge:
>but I'm pretty sure the backticks break IE11 support, which we haven't officially done yet.
Confirmed that this does break IE11.

tresf:
@klabarge I did a quick edit through the web interface if you don't mind trying it again.

tresf:
@klabarge bump. :)

klabarge:
This does work.

image

However, when using the `sample.html`, the code never makes it to the `catch`  therefore does not show an error to the user. Can we display an error message to the user?

```js
    function sendHidData() {
        qz.hid.sendData({
                            vendorId: $("#hidVendor").val(),
                            productId: $("#hidProduct").val(),
                            usagePage: $("#hidUsagePage").val(),
                            serial: $("#hidSerial").val(),
                            data: {data:"dd", type: "HEX"},
                            endpoint: $("#hidReport").val()
                        })
            .catch(displayError);
    }
```

tresf:
Good idea, prob shoulda been there all along.  Do you mind adding the commit?

klabarge:
@tresf I am struggling to catch this error properly.

If I switch the line to `Promise.reject`, the promise seems to hit the catch block.
   * Notice the additional  `no output reports` error in the console (which is because the HID device I am testing against doesn't support sending data to it).

However, the `Data format` error still shows as uncaught.

```js
Promise.reject("Data format is not supported with connected QZ Tray version " + _qz.websocket.connection.version);
```

image

tresf:
@klabarge yeah, I appears @bberenz is throwing this in the middle of a promise, so it's unhandled.

@bberenz is this behavior desired?  As-is, it can't be caught.

bberenz:
You need to change it to `return Promise.reject(new Error("..."));` for this to work how you are expecting.

tresf:
Assuming we wrap it in our `_qz.tools.proimse` to make it Promise agnostic (RSVP, Q, BlueBird, etc), would this work?

```js
   tools:
        // ... 
            /** Stub for rejecting with an Error from withing a promise */
            reject: function(error) {
                return _qz.tools.promise(new function(resolve, reject)) {
                    reject(error);
                }
            },
```


Usage:
```js
return _qz.tools.reject(new Error("..."));

```

tresf:
Quoting @bberenz 

> 👍 

😅 OK, @klabarge great catch.... Can you test e7e714 and see if it remedy's it?  It adds a new `_qz.tools.reject(new Error(...))` syntax that allows just this.

klabarge:
I think there was a syntax error which I fixed in [85540f2](https://github.com/qzind/tray/pull/951/commits/85540f248a0ff17871217eeec684b5f39fdeb325)

However, it's still showing as an `Uncaught (in promise) Error` in the browser console.

tresf:
Thanks for fixing the typo.  I think I forgot a `return`.  Can you test with f3efde9?

klabarge:
Yes that did the trick!

tresf (approved):

Issue #952, opened 2022-03-30
Option to ignore PDF transparency when printing


tresf:
Per:
https://lists.apache.org/thread/7ls9qpzlt21wqy2goj25b85l960j8dvb

Related: #278 

Quoting:

> Hi,
> Both options are related to transparency groups. These are an extension
> of `XObject` forms. I thought about ignoring them, i.e. treat them like
> ordinary forms.
> So I looked at `PDXobject.createXObject();` this one is called by
> `PDResources.getXObject()`, which is called by `DrawObject.process()`. That
> one class can officially be replaced if you extend `PageDrawer `and call
> "`addOperator(new DrawObject());`" in the constructor; that class should
> have  a slightly modified code that does NOT call
> `showTransparencyGroup()`, but only `showForm()` (there are 3 `DrawObject`
> classes, copy the one in
> `org.apache.pdfbox.contentstream.operator.graphics`). You can extend
> `PageDrawer`, but you'll have to extend `PDFRenderer` as well, see class
> comment of `PageDrawer`.
> 
>  (I haven't tested this myself)

tresf:
Documentation published https://qz.io/wiki/pixel#pdf-transparency.   This feature will be available starting with QZ Tray 2.2.2.

Issue #278, opened 2018-02-23
PDF rasterized when instructed not to


klabarge:
A client is experiencing an issue with PDF's that grow very large in the printer queue (700MB) that result in very slow print speeds. Reproducing the issue exposed that that vector PDF was being rasterized despite `rasterize: false`

### Environment

**OS:** Windows 10
**QZ Tray:** 2.0.5
**Java:** 8u161 64-bit

* [Here is the offending PDF](https://github.com/qzind/tray/files/1749977/test.pdf) that can be used to reproduce the exact issue

**UPDATE:**  The basic PDF ([the underlying form/template attached here](https://github.com/qzind/tray/files/1750080/template.pdf)) was made with Adobe Illustrator, the filled data is another PDF; merged together with [FPDI](https://www.setasign.com/products/fpdi/about/).

## Expected Results
![screen shot 2018-02-22 at 9 52 57 pm](https://user-images.githubusercontent.com/6345473/36575790-d730b36e-181a-11e8-8eea-855b05383b2e.png)

### PDFCreator

1. When `rasterize` is `false`, the print is a rasterized image.
    ![image](https://user-images.githubusercontent.com/12505463/36571015-c1ea7d5e-1802-11e8-8bfb-84983dd0479a.png)

1. Printing the offending PDF using the virtual printer PDFCreator results in a 700MB print job.
   * Once spooled, the file is about 6.5MB on disk

   ![image](https://user-images.githubusercontent.com/12505463/36570826-e6b6a99c-1801-11e8-99ad-ee8b9ffb0fb5.png)

### Microsoft Print to PDF

1. When `rasterize` is `false`, the print is a rasterized image.

   ![image](https://user-images.githubusercontent.com/12505463/36573252-e890d4e8-180d-11e8-97bf-96fbacb28eb1.png)

1. I was unable to reproduce the large print job using the Microsoft Print to PDF printer
   ![image](https://user-images.githubusercontent.com/12505463/36573218-beaa0f96-180d-11e8-840b-8d3d9fca3aca.png)

### Stacktrace

```
[DEBUG] 2018-02-22 20:14:31,692 @ qz.ws.PrintSocketClient:?
[...]

PR #959, opened 2022-04-13
Pdf transparency

bberenz:
for #952


src/qz/printer/action/PrintPDF.java
```diff
@@ -82,6 +83,10 @@ public void parseData(JSONArray printData, PrintOptions options) throws JSONExce
                     docHeight = dataOpt.optDouble("pageHeight") * convert;
                 }
 
+                if (!dataOpt.isNull("ignoreTransparency") && dataOpt.optBoolean("ignoreTransparency")) {
+                    ignoreTransparency = true;
+                }
```
tresf, line 88:
```suggestion
                ignoreTransparency = dataOpt.optBoolean("ignoreTransparency", false);
```

klabarge:
I tested this on macOS using the PDF in https://github.com/qzind/tray/issues/278 and this works great. The left is with `ignoreTransparency: true`, the right is with `ignoreTransparency: false`


image

tresf:
@klabarge thanks!  Will you be able to test on Windows as well as do some basic regression testing?

klabarge:
> Will you be able to test on Windows

I actually couldn't reproduce the poor printing quality in Windows, however I was able to reproduce the slow spooling/printing in #278. `ignoreTransparency: true` fixes the slow printing.

> as well as do some basic regression testing?

I tested backwards and forwards compatibility ( new api vs old app, and vice versa) on Windows with success.

PR #967, opened 2022-04-25
Move sample dependencies out of package

mamhoff:
These dependencies don't seem to be used from the package that is
shipped via NPM, but only from the sample app. Let's not vendor them
there, but rather put them into their own directory.


tresf:
Thanks.  We have no intentions on moving these files elsewhere however we would consider an `npmignore` strategy which would maintain the same structure we have now, but without sending them over to NPM.

mamhoff:
Updated the PR to use the provided `js/.npmignore` file.

PR #980, opened 2022-07-06
Add detection for missing SystemTray on Windows

### LINKED ISSUE #958: Detect hidden tray area on Windows
Windows 11 has introduced a feature that allows system tray icons to be hidden, which will prevent any access to the QZ Tray menu system.

https://www.neowin.net/news/windows-11-22h2-will-let-you-disable-all-app-icons-in-the-notification-area/

We should offer an alternative UI for this scenario, similar to Linux.

**Edit:** Windows 2003 and higher also has a policy to hide this area.  Title changed to reflect.

Related: #878

tresf:
Closes #958

### DIFF
--- a/src/qz/utils/SystemUtilities.java
+++ b/src/qz/utils/SystemUtilities.java
@@ -729,4 +729,25 @@ public class SystemUtilities {
         }
         return false;
     }
+
+    /**
+     * Cross-platform SystemTray detector
+     */
+    public static boolean isSystemTraySupported(boolean headless) {
+        if(!headless) {
+            switch(getOsType()) {
+                case WINDOWS:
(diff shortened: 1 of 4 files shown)


src/qz/utils/SystemUtilities.java
```diff
@@ -729,4 +729,25 @@ public static boolean validateSaltedChallenge(String message) {
...
+            switch(getOsType()) {
+                case WINDOWS:
+                    if(WindowsUtilities.isHiddenSystemTray()) {
+                        return false;
+                    }
+                    break;
+                case MAC:
+                    break;
+                default:
+                    // TODO: Default to false for Linux?
```
tresf, line 747: Will address this in a separate PR.

PR #994, opened 2022-08-03
Don't mandate primary adapter when fetching hostname

### LINKED ISSUE #982: getHostName() shouldn't require network adapter
QZ Tray offers the ability to get the computer's `hostname` (usually by calling "hostname" from a shell).

This API is naturally grouped into the [`qz.networking`](https://qz.io/api/qz.networking) namespace, but this puts an unnecessary dependency on obtaining the active network adapter, which can be problematic in environments with limited access to the internet.  This is because obtaining the active adapter is done by making a socket connection to google.com (configurable), which may fail.

We should offer an additional API to get just the hostname so that this can be used without the reliance on finding the active adapter.

tresf:
Closes #982


tresf:
The build errors are unrelated to this PR, and are explained in #995

PR #1008, opened 2022-08-18
Remove signing params on open calls

### LINKED ISSUE #993: A signature on getVersion() is confusing and not needed
The initial call to `getVersion` will contain a signature, but this signature is confusing since often it comes before the certificate is sent.  The signature for `getVersion` is never validated because it is not a privileged function. For this reason, we should NOT send a signature when calling `getVersion` from our API.

The reason this is confusing is because when troubleshooting signing issues, it symptomatically seems as if this signature is OK, but it's actually being ignored.

Example JSON that has a signature that can never be validated and can just add confusion when troubleshooting:

```json
{"call":"getVersion","signature":"AAAAAAAAAAAAAAAA...","promise":{},"timestamp":1659461292873,"uid":"yge4zr","signAlgorithm":"SHA512","position":{"x":720,"y":425.5}}
----------------------^
```
```json
{"certificate":"-----BEGIN CERTIFICATE-----...-----END CERTIFICATE-----\n","promise":{},"timestamp":1659461497358,"uid":"ag13wt","signAlgorithm":"SHA512","position":{"x":720,"y":425.5}}
```

bberenz:
Fixes #993

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -226,9 +226,6 @@ var qz = (function() {
                             _qz.websocket.pendingCalls[obj.uid] = obj.promise;
                         }
 
-                        //ensure we know how this was signed
-                        obj.signAlgorithm = _qz.security.signAlgorithm;
-
                         // track requesting monitor
                         obj.position = {
                             x: typeof screen !== 'undefined' ? ((screen.availWidth || screen.width) / 2) + (screen.left || screen.availLeft) : 0,
@@ -236,7 +233,7 @@ var qz = (function() {
                         };
 
                         try {
-                            if (obj.call != undefined && obj.signature == undefined) {
+                            if (obj.call != undefined && obj.signature == undefined && _qz.security.needsSigned(obj.call)) {
                                 var signObj = {
                                     call: obj.call,
                                     params: obj.params,
@@ -255,7 +252,8 @@ var qz = (function() {
                                     return _qz.security.callSign(hashed);
                                 }).then(function(signature) {
                                     _qz.log.trace("Signature for call", signature);
-                                    obj.signature = signature;
+                                    obj.signature = signature || "";
+                                    obj.signAlgorithm = _qz.security.signAlgorithm;
 
                                     _qz.signContent = undefined;
                                     _qz.websocket.connection.send(_qz.tools.stringify(obj));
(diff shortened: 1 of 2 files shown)


js/qz-tray.js
```diff
@@ -589,7 +587,25 @@ var qz = (function() {
             },
 
             /** Signing algorithm used on signatures */
-            signAlgorithm: "SHA1"
+            signAlgorithm: "SHA1",
+
+            needsSigned: function(callName) {
+                const undialoged = [
+                    "printers.getStatus",
```
tresf, line 594: Just a note... I was curious why "printers.getStatus" wasn't signed and it's because you MUST start listening first, so order of operations mandates that.  I tried abusing this cross-site (starting listener on one website, but getting status from another) and it's handled properly, the second site won't allow getting the status from the former listener, so this is good despite how odd it looks in the listing.
https://github.com/qzind/tray/blob/48eda67666aa11bc29299d78d1568fe52e93cc0a/src/qz/ws/SocketMethod.java#L8

src/qz/ws/PrintSocketClient.java
```diff
@@ -176,6 +176,19 @@ public void onMessage(Session session, Reader reader) throws IOException {
         }
     }
 
+    private JSONObject cleanupMessage(JSONObject msg) {
+        msg.remove("promise"); //never needed java side
```
tresf, line 180: @bberenz if this is getting accidentally serialized, should we be cleaning this up JavaScript side too or is that too difficult?

tresf:
@klabarge checking for availability for testing and regression testing (old+new, new+old, new+new).

src/qz/ws/PrintSocketClient.java
bberenz: we use it js side, it is functions though which don't serialize so its always blank here

tresf: So it turns out stringify supports a replacer function, so we can remove this from the API as well.

https://stackoverflow.com/questions/4910567/hide-certain-values-in-output-from-json-stringify
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#parameters

tresf: @bberenz checking in on the `replacer` proposal to remove this from JS as well.

tresf:
@bberenz this code will still show the following in the logs:

```diff
	Message: {"call":"getVersion","timestamp":1661355146986,"uid":"knjijj","position":{"x":720,"y":425.5}}
[WARN] 2022-08-24T11:32:26,988 @ qz.ws.PrintSocketClient:-1
-	Bad signature on request
```

klabarge:
Initial testing looks good. I did test this in an environment where signing is not set up, so I need to verify these results again.

|API                               |   QZ      |  Pass/Fail  | Notes                                 |
|-------------------- |---------|-----------|--------------------------|
| #1008                       | #1008    |        ✅      |  Signing not in logs           |
| 2.2.2-SNAPSHOT   | #1008     |        ✅      |  Signing not in logs           |
| #1008                      | 2.1.6(+3)|         🟡     | Signing attempt is in logs |

klabarge (approved):

Issue #874, opened 2021-10-06
QZ Tray uninstall does not work on Mac


tresf:
Since QZ Tray 2.2.0, the `uninstall` script no longer seem to remove QZ Tray from the Applicaitons folder.

```bash
sudo /Applications/QZ\ Tray.app/Contents/Resources/uninstall
```

tresf:
@Vzor- can you please patch in a fix before the 2.2.2 version?

hellokittychaos:
Having same issue on PC Win 10 @tresf  Did you find a fix?  Tried to run it in cmd as a service then uninstall in cmd but it isn't working either. cd C:\nssm\win64

tresf:
> Having same issue on PC Win 10 @tresf Did you find a fix? Tried to run it in cmd as a service then uninstall in cmd but it isn't working either. cd C:\nssm\win64

@hellokittychaos I think your issue is unrelated.  The above issue is for the MacOS operating system only.

If you're having issues uninstalling QZ Tray on a Windows 10 PC, please check that your antivirus software isn't interfering.  Some AVs (such as Comodo) cause a lot of problems with QZ Tray: https://qz.io/wiki/faq#conflicting-applications

Issue #1006, opened 2022-08-14
MacOS:  Uninstall script doesn't remove QZ Tray


tresf:
QZ Tray 2.2.1, running `/Applications/QZ\ Tray.app/Contents/Resources/uninstall` does not fully remove the application.

tresf:
@Vzor- I believe we need to check the location of the uninstall script and handle the new `Contents/Resources` home.

https://github.com/qzind/tray/blob/875c7038cf18a4ae8ac440627f13230aeddaecee/ant/unix/unix-uninstall.sh.in#L11-L23

tresf:
Closed via #1009.

PR #1009, opened 2022-08-18
Unix uninstaller fix

Vzor-:
Fixes https://github.com/qzind/tray/issues/874 and https://github.com/qzind/tray/issues/1006

linux-uninstall was delete, it seems to be redundant.


tresf:
@Vzor- I think we can avoid a secondary variable by doing "Contents/Resources" detection, like we do here:
https://github.com/qzind/tray/blob/edcfa39b04df7fc12da6c49e882f7dadc6bd56cc/src/qz/utils/SystemUtilities.java#L332

This would allow us to simplify our logic to a single conditional, and just track `$APP_DIR` (what you were calling `$ROOT_DIR`) since the `$DIR` value really has no value to us otherwise.

For example:

```bash
#!/bin/bash

# Halt on first error
set -e

if [ "$(id -u)" != "0" ]; then
   echo "This script must be run with root (sudo) privileges" 1>&2
   exit 1
fi

# Calculate application directory from working directory
APP_DIR=$(cd "$(dirname "$0")" && pwd)

echo "Running uninstall tasks..."
if [[ "$APP_DIR" == *"Contents/Resources"* ]]; then
    APP_DIR="$APP_DIR/../../" # Calculate path to .app
    pushd "$APP_DIR"
    "$APP_DIR/Contents/MacOS/${project.name}" uninstall
else
    pushd "$APP_DIR"
    "$APP_DIR/${project.filename}" uninstall
fi

echo "Deleting files..."
rm -rf "$APP_DIR"
echo -e "\nUninstall of ${project.name} complete.\n"

popd &>/dev/null
```

tresf:
I do really like the additional `.app` guard, but I couldn't find an equivalent command for Linux, so I removed it for now.

tresf:
I've changed my mind.  I think the `$ROOT_DIR` approach is cleaner and I really want to keep the `.app` guard.  I'm pushing a commit which organizes and comments the code in a way I think will be easier to maintain.  Feedback welcome.

Issue #1007, opened 2022-08-17
MacOS: Block & Remember this decision leave hanging Window


tresf:
On MacOS, if you visit https://demo.qz.io and click "Block" and "Remember this decision", the "Are you sure" dialog appears below the "Allow".  Then, once you click "Yes", the "Allow" hangs forever and there's no way to close it.

The bug occurs on MacOS Ventura with QZ Tray 2.2.1.  I'm not sure if this bug exists on other MacOS and QZ Tray versions.

Vzor-:
This issue did not present on 12.5.1 Monterey

Vzor-:
This issue did present on M1 ventura 13.0

tresf:
Minimal code to reproduce: https://gist.github.com/Vzor-/e23c5faa79218d12644d7fe664e3831e

Bug filed (private) with vendor `#QZ-8`.

tresf:
Workaround merged to master via https://github.com/qzind/tray/pull/1012.  This will be part of QZ Tray 2.2.2 and higher.

tresf:
Workaround is being reverted.  The latest macOS Ventura beta fixes this upstream, so we don't need the workaround.

tresf:
Workaround reverted via f24f80b.  Closing.

PR #1012, opened 2022-08-25
Gateway fix

Vzor-:
A temporary workaround for a race condition reported in #1007.
A demonstration and 'fix' can be seen [here](https://gist.github.com/Vzor-/e23c5faa79218d12644d7fe664e3831e).


tresf:
@Vzor- elegantly simply.  Thanks, we'll add this to 2.2.2+ until we can hear back from our JDK support provider.

PR #1016, opened 2022-08-31
Dialog Location Validation

### LINKED ISSUE #1013: Manually sending JSON breaks UI
~~Two~~ issues:

~~1. On Windows 11, certain JSON payloads, when sent to QZ Tray out of order will cause UI anomalies, such as a blank window.~~ (CANNOT REPRODUCE, POSSIBLE DUPLICATE OF #823)
 

2. By modifying the dialog's `x` and `y` positions, you can place the dialog non-centered, which can make it harder to find (e.g. upper right corner, etc).
   ```js
   var obj =  {"call":"printers.find","timestamp":1661263106002,"uid":"5w9ad2","position":{"x":-1,"y":-1}}
   _qz.websocket.connection.send(JSON.stringify(obj));
   ```

Vzor-:
Added a check to verify a dialog is entirely inside of the screen. If the dialog is not, it will be centered on the primary screen as per the old logic. This logic supports multiple monitors on windows.

Fixes #1013 

Tests for @Vzor- : 
- Does this work with multiple screens on other OSs
- Does this work with multiple monitors in other modes such as duplicate

### DIFF
--- a/src/qz/utils/SystemUtilities.java
+++ b/src/qz/utils/SystemUtilities.java
@@ -469,8 +469,9 @@ public class SystemUtilities {
      * @return <code>true</code> if the operation is successful
      */
     public static void centerDialog(Dialog dialog, Point position) {
-        if (position == null) {
-            log.debug("No dialog position provided, we'll center on the primary monitor instead");
+        // Assume 0,0 are bad coordinates
+        if (position == null || (position.getX() == 0 && position.getY() == 0)) {
+            log.debug("Invalid dialog position provided: {}, we'll center on first monitor instead", position);
             dialog.setLocationRelativeTo(null);
             return;
         }
--- a/src/qz/utils/SystemUtilities.java
+++ b/src/qz/utils/SystemUtilities.java
@@ -469,8 +469,8 @@ public class SystemUtilities {
      * @return <code>true</code> if the operation is successful
      */
     public static void centerDialog(Dialog dialog, Point position) {
-        if (position == null || position.getX() == 0 || position.getY() == 0) {
-            log.debug("Invalid dialog position provided: {}, we'll center on the primary monitor instead", position);
+        if (position == null) {
+            log.debug("No dialog position provided, we'll center on the primary monitor instead");
             dialog.setLocationRelativeTo(null);
             return;
         }
@@ -484,16 +484,17 @@ public class SystemUtilities {
         }
 
         Rectangle rect = new Rectangle((int)(position.x * dpiScale), (int)(position.y * dpiScale), dialog.getWidth(), dialog.getHeight());
+        rect.translate(-dialog.getWidth() / 2, -dialog.getHeight() / 2);
         Point p = new Point((int)rect.getCenterX(), (int)rect.getCenterY());
         log.debug("Calculated dialog centered at: {}", p);
 
         if (!isWindowLocationValid(rect)) {
-            log.debug("Dialog position provided is out of bounds: {}, we'll center on the primary monitor instead", position);
+            log.debug("Dialog position provided is out of bounds: {}, we'll center on the primary monitor instead", p);
             dialog.setLocationRelativeTo(null);
             return;
         }
 
-        dialog.setLocation(p);
+        dialog.setLocation(rect.getLocation());
     }
 
     public static boolean isWindowLocationValid(Rectangle window) {
--- a/src/qz/utils/SystemUtilities.java
+++ b/src/qz/utils/SystemUtilities.java
@@ -498,11 +498,18 @@ public class SystemUtilities {
         dialog.setLocation(rect.getLocation());
     }
 
+    /**
+     * Validates if a given rectangle is within screen bounds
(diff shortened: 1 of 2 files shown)


src/qz/utils/SystemUtilities.java
```diff
@@ -468,26 +469,43 @@ public static boolean setSystemLookAndFeel() {
      * @return <code>true</code> if the operation is successful
      */
     public static void centerDialog(Dialog dialog, Point position) {
-        if (position == null || position.getX() == 0 || position.getY() == 0) {
```
Vzor-, line 471: I assumed the `getXY() != 0` check is now redundant. Was there another purpose for these checks?

src/qz/utils/SystemUtilities.java
```diff
@@ -468,26 +469,43 @@ public static boolean setSystemLookAndFeel() {
...
+
+        if (!isWindowLocationValid(rect)) {
+            log.debug("Dialog position provided is out of bounds: {}, we'll center on the primary monitor instead", p);
+            dialog.setLocationRelativeTo(null);
+            return;
+        }
+
+        dialog.setLocation(rect.getLocation());
+    }
+
```
Vzor-, line 499: Forgot to add a doc for this function. I'll do that

src/qz/utils/SystemUtilities.java
tresf: It's not the best logic in the world, but we're actually using `0,0` as a very unlikely magic number per https://github.com/qzind/tray/blob/a07a96b54691a215e9f93aaff4fcee7a0c5ef436/js/qz-tray.js#L233-L236.

tresf: Despite it's obvious flaws, I would recommend keeping the `0,0` guard, otherwise, the logs will read that it's out of bounds.  One could argue that with the right monitor setup, it may be possible to have the center of the screen at `0,0`, but at the time Brett and I chose this because the likelihood was very low and it allowed useful logging.

In hindsight, we probably should have just sent `{ position: null }` (or better, [not send it at all](https://github.com/qzind/tray/blob/32192244f9d3051abc3b4efd1c9abf9170129674/src/qz/ws/PrintSocketClient.java#L137)) but reverting this decision would need regression testing in both directions for API and app.

tresf: It may also be the case that the browsers which didn't support this (e.g. IE and the [old Chakra (pre-Chromium) powered Edge](https://en.wikipedia.org/wiki/Chakra_(JScript_engine) and would return a zero as a safeguard.  Unfortunately whatever discussions Brett and I had at that time are lost as they don't appear to be in the source code, Pull Request or bug report.

tresf: I tested this on Legacy Edge and IE11 and as it turns out, they send null due to a bug in the JavaScript screen calculation.  This is fixed in 7ead436 and now they too will center the dialog on the correct screen.   I also edited the `0, 0` guard back into the code, since the JSON parsing creates a `Point(0, 0)` and then uses `move(...)`, so additional regression testing would be needed if we wanted to eliminate the magic numbers altogether.

https://github.com/qzind/tray/blob/32192244f9d3051abc3b4efd1c9abf9170129674/src/qz/ws/PrintSocketClient.java#L669-L673

src/qz/utils/SystemUtilities.java
tresf: Consider keeping the point calculation, moving the window, then calculating the bounds with the rectangle logic.  The advantage of this approach is that we can write a generic on-screen-bounds calculation, e.g... 

```java
   /**
     * Validates if a given component is within screen bounds
     */
    public static boolean isComponentOnScreen(Component c) {
        // NOT YET TESTED
        Rectangle bounds = new Rectangle(c.getX(), c.getY(), c.getWidth(), c.getHeight());

        GraphicsDevice[] devices = GraphicsEnvironment.getLocalGraphicsEnvironment().getScreenDevices();
        Area area = new Area();
        for (GraphicsDevice gd : devices) {
            for (GraphicsConfiguration gc :  gd.getConfigurations()) {
                area.add(new Area(gc.getBounds()));
            }
        }
        return area.contains(bounds);
    }
```

Vzor-: getLocationOnScreen
Throws:
    IllegalComponentStateException - if the component is not showing on the screen

looks like we may need to find another way

tresf: Hmm.... Perhaps just `getLocation()` then?

tresf: Since we don't move the component until AFTER we calculate its location, it's probably best to leave it as-is for now.

src/qz/utils/SystemUtilities.java
tresf: Leaving the magic numbers 0,0 for now.  We can revisit at a later time if needed.

Issue #1017, opened 2022-09-02
Printer status: "Printer state-reason "media-low-report" was not found" - leaves user confused as to print job status


lite1979:
Writing this up as a user recently contacted us regarding a failed print. 

While the failed print is most likely the result of PC or printer settings, perhaps a better message is available when the status call is less informative like this one.

tresf:
@Vzor- is this a missing mapping?

tresf:
@Vzor- here's some additional information:

https://github.com/OpenPrinting/cups/blob/1da735b3f0cf258eba0836b54796dc99c4950ac3/backend/snmp-supplies.c#L161-L176

tresf:
So it appears to be some SNMP messaging that doesn't exactly match that from within CUPS... 

https://github.com/qzind/tray/blob/875c7038cf18a4ae8ac440627f13230aeddaecee/src/qz/printer/status/printer/CupsPrinterStatusMap.java#L39

For example, QZ Tray looks for:

| QZ Looks for | SNMP reports |
|-------|------------|
| `media-low` | `media-low-report` | 
| `media-empty` | `media-empty-warning` | 
| `toner-low` | `toner-low-report` | 
| `toner-empty` | `toner-empty-warning` |
| `door-open` | `door-open-report` | 
| `media-jam` | `media-jam-warning` |
| `offline-report` |  :white_check_mark: `offline-report` |
| | `service-needed-warning` |
| `input-tray-missing` | `input-tray-missing-warning` |
| `output-tray-missing` | `output-tray-missing-warning` |
| `marker-supply-missing` | `marker-supply-missing-warning` |
| `output-area-almost-full` | `output-area-almost-full-report` |
| `output-area-full` | `output-area-full-warning` |

@Vzor- I feel like we could populate the sorted reason table with the additional `-warning` and `-report` strings rather quickly and handle any/all of these.  Alternately, we could truncate `-report` or `-warning` when the message arrives, taking special consideration for the ones that actually end with that suffix.

tresf:
@Vzor- please provide an update on this.

Vzor-:
Pr soon. I took the truncation approach. It should be a lot easier for the end user that way. I'd hate for someone to filter for toner-empty only to have a toner-empty-warning slip through.

Vzor-:
Some minor consequences of this fix, 
-If there is an unknown status in the future that ends in report or warning, it will be truncated. exceptions like offline-report are hard-coded.
[...]

PR #1026, opened 2022-09-28
Status fix for SNMP

Vzor-:
Fixes SNMP status mismatch with cups

see #1017

### DIFF
--- a/src/qz/printer/status/CupsStatusHandler.java
+++ b/src/qz/printer/status/CupsStatusHandler.java
@@ -116,7 +116,7 @@ public class CupsStatusHandler extends AbstractHandler {
 
     // fix for #1017
     // Removes redundant '-warning' '-report' from SNMP originated statuses, with some exceptions
-    public static String statusFix(String cupsString) {
+    private static String statusFix(String cupsString) {
         ArrayList exceptions = new ArrayList(Arrays.asList("offline-report", "cups-insecure-filter-warning", "cups-missing-filter-warning"));
         String[] badEnding = {"-warning","-report"};
         if (exceptions.contains(cupsString)) return cupsString;
--- a/src/qz/printer/status/CupsStatusHandler.java
+++ b/src/qz/printer/status/CupsStatusHandler.java
@@ -12,7 +12,6 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 
 /**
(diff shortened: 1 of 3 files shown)


src/qz/printer/status/CupsStatusHandler.java
```diff
@@ -64,6 +65,7 @@ private synchronized void getNotifications() {
                 int attrCount = cups.ippGetCount(jobStateReasonsAttr);
                 for (int i = 0;  i < attrCount; i++) {
                     String reason = cups.ippGetString(jobStateReasonsAttr, i, "");
+                    reason = statusFix(reason);
```
tresf, line 68: Why not just do this in the map?

Vzor-: I wanted to apply the fix as close to the cups side of things a possible. That way the status will already be changed for the duplicate detection.

tresf: The map's job is to parse these reasons into enumerated types.  I seems best there, no (and it's only one line lower 😉 )

src/qz/printer/status/printer/CupsPrinterStatusMap.java
```diff
@@ -68,6 +66,9 @@ public enum CupsPrinterStatusMap implements NativeStatus.NativeMap {
     CUPS_INSECURE_FILTER(REASON, NativePrinterStatus.SERVER_UNKNOWN), // "cups-insecure-filter"
     CUPS_MISSING_FILTER(REASON, NativePrinterStatus.ERROR), // "cups-missing-filter"
 
+    // SNMP warning with no existing CUPS definition
```
tresf, line 69: Minor, but this should be `// SNMP statuses with no existing CUPS definition`

src/qz/printer/status/printer/CupsPrinterStatusMap.java
```diff
@@ -918,4 +921,15 @@ public NativePrinterStatus getParent() {
     public Object getRawCode() {
         return name().toLowerCase(Locale.ENGLISH).replace("_", "-");
     }
+
+    // Bugfix for issue #1017
+    // Removes redundant '-warning' '-report' from SNMP originated statuses, with some exceptions
+    private static String statusFix(String cupsString) {
+        ArrayList exceptions = new ArrayList(Arrays.asList("offline-report", "cups-insecure-filter-warning", "cups-missing-filter-warning"));
```
tresf, line 928: Please convert exceptions from the enums.

tresf: ... arguably, you could programatically create this list too.  I'm going to take a swing at a small refactor and send what I come up with.

tresf:
@Vzor- can you please do some basic testing so that we can get this merged?

src/qz/printer/status/printer/CupsPrinterStatusMap.java
```diff
@@ -918,6 +908,32 @@ public static NativePrinterStatus matchState(String state) {
         return sortedStateLookupTable.getOrDefault(state, NativePrinterStatus.UNMAPPED);
     }
 
+    public static Status createStatus(String reason, String state, String printer) {
+        NativePrinterStatus cupsPrinterStatus = matchReason(reason);
+
```
tresf, line 913: The comment was lost in this refactor: `// Edge-case for snmp statuses`

PR #1031, opened 2022-10-06
Os versioning update

Vzor-:
Fixes out of date versioning structure for windows 10+
adds more detailed version string to security info for all os


src/qz/utils/WindowsUtilities.java
```diff
@@ -64,12 +66,27 @@ public static boolean isDarkTaskbar() {
...
         try {
             String productName = WindowsUtilities.getRegString(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "ProductName");
-            int major = WindowsUtilities.getRegInt(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "CurrentMajorVersionNumber");
-            int minor = WindowsUtilities.getRegInt(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "CurrentMinorVersionNumber");
-            String build = WindowsUtilities.getRegString(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "CurrentBuild");
-            int ubr = WindowsUtilities.getRegInt(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "UBR");
-            return String.format("%s %d.%d.%s.%d", productName, major, minor, build, ubr);
-        } catch(Exception ignore) {}
+            WinNT.OSVERSIONINFO versionInfo = new WinNT.OSVERSIONINFO();
+            if (!Kernel32.INSTANCE.GetVersionEx(versionInfo)) throw new RuntimeException();
```
tresf, line 70: Was the intention to throw an empty exception here?

Vzor-: Yea that function can return false, which results in a null versionInfo. In that case we want to return before we run into any null pointer exceptions. I should probably include an explanation in the exception. The class is deprecated so it isn't an impossible null scenario, and that check was suggested in some example code.

tresf: Sorry, I meant, "Was the intention to throw an EMPTY exception here?" 😄

tresf: >  should probably include an explanation in the exception.

☝️

PR #1042, opened 2022-11-11
Add graceful renewal for shared directory too

### LINKED ISSUE #916: Graceful renewal of certificate not working for SHARED directory
When we read the renewal information for graceful handling of certificates, we fail to check in the SHARED directory, e.g. `%ProgramData%\qz\allowed.dat`.

A patch should be back ported to 2.1 branch.

https://github.com/qzind/tray/blob/2a7c6d44b1ebb5db32f995a6d8c81917cbfcb250/src/qz/auth/Certificate.java#L272-L296

Related #719 #43

tresf:
Closed #916

### DIFF
--- a/src/qz/auth/Certificate.java
+++ b/src/qz/auth/Certificate.java
@@ -288,9 +288,20 @@ public class Certificate {
             }
 
             // Add this certificate to the whitelist if the previous certificate was whitelisted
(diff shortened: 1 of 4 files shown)


tresf:
Backported to 2.1 via ecc20ec.  Note, 2.1 is end-of-life, so the 2.1 fix will only make it to company branded builds.

Issue #1052, opened 2022-12-06
LogDialog cannot display UTF-8


tresf:
The `LogDialog` cannot print UTF-8 characters.  It appears to be a bad assumption that characters are one-byte long.

Problematic code:
https://github.com/qzind/tray/blob/875c7038cf18a4ae8ac440627f13230aeddaecee/src/qz/ui/LogDialog.java#L71-L75

tresf:
Closed via #1059

PR #1059, opened 2022-12-16
Fix LogDialog encoding

bberenz:
for #1052, utf8 support in logdialog ui


tresf:
Mac:

image

Ubuntu:

image

Windows:

image

Issue #1070, opened 2023-01-10
Remove deprecated reference to escape()/unescape()


tresf:
@bberenz any thoughts on how we can remove this?

https://github.com/qzind/tray/blob/b6849b20b770e29b618732c317ad4032008a90b5/js/qz-tray.js#L27-L36

tresf:
Closed via #1073

PR #1073, opened 2023-01-12
Remove deprecated calls

bberenz:
For #1070 

So looks like the `utf8Decode` is never used, so that's been dropped.
As for the `utf8Encode` my understanding was that unescape was needed to counteract some of the [encoding](https://stackoverflow.com/questions/26342123/replacement-for-javascript-escape) the encodeURI* calls do to remain consistent with how java handles those. But going off the utf8decode call I think we just pulled these as-is from the sha library, since testing with unescape removed seems to work fine when special characters are used. 
I tend to be a bit limited on testing signing cases though, so some additional checks before merging this in would be appreciated.

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -26,13 +26,7 @@ var qz = (function() {
 
     // from SHA implementation
     if (typeof String.prototype.utf8Encode == 'undefined') {
-        String.prototype.utf8Encode = function() { return unescape(encodeURIComponent(this)); };
-    }
-    if (typeof String.prototype.utf8Decode == 'undefined') {
-        String.prototype.utf8Decode = function() {
-            try { return decodeURIComponent(escape(this)); }
-            catch(e) { return this; } // invalid UTF-8? return as-is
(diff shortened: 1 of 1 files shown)


tresf:
> 

This fails, signing completely breaks.

```js
"\x02".utf8Encode()
# Before: '\x02'
# After: '%02'
```

tresf:
Ok, the above test is a bad one.... here's a better one... 

```js
unescape(encodeURIComponent("こんにちわ")); // konnichiwa
// returns: 'ã\x81\x93ã\x82\x93ã\x81«ã\x81¡ã\x82\x8F'
```

tresf:
Our existing signing logic seems to expect the data in the odd `unescape(...)` format, so I've scoured the web to find a 12 year old [blog](https://cwestblog.com/2011/05/23/escape-unescape-deprecated/) about a polyfill which -- when combined with the comments -- have a version which seems to be on parity with the browser.  Implemented via 966a627.

PR #1080, opened 2023-01-20
Add native macOS Security support

### LINKED ISSUE #1076: macOS silent deployment is not silent
The cause is best described here: https://github.com/actions/runner-images/issues/4519, certificate installation requires intervention.

Related: https://github.com/qzind/tray/pull/1056

tresf:
### Goal

Create a `MacCertificateInstaller` similar to [`WindowsCertificateInstaller`](https://github.com/qzind/tray/blob/60e4c8c87d952963aec18c63ae182ed0533ae285/src/qz/installer/certificate/WindowsCertificateInstaller.java) that leverages JNA and native system libraries for interacting with the computer certificate store.

### Milestones

* [ ] Add certificate (started)
* [ ] Remove certificate
* [ ] Verify certificate

Fixes #1076

### DIFF
--- a/src/org/dyorgio/jna/platform/mac/Security.java
+++ b/src/org/dyorgio/jna/platform/mac/Security.java
@@ -21,7 +21,6 @@ import com.sun.jna.Pointer;
 import com.sun.jna.platform.mac.CoreFoundation;
 import com.sun.jna.ptr.IntByReference;
 import com.sun.jna.ptr.PointerByReference;
-import qz.printer.status.Cups;
 
 /**
  * The JNA library interface to access the Security library under MacOS.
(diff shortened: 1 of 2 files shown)


Vzor-:
> This has some improvements that -- I believe -- are unrelated to this task and would make the PR easier to review if they were split out (e.g. `removeAppLauncher`, `removeStartupEntry`). @Vzor- can you please split this PR into two PRs so that we can take advantage of the refactor?
> 
> With regards to the Mac security stuff, I think it's still a good idea to add this if the efforts are fruitful. I understand that merging the Mac security stuff does not close #1076, however I believe the approach of using MacOS native libraries to be more professional approach at installing our certificates.

Agreed. It may take a couple of days until I'm back on this issue.

PR #1092, opened 2023-01-26
Keep only a single keep alive interval on websocket

### LINKED ISSUE #1081: Cleanup pings on reconnect
This email says it best... 

> On connection to QZTray a websocket is established and (if configured) will send a ping in a set interval as intended, however if the connection is disconnected and then subsequently reconnected there will now be an additional pinging interval on the websocket (And will continuously add another on each reconnect).
> 
> This issue causes unwanted processing and network traffic that can potentially lead to performance issues depending on how many times a client's connection is interrupted (For example, a user closing their laptop).
> 
> The issue can be replicated and observed on the demo site (https://demo.qz.io/) by disconnecting and reconnecting while recording the websocket network traffic.
> 
> For example, with an expected 1 minute interval ping:
> SingleMinutePing

> After multiple reconnections:
> MultipleRepeatingPing

> 
> The code in question is never stopped when the connection is closed:
> https://github.com/qzind/tray/blob/60e4c8c87d952963aec18c63ae182ed0533ae285/js/qz-tray.js#L139-L148
>

bberenz:
Fixes #1081

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -138,13 +138,15 @@ var qz = (function() {
 
                                 if (config.keepAlive > 0) {
                                     var interval = setInterval(function() {
-                                        if (!_qz.tools.isActive()) {
+                                        if (!_qz.tools.isActive() || _qz.websocket.connection.interval !== interval) {
                                             clearInterval(interval);
                                             return;
                                         }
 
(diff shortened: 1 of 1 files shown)


tresf:
The OP from #1081 reports that this is tested and working.  Merging. :)

Issue #503, opened 2019-10-17
PDF to Zebra printer is blurry and/or poor quality


lite1979:
Using 2.1 RC7, Windows 10, Zebra GK420d printer, and compared to Adobe results, QZ-Tray's print results are poor by comparison. 

Printing to  "Microsoft Print to PDF", the resulting PDF is not blurry, so I suspect either the Zebra driver or the font associated with the print job vs. printer capability in this case. 

[Drug-print.pdf](https://github.com/qzind/tray/files/3756942/Drug-print.pdf)

tresf:
Here are the details of my testing:
* The PDF is confirmed to be printing vector (`rasterize: false` is being honored), but the font is curtailing at such a small size. Making it bold will help slightly.  I recommend the label is redesigned so that it prints better.  I feel there's room for improvement in regards to print quality, but it would be an upstream issue for PDFBOX.
* I've tested turning off dithering in the driver settings and it did not seem to help this particular problem.
* I've tested the PDF on the latest release and it has the same results.

In summary what worked best (still not the same quality as Adobe) was setting the following in the ZDesigner driver:
* Darkness: 15
* Speed: 1.5/s

If we decide to reopen this, it will need to be filed upstream.  The tests were done [using PDFBOX on command line](https://pdfbox.apache.org/2.0/commandline.html) so QZ Tray is not a factor.

Closing as `wontfix` for now.  We may have to revisit this.

tresf:
We'll take a deeper dive into this.  Assigning to @bberenz and @lite1979.

@bberenz can you provide Lite with a unit test that does the following:
1. Prints something like a JLabel text, optimized for thermal printing (aliasing off)
2. Prints a PDF text of same size, optimized for thermal printing

Then we'll raise the concerns to PDFBOX.

lite1979:
https://issues.apache.org/jira/browse/PDFBOX-4709

lite1979:
Unfortunately, this issue has been labeled as "won't fix" by Apache, and my bug was labeled as a duplicate of this one:

https://issues.apache.org/jira/browse/PDFBOX-3293

The recommendation is to use another font; the user that raised the issue for the link above lists his fonts as [NOTO CJK](https://www.google.com/get/noto/help/cjk/) and [CNS11643](http://data.gov.tw/node/5961)

tresf:
Per recommendation of PDFBOX, filed upstream bug report: https://github.com/AdoptOpenJDK/openjdk-support/issues/45

[...]

PR #1108, opened 2023-04-01
First draft of pdf font render improvements

### LINKED ISSUE #1128: Multi-page PDF only prints first page, only on Zebra, only on Windows
Using the same source PDF, the issue is also evident using PDFBox (tested both 2.0.27 and 3.0.x).

To further complicate things, when printing to another printer, such as a laser printer or a virtual printer, not only do both page print, but the _second page appears to be rasterized_

The PDF in question was created using Qt 4.8.7 via wkhtmltopdf 0.12.6 (which is unfortunately archived as of January of this year).

TODO: Open issue with PDFBox once a publicly shareable PDF is available to reproduce the problem.

tresf:
Addresses #503 
Closes #1128 

To toggle on:

```js
var config = qz.configs.create("Printer Name", { altFontRendering: true });
```

Opening as a placeholder.  This PR is outdated and as such the below "issues" may since be outdated.

Known issues:
- ~~snapshot build has custom edits to pagedrawer function visibility~~
- ~~calibri font is rendering with wrong glyphs~~


tresf:
> ✅  * calibri font is rendering with wrong glyphs

This seems to be merged into Java 11 since 11.0.13 https://github.com/adoptium/jdk11u/commit/5a7e7d3bd6dea041b58896a0d84d4b7eb3658684.

>  ✅  * snapshot build has custom edits to pagedrawer function visibility

Patched via https://issues.apache.org/jira/browse/PDFBOX-5093.

I'll rebase this PR on master so we can start to send out experimental builds.

tresf:
Rebase is done.  @bberenz since we originally wrote this feature, we've added our "ignore transparency" feature #952, so we now have two separate PDFRenderers.  Assuming we want to be able to take advantage of both, can you help combine these in a way that we can toggle either/or?

tresf:
I tried to combine the two PDF workaround classes myself and it resulted in a blank printout, I'm not sure what I'm doing wrong.  For now, the flags are mutually exclusive until this code can be refactored.

Unfortunately the proposed code seems to have a bug when added to QZ Tray.  This doesn't occur when using the same code standalone:

image

tresf:
@bberenz thanks, everything seems to be working now.

tresf:
Code has been added to this PR to also close #1128.

tresf:
Wiki updated: https://github.com/qzind/tray/wiki/Pixel#pdf-text-quality

PR #1117, opened 2023-04-19
Allow disconnect call to cancel current connection attempt

### LINKED ISSUE #1111: qz.websocket.disconnect should allow abruptly interrupting any active connection requests
Quoting:

> There's a small bug I want to draw your attention to:
> 
> The issue occurs when we try to connect to a QZ Tray app on a custom IP address that doesn't exist. In this case websocket is hanging on pending state and there's no way to close the connection until it times out. There's also no way to create a new connection to another IP address either. So we just have to wait or refresh the browser page. 
> 
> ```js
> qz.websocket.disconnect()  // => Error: No open connection with QZ Tray
> qz.websocket.connect()     // => The current connection attempt has not returned yet

akberenz:
closes #1111

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -52,6 +52,8 @@ var qz = (function() {
         websocket: {
             /** The actual websocket object managing the connection. */
             connection: null,
+            /** Track if a connection attempt is being cancelled. */
+            shutdown: false,
 
             /** Default parameters used on new connections. Override values using options parameter on {@link qz.websocket.connect}. */
             connectConfig: {
@@ -75,6 +77,11 @@ var qz = (function() {
             setup: {
                 /** Loop through possible ports to open connection, sets web socket calls that will settle the promise. */
                 findConnection: function(config, resolve, reject) {
+                    if (_qz.websocket.shutdown) {
+                        reject(new Error("Connection attempt cancelled by user"));
+                        return;
+                    }
+
                     //force flag if missing ports
                     if (!config.port.secure.length) {
                         if (!config.port.insecure.length) {
@@ -758,7 +765,9 @@ var qz = (function() {
             },
 
             isActive: function() {
-                return _qz.websocket.connection != null && _qz.websocket.connection.established;
+                return _qz.websocket.connection != null
+                    && (_qz.websocket.connection.readyState === WebSocket.OPEN
+                        || _qz.websocket.connection.readyState === WebSocket.CONNECTING);
             },
 
             assertActive: function() {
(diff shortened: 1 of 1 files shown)


js/qz-tray.js
```diff
@@ -758,7 +765,9 @@ var qz = (function() {
             },
 
             isActive: function() {
-                return _qz.websocket.connection != null && _qz.websocket.connection.established;
+                return _qz.websocket.connection != null
+                    && (_qz.websocket.connection.readyState === WebSocket.OPEN
+                        || _qz.websocket.connection.readyState === WebSocket.CONNECTING);
```
tresf, line 770:
```suggestion
                    && (_qz.websocket.connection.readyState === _qz.tools.ws.OPEN
                        || _qz.websocket.connection.readyState === _qz.tools.ws.CONNECTING);
```

js/qz-tray.js
```diff
@@ -1100,12 +1109,19 @@ var qz = (function() {
...
+
+                        if (state === WebSocket.OPEN) {
+                            reject(new Error("An open connection with QZ Tray already exists"));
+                            return;
+                        } else if (state === WebSocket.CONNECTING) {
+                            reject(new Error("The current connection attempt has not returned yet"));
+                            return;
+                        } else if (state === WebSocket.CLOSING) {
+                            reject(new Error("Waiting for previous disconnect request to complete"));
+                            return;
```
tresf, line 1123:
```suggestion
                        if (state === _qz.tools.ws.OPEN) {
                            reject(new Error("An open connection with QZ Tray already exists"));
                            return;
                        } else if (state === _qz.tools.ws.CONNECTING) {
                            reject(new Error("The current connection attempt has not returned yet"));
                            return;
                        } else if (state === _qz.tools.ws.CLOSING) {
                            reject(new Error("Waiting for previous disconnect request to complete"));
                            return;
```

tresf:
@bberenz with this new change, the `sample.html` seems to no longer get into `.then()` after disconnect is called.

https://github.com/qzind/tray/blob/49190abd3d3a05487c901ae84dbd76c458b4c7ab/sample.html#L1510-L1512

image

tresf:
Another issue:

Clicking "Disconnect" twice causes the following error:

```
sample.html:2921 TypeError: Cannot set properties of null (setting 'promise')
    at qz-tray.js:1196:62
    at new Promise (<anonymous>)
    at Object.promise (qz-tray.js:616:28)
    at Object.disconnect (qz-tray.js:1191:34)
    at endConnection (sample.html:1510:26)
    at HTMLButtonElement.onclick (sample.html:69:102)
```

bberenz:
Can't replicate either of your disconnect issues. I added an additional check to the `isActive` call that should take care of the double call you saw. But it always seems to get into the `.then` call just fine for me, can you provide steps?

tresf:
> I added an additional check to the `isActive` call that should take care of the double call you saw. 

Double call is fixed, thanks!

> Can't replicate either of your disconnect issues. But it always seems to get into the `.then` call just fine for me, can you provide steps?

In Edge or Chrome on Windows 11 or MacOS 13:

1. Open `sample.html` from local filesystem
2. Click "Disconnect" to cancel the `localhost` connection
3. Click the dropdown arrow next to the "Connect" button
4. Type `yahoo.com`
5. Click "Connect"
6. Click "Disconnect"
7. Web page is stuck in "Waiting"

Note, the disconnect does eventually process, but it takes minutes.

tresf:
@akberenz I did some more testing and can confirm that the code snippet from #1111 works:

```js
var ws = new WebSocket('ws://unknownhost.local');
setTimeout(() => {
  if (ws.readyState == WebSocket.CONNECTING) {
    ws.close();
    console.log("Connection cancelled.");
  }
}, 1000);
```

... while ours doesn't.  For some reason `ws.readyState` is never properly switching to `CLOSED` (or perhaps the WebSocket is attempting a subsequent connection after we've called `.disconnect()`?).

Can you help investigate why this behavior is different?  In its current form, this PR still doesn't work in Chromium-based browsers.

PR #1124, opened 2023-05-01
Printer Status: Performance improvements

### LINKED ISSUE #1116: MacOS: Unusually high CPU when listening for printer events
**QZ Version:** 2.2.2
**OS:** MacOS Catalina 10.15.7
**Printer:** Zebra ZD620 ZPL

Printing a base64 PDF to a USB printer while the status listener is running causes significant CPU resources (e.g. 80% or higher).  Eventually these settle down, however depending on the number of documents, it can block the websocket connections, requiring killing the QZ Tray process.

Workaround:  Disable printer status listener
Permanent fix: (TBD)

### LINKED ISSUE #1189: printers.putAll(...) clears NativePrinterMap
As identified in #1124 by @Vzor-, `printers.putAll(...)` will wipe out all entries of `NativePrinterMap`.

https://github.com/qzind/tray/blob/402f9243b256b9c30e8206c1e8e18bce1ff809bf/src/qz/printer/PrintServiceMatcher.java#L54-L57

This is because `NativePrinterMap.findMissing(...)` performs a destructive operation:

https://github.com/qzind/tray/blob/402f9243b256b9c30e8206c1e8e18bce1ff809bf/src/qz/printer/info/NativePrinterMap.java#L46-L64

tresf:
Introduce `PrintService` caching to workaround https://bugs.openjdk.org/browse/JDK-7001133

Upstream: [`JDK-7001133`](https://bugs.openjdk.org/browse/JDK-7001133)
BellSoft: `QZ-14` (private)

Related: #1181 
Closes #1116 
Closes #1189

### DIFF
--- a/src/qz/common/CachedObject.java
+++ b/src/qz/common/CachedObject.java
@@ -1,0 +1,96 @@
+package qz.common;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+
+/**
+ * A generic class that encapsulates an object for caching. The cached object
+ * will be refreshed automatically when accessed after its lifespan has expired.
+ *
+ * @param <T> The type of object to be cached.
+ */
+public class CachedObject<T> {
+    public static final long DEFAULT_LIFESPAN = 5000; // in milliseconds
+    T lastObject;
+    Supplier<T> supplier;
+    private long timestamp;
+    private long lifespan;
+
+    /**
+     * Creates a new CachedObject with a default lifespan of 5000 milliseconds
+     *
+     * @param supplier The function to pull new values from
+     */
+    public CachedObject(Supplier<T> supplier) {
+        this(supplier, DEFAULT_LIFESPAN);
+    }
+
+    /**
+     * Creates a new CachedObject
+     *
+     * @param supplier The function to pull new values from
+     * @param lifespan The lifespan of the cached object in milliseconds
+     */
+    public CachedObject(Supplier<T> supplier, long lifespan) {
+        this.supplier = supplier;
+        setLifespan(lifespan);
+        timestamp = Long.MIN_VALUE;  // System.nanoTime() can be negative, MIN_VALUE guarantees a first-run.
+    }
+
+    /**
+     * Registers a new supplier for the CachedObject
+     *
+     * @param supplier The function to pull new values from
+     */
+    public void registerSupplier(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    /**
+     * Sets the lifespan of the cached object
+     *
+     * @param milliseconds The lifespan of the cached object in milliseconds
+     */
+    public void setLifespan(long milliseconds) {
+        lifespan = Math.max(0, milliseconds); // prevent overflow
+    }
+
+    /**
+     * Retrieves the cached object.
+     * If the cached object's lifespan has expired, it gets refreshed before being returned.
+     *
+     * @return The cached object
+     */
+    public T get() {
+        return get(false);
+    }
+
+    /**
+     * Retrieves the cached object.
+     * If the cached object's lifespan is expired or forceRefresh is true, it gets refreshed before being returned.
+     *
+     * @param forceRefresh If true, the cached object will be refreshed before being returned regardless of its lifespan
+     * @return The cached object
+     */
+    public T get(boolean forceRefresh) {
+        long now = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
+        // check lifespan
+        if (forceRefresh || (timestamp + lifespan <= now)) {
+            timestamp = now;
+            lastObject = supplier.get();
+        }
+        return lastObject;
+    }
+
+    // Test
+    public static void main(String ... args) throws InterruptedException {
+        final AtomicInteger testInt = new AtomicInteger(0);
+
+        CachedObject<Integer> cachedString = new CachedObject<>(testInt::incrementAndGet);
+        for(int i = 0; i < 100; i++) {
+            Thread.sleep(1500);
+            System.out.println(cachedString.get());
+        }
+    }
+}
--- a/src/qz/common/CachedObject.java
+++ b/src/qz/common/CachedObject.java
@@ -43,6 +43,7 @@ public class CachedObject<T> {
      *
      * @param supplier The function to pull new values from
      */
+    @SuppressWarnings("unused")
     public void registerSupplier(Supplier<T> supplier) {
         this.supplier = supplier;
     }
--- a/src/qz/printer/PrintServiceMatcher.java
+++ b/src/qz/printer/PrintServiceMatcher.java
@@ -29,8 +29,10 @@ import java.util.*;
 
 public class PrintServiceMatcher {
     private static final Logger log = LogManager.getLogger(PrintServiceMatcher.class);
-    //todo: include jdk version test for caching when JDK-7001133 is resolved
-    private static final boolean useCache = SystemUtilities.isMac();   // PrintService is slow, use a cache instead per JDK-7001133
+
+    // PrintService is slow in CUPS, use a cache instead per JDK-7001133
+    // TODO: Include JDK version test for caching when JDK-7001133 is fixed upstream
+    private static final boolean useCache = !SystemUtilities.isWindows();
 
     public static NativePrinterMap getNativePrinterList(boolean silent, boolean withAttributes) {
         NativePrinterMap printers = NativePrinterMap.getInstance();
@@ -67,7 +69,7 @@ public class PrintServiceMatcher {
 
         NativePrinterMap printers = NativePrinterMap.getInstance();
         if (!printers.contains(defaultService)) {
-            //todo: is this working correctly? it seems to set the printers list = to [1] {defaultPrinter}
+            // FIXME: Determine if this is this working correctly.  It seems to set the printers list = to [1] {defaultPrinter}
             printers.putAll(defaultService);
         }
 
--- a/src/qz/printer/PrintServiceMatcher.java
+++ b/src/qz/printer/PrintServiceMatcher.java
@@ -15,7 +15,6 @@ import org.apache.logging.log4j.Logger;
 import org.codehaus.jettison.json.JSONArray;
 import org.codehaus.jettison.json.JSONException;
 import org.codehaus.jettison.json.JSONObject;
-import qz.printer.info.CachedPrintService;
 import qz.printer.info.CachedPrintServiceLookup;
 import qz.printer.info.NativePrinter;
 import qz.printer.info.NativePrinterMap;
@@ -32,7 +31,7 @@ public class PrintServiceMatcher {
 
     // PrintService is slow in CUPS, use a cache instead per JDK-7001133
     // TODO: Include JDK version test for caching when JDK-7001133 is fixed upstream
-    private static final boolean useCache = !SystemUtilities.isWindows();
+    private static final boolean useCache = SystemUtilities.isUnix();
 
     public static NativePrinterMap getNativePrinterList(boolean silent, boolean withAttributes) {
         NativePrinterMap printers = NativePrinterMap.getInstance();
@@ -69,7 +68,6 @@ public class PrintServiceMatcher {
 
         NativePrinterMap printers = NativePrinterMap.getInstance();
         if (!printers.contains(defaultService)) {
-            // FIXME: Determine if this is this working correctly.  It seems to set the printers list = to [1] {defaultPrinter}
             printers.putAll(false, defaultService);
         }
 
--- a/src/qz/printer/PrintServiceMatcher.java
+++ b/src/qz/printer/PrintServiceMatcher.java
@@ -15,6 +15,7 @@ import org.apache.logging.log4j.Logger;
 import org.codehaus.jettison.json.JSONArray;
 import org.codehaus.jettison.json.JSONException;
 import org.codehaus.jettison.json.JSONObject;
+import qz.printer.info.CachedPrintService;
 import qz.printer.info.NativePrinter;
 import qz.printer.info.NativePrinterMap;
 import qz.utils.SystemUtilities;
@@ -27,13 +28,25 @@ import java.util.*;
 
 public class PrintServiceMatcher {
     private static final Logger log = LogManager.getLogger(PrintServiceMatcher.class);
+    //todo: include jdk version test for caching when JDK-7001133 is resolved
+    private static final boolean useCache = SystemUtilities.isMac();   // PrintService is slow, use a cache instead per JDK-7001133
 
     public static NativePrinterMap getNativePrinterList(boolean silent, boolean withAttributes) {
         NativePrinterMap printers = NativePrinterMap.getInstance();
-        printers.putAll(PrintServiceLookup.lookupPrintServices(null, null));
+        printers.putAll(lookupPrintServices());
         if (withAttributes) { printers.values().forEach(NativePrinter::getDriverAttributes); }
         if (!silent) { log.debug("Found {} printers", printers.size()); }
         return printers;
+    }
+
+    private static PrintService[] lookupPrintServices() {
+        if (useCache) return CachedPrintService.lookupPrintServices();
+        return  PrintServiceLookup.lookupPrintServices(null, null);
+    }
+
+    private static PrintService lookupDefaultPrintService() {
+        if (useCache) return CachedPrintService.lookupDefaultPrintService();
+        return  PrintServiceLookup.lookupDefaultPrintService();
     }
 
     public static NativePrinterMap getNativePrinterList(boolean silent) {
@@ -45,7 +58,7 @@ public class PrintServiceMatcher {
     }
 
     public static NativePrinter getDefaultPrinter() {
-        PrintService defaultService = PrintServiceLookup.lookupDefaultPrintService();
+        PrintService defaultService = lookupDefaultPrintService();
 
         if(defaultService == null) {
             return null;
@@ -53,6 +66,7 @@ public class PrintServiceMatcher {
 
         NativePrinterMap printers = NativePrinterMap.getInstance();
         if (!printers.contains(defaultService)) {
+            //todo: is this working correctly? it seems to set the printers list = to [1] {defaultPrinter}
             printers.putAll(defaultService);
         }
 
@@ -81,6 +95,8 @@ public class PrintServiceMatcher {
 
         if (!silent) { log.debug("Searching for PrintService matching {}", printerSearch); }
 
+        // Fix for https://github.com/qzind/tray/issues/931
+        // This is more than an optimization, removal will lead to a regression
         NativePrinter defaultPrinter = getDefaultPrinter();
         if (defaultPrinter != null && printerSearch.equals(defaultPrinter.getName())) {
             if (!silent) { log.debug("Matched default printer, skipping further search"); }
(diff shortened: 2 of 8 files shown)


Vzor-:
A lot of the slowdown with #1116 was due to printService.getName() which was reaching out to native and was called dozens of time for each status returned. I added a cached value that has a TTL of 10 seconds. I want to take a look at the various places getname is used to make sure this wont lead to any issues.

Vzor-:
Quick test on my system results in high CPU utilization for 4 seconds after 10 prints to a usb label printer. Prior it would take 1:30s with the same test. I think the actual print time improved too, though I don't have an easy way to measure that.

Vzor-:
Both `getName()` and `lookupDefaultPrintService()` are pretty expensive under certain circumstances. Statuses usually come in bundles of 5-10 events, depending on the type of printer. Each event calls `PrintServiceMatcher.matchPrinter()`, which calls `lookupDefaultPrintService` once and  `getName` n times, where n is the number of printers on the system. This means `getName` is called more often than `lookupDefaultPrintService`.

In a stand alone test, I made 100 calls to each and recorded the execution time.

|       | Mac        | Linux           | Windows  |
| ---|:-------------: |:-------------:| :-----:|
| printService      | 6000ms| 10000ms|75ms|
| getName   | 255ms      |   1ms |0ms|

I cached the values for both and tested the cached version on arm Window, Mac and Linux and tested how fast Tray could match 10 arbitrary printer names. On Mac, both helped. On linux, only defaultPrintService caching helped, and on windows none of this seemed to matter.

|       | Mac        | Linux           | Windows  |
| ---|:-------------: |:-------------:| :-----:|
| No caching      | 1350ms| 1500ms|25ms|
| Name caching     | 950ms      |   1500ms |25ms|
| Service caching   | 800ms    |    560ms|25ms|
| Both caching   | 410ms    |    550ms |30ms|

src/qz/printer/PrintServiceMatcher.java
```diff
@@ -27,9 +27,13 @@
 import javax.print.attribute.standard.PrinterName;
 import javax.print.attribute.standard.PrinterResolution;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
 
 public class PrintServiceMatcher {
     private static final Logger log = LogManager.getLogger(PrintServiceMatcher.class);
+    private static final long cachedDefaultServiceLifespan = TimeUnit.SECONDS.toNanos(10); //Set to 0 to disable, do not set to a negative number
```
tresf, line 34: Why a default of 10 nanoseconds specifically?  What's the reasoning behind this value?

Vzor-: That is actually 10 seconds, in nanoseconds. it could also be expressed as 10 
* 1000000000.

Vzor-:
FYI this last commit is 100% untested. I just needed to commit it.

src/qz/common/CachedObject.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.function.Supplier;
+
+/**
+ * A generic class that encapsulates an object for caching. The cached object
+ * will be refreshed automatically when accessed after its lifespan has expired.
+ *
+ * @param <T> The type of object to be cached.
+ */
+public class CachedObject<T> {
+    public static final long DEFAULT_LIFESPAN = 5000;
```
tresf, line 14: What is `5000` ?

src/qz/common/CachedObject.java
```diff
@@ -0,0 +1,97 @@
...
+    T cachedObject;
```
tresf, line 15: Consider rename `cachedObject` is not of type `CachedObject`, which is confusing.

src/qz/common/CachedObject.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Sets the lifespan of the cached object
+     *
+     * @param milliseconds The lifespan of the cached object in milliseconds
+     */
+    public void setLifespan(long milliseconds) {
+        // we never want lifespan to be negative, since timestamp initializes as long.MIN_VALUE
+        lifespan = Math.max(0, milliseconds);
+    }
```
tresf, line 58:
```suggestion
    public void setLifespan(long lifespan) {
        this.lifespan = Math.max(0, lifespan); // prevent overflow
    }
```

src/qz/common/CachedObject.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public CachedObject(Supplier<T> supplier) {
+        this(supplier, DEFAULT_LIFESPAN);
+    }
+
+    /**
+     * Creates a new CachedObject
+     *
+     * @param supplier The function to pull new values from
+     * @param lifespan The lifespan of the cached object in milliseconds
```
tresf, line 33: Hmm.... has this been tested?  You cannot safely compare or add `millis` and `nanos`.  If `nanos` is preferred (I think it should be), it should be converted.  1 millisecond = 1000000 nanoseconds, but to that point, why use nanos to begin with?

Relevant: https://stackoverflow.com/q/351565/3196753

src/qz/printer/info/NativePrinter.java
```diff
@@ -112,6 +118,10 @@ public PrinterProperty<String> getDriver() {
     }
 
     public String getName() {
+        return cachedName.get();
+    }
+
+    private String getNameNative() {
```
tresf, line 124: I don't think we should duplicate our functions just to satisfy a cache.

Perhaps... 

```java
    // PrintService.getName() is slow, use a cache instead per JDK-XXXXXXX
    // TODO: Remove this comment when upstream bug report is filed
    private CachedObject<String> cachedName = new CachedObject<>(printService.value::getName, lifespan);

    public String getName() {
        if (printService != null && printService.value() != null) {
            return cachedName.get();
        }
        return null;
    }

// REMOVE
//    private String getNameNative() {
//    ...    
//    }
```

src/qz/printer/PrintServiceMatcher.java
```diff
@@ -149,7 +157,8 @@ public static NativePrinter matchPrinter(String printerSearch, boolean silent) {
         return use;
     }
 
-    public static NativePrinter matchPrinter(String printerSearch) {
+    public static NativePrinter
+    matchPrinter(String printerSearch) {
```
tresf, line 161: Formatting mistake

src/qz/printer/PrintServiceMatcher.java
```diff
@@ -31,6 +32,11 @@
 public class PrintServiceMatcher {
     private static final Logger log = LogManager.getLogger(PrintServiceMatcher.class);
 
+    // lookupDefaultPrintService() reaches out to native and is expensive on mac/linux. This printService is cached, and refreshed if lifespan has elapsed
+    // todo Fix before merging: mention upsteam bug report
+    private static final long lifespan = SystemUtilities.isWindows() ? 0 : CachedObject.DEFAULT_LIFESPAN;
+    private static CachedObject<PrintService> cachedDefault = new CachedObject<>(PrintServiceLookup::lookupDefaultPrintService, lifespan);
+
```
tresf, line 39:
```suggestion
    // PrintServiceLookup.lookupDefaultPrintService() is slow, use a cache instead per JDK-XXXXXXX
    // TODO: Remove this comment when upstream bug report is filed
    private static final long lifespan = SystemUtilities.isWindows() ? 0 : CachedObject.DEFAULT_LIFESPAN;
    private static CachedObject<PrintService> cachedDefault = new CachedObject<>(PrintServiceLookup::lookupDefaultPrintService, lifespan);
```

src/qz/printer/info/NativePrinter.java
```diff
@@ -69,6 +70,11 @@ public boolean equals(Object o) {
     }
 
     private final String printerId;
+
+    // getName() reaches out to native and is expensive on mac. This name is cached, and is refreshed if lifespan has elapsed
+    // todo Fix before merging: mention upsteam bug report
```
tresf, line 75:
```suggestion
    // PrintService.getName() is slow, use a cache instead per JDK-XXXXXXX
    // TODO: Remove this comment when upstream bug report is filed
```

src/qz/common/CachedObject.java
tresf: Nevermind, you changed it from [your gist](https://gist.github.com/Vzor-/aeb69492221b89961ef53fde2e56b9be#file-cachedobject-java-L23) and it's using `toMillis` now. 🍻

src/qz/common/CachedObject.java
tresf, line 14:
```suggestion
    public static final long DEFAULT_LIFESPAN = 5000; // in millis
```

tresf: Oh, it's `millis`.  Let's comment that.

tresf:
I really like the introduction of the cached object handling.  I have a few cosmetic requests but I think this is ready for testing.

Vzor-:
I have been load testing this with a handful of virtual printer. Under high status traffic, sanitizePrinterName slows down on mac. I am investigating and will have a commit for it soon.

Vzor-:
After further examination I found the slowdown was occurring when we called `PrintServiceLookup.lookupPrintServices(null, null)`. In a stand-alone test, I was only able to recreate the issue after setting "sun.java2d.print.polling" to `false` as we do in [TrayManager](https://github.com/qzind/tray/blob/a7c0b7e0adf919c56817d7b98993a4a3e27b3568/src/qz/common/TrayManager.java#L156), and conversely, removing this line fixes the slowdown in our code. I dug into the JRE and found that internally, java is adding PrintServices to an arraylist and calling .indexOf on that list. This calls `equals()` many times, which itself calls getName, the original slowdown associated with this PR. It seems that not only is this function expensive, but it becomes slower the more it is called. I have demonstrated the following stand0alone test.

https://gist.github.com/Vzor-/0a4d5f2883dcfc4539426b6948c029ca

tresf:
> After further examination I found the slowdown was occurring when we called `PrintServiceLookup.lookupPrintServices(null, null)`.
> 
> In a stand-alone test, I was only able to recreate the issue after setting "sun.java2d.print.polling" to `false` as we do in [TrayManager](https://github.com/qzind/tray/blob/a7c0b7e0adf919c56817d7b98993a4a3e27b3568/src/qz/common/TrayManager.java#L156), and conversely, removing this line fixes the slowdown in our code.
> 
> I dug into the JRE and found that internally, java is adding PrintServices to an arraylist and calling .indexOf on that list. This calls `equals()` many times, which itself calls getName, the original slowdown associated with this PR. It seems that not only is this function expensive, but it becomes slower the more it is called. I have demonstrated the following stand-alone test.
> 
> https://gist.github.com/Vzor-/0a4d5f2883dcfc4539426b6948c029ca

A friendly reminder, this MUST be filed upstream (first with our JDK support provider, and then they'll report it to OpenJDK).

In regards to the caching improvements, the benchmarks above are helpful, but hard to understand.  I think they should be better quantified in terms of API calls.  For example, `startListening` with 100 events BEFORE/AFTER.  Then again with 1,000 events, then 10,000 events.  This way we can decide if complicating the codebase for this upstream bug is warranted or not.

src/qz/printer/info/NativePrinter.java
Vzor-: In this instance, calling cachedName.get() would invoke getName(), causing an infinite loop. getNameNative is already private (in retrospect, it's interesting that we have access to the function from CachedObject).

src/qz/common/CachedObject.java
Vzor-: That is a good point. lastObject? storedObject?

Vzor-:
Screenshot 2023-09-12 at 2 31 16 AM

WIP, I am running a test with statuses now, and I think I will run 1 more where there is a 5 second delay between prints, so each one starts with a expired cache.

src/qz/printer/info/NativePrinter.java
tresf: No, it's not infinite, per native caller passed to the instantiation:

```
printService.value::getName
```

src/qz/common/CachedObject.java
tresf: I'd be fine with either.

Vzor-:
I am removing the threading from this PR and moving it to its own PR.

Vzor-:
The printer list is falling back on [lpstat, -l, -p] until `CachedObject<PrintService> cachedDefault` goes stale and refreshes. ill look into that now

Vzor-:
So the results with the newest commit are very good, much better than expected.
Screenshot 2023-09-27 at 2 01 19 AM
As we can see the "Wrapped Cache" is so fast, we can't really see it. Here it is with the uncached line removed.
Screenshot 2023-09-27 at 2 03 00 AM
The entire time of execution is 15 seconds, as opposed to 247 seconds with cached. Since our lifespan is only 5 seconds, that means we only updated the cache 3 times, which is why we see virtually no slowdown. I will run a test overnight that delays each print by 5 seconds to maximize the cache updates.

src/qz/printer/PrintServiceMatcher.java
```diff
@@ -30,15 +31,27 @@
 
 public class PrintServiceMatcher {
     private static final Logger log = LogManager.getLogger(PrintServiceMatcher.class);
+    //todo: include jdk version test for caching when JDK-7001133 is resolved
+    private static final boolean useCache = SystemUtilities.isMac();   // PrintService is slow, use a cache instead per JDK-7001133
```
tresf, line 35: Shouldn't this be `!SystemUtilities.isWindows()` so that it applies to all CUPS-based systems?

src/qz/printer/PrintServiceMatcher.java
```diff
@@ -84,6 +98,8 @@ public static NativePrinter matchPrinter(String printerSearch, boolean silent) {
 
         if (!silent) { log.debug("Searching for PrintService matching {}", printerSearch); }
 
+        // Fix for https://github.com/qzind/tray/issues/931
+        // This is more than an optimization, removal will lead to a regression
```
tresf, line 102: This comment does not appear to reference any code changes.  Is it outdated now?

src/qz/printer/info/CachedPrintService.java
```diff
@@ -0,0 +1,130 @@
...
+        return new CachedPrintService(PrintServiceLookup.lookupDefaultPrintService());
+    }
+
+    private static PrintService[] innerLookupPrintServices() {
+        PrintService[] printServices = PrintServiceLookup.lookupPrintServices(null, null);
+        for (int i = 0; i < printServices.length; i++) {
+            printServices[i] = new CachedPrintService(printServices[i]);
+        }
+        return printServices;
+    }
```
tresf, line 43: Do these static functions belong here?  They seem like they're (mostly) unrelated to the class' function and may be better suited for a dedicated class, e.g. `PrintServiceLookupCache`, etc.

src/qz/printer/PrintServiceMatcher.java
Vzor-: yes

src/qz/printer/PrintServiceMatcher.java
Vzor-: yea we can ditch it. it was for me, because my first pass i broke this code, thinking the getdefault was just an optimization.

src/qz/printer/info/CachedPrintService.java
Vzor-: yea they could exist in a dedicated class. They could even live in printServiceMatcher. The only reason they are here is because I wanted all of the caching stuff in one place

src/qz/printer/PrintServiceMatcher.java
tresf: Oh whoops, wrong bug report.  Yeah it can stay!

src/qz/printer/info/CachedPrintServiceLookup.java
```diff
@@ -31,15 +32,32 @@ public static PrintService[] lookupPrintServices() {
         return cachedPrintServices.get();
     }
 
-    private static PrintService innerLookupDefaultPrintService() {
+    private static PrintService wrapDefaultPrintService() {
+        PrintService javaxPrintService = PrintServiceLookup.lookupDefaultPrintService();
+        // If this CachedPrintService already exists, reuse it rather than wrapping a new one
+        CachedPrintService oldCachedPrintService = getMatch(oldPrintServices, javaxPrintService);
+        if (oldCachedPrintService != null) return oldCachedPrintService;
         return new CachedPrintService(PrintServiceLookup.lookupDefaultPrintService());
```
tresf, line 40: This will use a different pointer for the default print service and the copy of it in the listing.

Vzor-: hmm, ill test that

Vzor-: We could do this 
``` java
private static PrintService wrapDefaultPrintService() {
    PrintService javaxPrintService = PrintServiceLookup.lookupDefaultPrintService();
    // If this CachedPrintService already exists, reuse it rather than wrapping a new one
    CachedPrintService oldCachedPrintService = getMatch(oldPrintServices, javaxPrintService);
    if (oldCachedPrintService != null) return oldCachedPrintService;
    // If its not in our list, force a refresh of the cache
    cachedPrintServices.get(true);
    return getMatch(oldPrintServices, javaxPrintService);
}
```
note, this could in theory return null, though i don't think that scenario is actually possible.

Vzor-: This was fixed here https://github.com/qzind/tray/pull/1124/commits/49bb0691c313bc65c7ad056fab3223ba5ea45816#diff-74db1b7f463a02e2ec7078941f6f0f0cf8ef7cb146b0cd95576229069272ec89R42
It also avoids calling `lookupDefaultPrintService` when we are only looking for default.

tresf: We also need to ensure this works when there is no default print service (e.g. when `PrintServiceLookup.lookupDefaultPrintService()` returns `null`).

tresf:
This code should be good to be merged.  Just some final testing.

Vzor-:
Tests:
If a service is stale, and you attempt to print to it, we need to verify that this failure is graceful.

- [ ] Windows (shouldn't be affected)
- [ ] Mac
- [ ] Linux

Simple congestion test, lots of prints, lots of statuses 

- [ ] Windows (shouldn't be affected)
- [ ] Mac
- [ ] Linux

Vzor-:
Do not merge yet. I think i found an issue on linux but I will have to confirm it tomorrow.

tresf:
FYI, if a job is sent to Linux after a printer has been removed, but before the cache has refreshed, this may throw the following error: (this is a bit misleading, but I consider it a rather graceful catch).

```
[ERROR] 2023-10-11T14:57:08,332 @ qz.utils.PrintingUtilities:214
    Failed to print
sun.print.PrintJobFlavorException: invalid flavor
    at sun.print.UnixPrintJob.print(UnixPrintJob.java:347) ~[?:?]
    at qz.printer.action.PrintRaw.waitForPrint(PrintRaw.java:470) ~[production/:?]
    at qz.printer.action.PrintRaw.printToPrinter(PrintRaw.java:427) ~[production/:?]
    at qz.printer.action.PrintRaw.print(PrintRaw.java:354) ~[production/:?]
    at qz.utils.PrintingUtilities.processPrintRequest(PrintingUtilities.java:204) ~[production/:?]
    at qz.ws.PrintSocketClient.processMessage(PrintSocketClient.java:311) ~[production/:?]
    at qz.ws.PrintSocketClient.lambda$onMessage$0(PrintSocketClient.java:169) ~[production/:?]
    at java.lang.Thread.run(Thread.java:833) [?:?]
```

Vzor-:
All of my tests passed. Looks ready to merge from my end.

PR #1131, opened 2023-05-21
Initial support for Linux system-wide certificates

### LINKED ISSUE #518: Add Epiphany support
Epiphany web browser currently is incompatible with QZ Tray.  It appears to use a separate certificate store.  The resutl of this is OSs which ship with Epiphany by default (e.g. ElementaryOS) can't use QZ Tray unless they install an alternative browser (e.g. Firefox, Chrome, etc).

According to the Gnome wiki, installing a certificate into the Gtk shared certificate store is as simple as:

```bash
sudo trust anchor path/to/root-ca.crt
```

As part of the certs-rewrite effort, we're improving the browser support for the 2.1 launch.  This change should be pretty easy to add into the new Java-based installer.

tresf:
Currently, Linux browser support is limited to Mozilla (Firefox, IceCat, PaleMoon, etc) and Chromium (Chrome, Edge, Brave, etc) browsers.

To support Gnome's Epiphany browser (simply called "Web" in the OS), we need to be able to install our certificate OS-wide, similar to how we do this on Windows and MacOS.

Closes #518

### DIFF
--- a/src/qz/installer/LinuxInstaller.java
+++ b/src/qz/installer/LinuxInstaller.java
@@ -1,6 +1,7 @@
 package qz.installer;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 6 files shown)


tresf:
Hmm... unrelated to this PR, but for some reason Firefox's restart warning isn't displaying on Fedora.

```bash
pgrep -x "firefox|MainThread|GeckoMain"
# returns a process
```

tresf:
Tested:
✅ Ubuntu 20.04
✅ Ubuntu 18.04
✅ Fedora 34

PR #1139, opened 2023-06-03
Add printer debug information

tresf:


tresf:
Rebased on master so we can use this in the future if we decide to.  This PR is only partially helpful because it helps capture crashes due to iterating over Media but doesn't do much in regards to the original building of printer listings, so if a driver crashes the iteration earlier in the crawling process, this PR doesn't offer any insight.

For now, I'm going to close this out and leave the code around.  In the future, we should probably add a few more capture points and toggle this on with a system property.

Issue #1143, opened 2023-06-13
Set default Socket timeout


tresf:
By default in Java, sockets have no timeout value, meaning if there is a very long time connecting to a host or port, the websocket will crash.

An example of this is in `qz.networking.info(...)`, which calls the following code:

https://github.com/qzind/tray/blob/b3ae4b81db19a2c9c747aff09998c87e92533302/src/qz/utils/NetworkUtilities.java#L105-L112

If this Socket connection takes more than 5 minutes, it can crash out the entire websocket.  According to the JavaDoc, the default value is `infinity`.

https://docs.oracle.com/javase/8/docs/api/java/net/ServerSocket.html#getSoTimeout--

```
2023-06-12T13:44:35,602 [INFO] Initiating a temporary connection to "google.com:443" to determine main Network Interface
2023-06-12T13:49:35,906 [ERROR] Timeout error (Lost connection with client)
```

We should also ensure other areas aren't making the same mistake.

PR #1146, opened 2023-06-15
Add default timeouts to sockets

akberenz:
Addresses #1143


tresf:
This theoretically fixes the timeout issue, but it will not fix the long blocking, so we might have to tweak this value down the road.  I like the 4 minute timeout for not, as it will cause the least amount of regression.

PR #1172, opened 2023-08-31
Try to make callCert() rejectable

### LINKED ISSUE #1166: Allow certificate rejection to prevent further calls
Currently the default behavior when a certificate cannot be retrieved is to send a `null` certificate.  This is good default behavior as it allows printing to succeed in cases where a certificate is unavailable or misconfigured.

However, this may not always be desired, there may be scenarios where a customer would rather queue up this print request until the certificate promise succeeds.   In this case, we should allow a configuration option to `reject()` WITHOUT calling `sendCert(null)`.

https://github.com/qzind/tray/blob/95db68d7e9b883380d4a949de780dd32c31a239f/js/qz-tray.js#L364-L367

tresf:
Closes #1166

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -363,7 +363,11 @@ var qz = (function() {
 
                     _qz.security.callCert().then(sendCert).catch(function(error) {
                         _qz.log.warn("Failed to get certificate:", error);
-                        sendCert(null);
+                        if(_qz.security.rejectOnCertFailure) {
+                            return _qz.tools.reject(error);
+                        } else {
+                            sendCert(null);
+                        }
                     });
                 },
 
@@ -578,6 +582,8 @@ var qz = (function() {
 
             /** Signing algorithm used on signatures */
             signAlgorithm: "SHA1",
+
+            rejectOnCertFailure: false,
 
             needsSigned: function(callName) {
                 const undialoged = [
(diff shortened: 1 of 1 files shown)


js/qz-tray.js
```diff
@@ -2595,11 +2602,15 @@ var qz = (function() {
...
+             * @param {Object} [options] Configuration options for the certificate resolver
+             *  @param {boolean} [options.rejectOnFailure=[false]] Overrides default behavior to call resolve with a blank certificate on failure.
              * @memberof qz.security
              */
-            setCertificatePromise: function(promiseHandler) {
+            setCertificatePromise: function(promiseHandler, options) {
                 _qz.security.certHandler = promiseHandler;
+                if(options && options.rejectOnFailure) {
+                    _qz.security.rejectOnCertFailure = options.rejectOnFailure;
+                }
```
tresf, line 2613:
```suggestion
                _qz.security.rejectOnCertFailure = options && options.rejectOnFailure;
```

@akberenz should we switch this so that it resets every call?  I was thinking and it may be unobvious if this setting is sticky and remains its last value when it's omitted on subsequent calls.

akberenz: Yeah, I think that change would be for the best from a usability view.

PR #1181, opened 2023-09-25
Fix a deadlock with printer status threading

Vzor-:
This fixes a deadlock discovered while testing https://github.com/qzind/tray/issues/1116. It should also alleviate the slowdown to a degree, though not as much as the JDK fix that is in the works, or the caching in #1124.

I still need to run some test, but this is the threading part of #1124 with the caching removed. The two features are a bit too disparate to exist in a PR together.

### DIFF
--- a/src/qz/printer/status/StatusMonitor.java
+++ b/src/qz/printer/status/StatusMonitor.java
@@ -8,6 +8,7 @@ import org.codehaus.jettison.json.JSONObject;
 import org.eclipse.jetty.util.MultiMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.eclipse.jetty.websocket.api.Session;
 import qz.printer.PrintServiceMatcher;
 import qz.printer.info.NativePrinterMap;
 import qz.utils.PrintingUtilities;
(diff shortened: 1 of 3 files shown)


src/qz/printer/status/StatusMonitor.java
```diff
@@ -74,45 +78,25 @@ public synchronized static void closeNotificationThreads() {
...
     }
 
-    public synchronized static boolean startListening(SocketConnection connection, JSONObject params) throws JSONException {
+    public synchronized static boolean isListening(SocketConnection connection) {
+        return statusSessions.containsKey(connection);
+    }
+
+    public synchronized static boolean startListening(SocketConnection connection, Session session, JSONObject params) throws JSONException {
+        statusSessions.putIfAbsent(connection, new StatusSession(session));
         JSONArray printerNames = params.getJSONArray("printerNames");
```
Vzor-, line 87: We should do all JSON work BEFORE editing any list, just in case we throw a JSONException

tresf, line 87:
```suggestion
        JSONArray printerNames = params.getJSONArray("printerNames");
        statusSessions.putIfAbsent(connection, new StatusSession(session));
```

tresf:
Testing: 

* MacOS: No observable regressions
* Windows: No observable regressions
* Ubuntu: No observable regressions

PR #1184, opened 2023-10-02
Capture CANCELLED WebView state

### LINKED ISSUE #1183: HTML Worker.State.CANCELLED throws TimeoutException
Steps:

```js

var config = qz.configs.create("PDF");
var data = [{
   type: "pixel",
   format: "html",
   flavor: "file",
   data: "https://filebin.net/1jij1l087a0ts7pw/receipt.html"
}];

qz.print(config, data).catch(e => console.error(e));
```

tresf:
Closes #1183


tresf:
Thanks @akberenz!  12a4b6d perfectly shows the redirect now, which shows the `response-content-disposition=attachment`, which we don't currently support.

Gluon did offer a workaround to save this file to a temporary location, but this has other implications, such as relative links breaking.  I think the best approach is to log, unlatch and let the developer fix it on the back-end.

Merging.

PR #1191, opened 2023-10-16
Clear Queue (cancel-jobs)

### LINKED ISSUE #1185: Add ability to clear a printer's queue
Request to add the ability to clear a print queue.  It should be assumed that a person can only clear their own jobs.

> We [...] had an issue related to the OS spooler caused by sending a bad ESC command to Star printers. Despite immediately correcting the change, we had no way to clear the queue and had to manually clear it on every device which had a huge impact on customers and our support team. Our primary objective is to avoid this havoc in the future.

**Windows:**

* Perhaps using `JOB_CONTROL_DELETE` on [`SetJob`](https://learn.microsoft.com/en-us/windows/win32/printdocs/setjob).

**MacOS/Linux:**

* TBD... Perhaps an API that calls [`CANCEL`](https://www.cups.org/doc/man-cancel.html) on the print job?

Vzor-:
This is a feature addition requested in #1185. It adds the ability to clear a specific printer's queue. The delete jobs should be pending only, leaving behind 'printed' and 'retained' jobs. 

Todo: 
- [x] Support specific jobIDs

closes #1185

### DIFF
--- a/src/qz/communication/WinspoolEx.java
+++ b/src/qz/communication/WinspoolEx.java
@@ -1,0 +1,24 @@
+package qz.communication;
+
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+import com.sun.jna.platform.win32.Winspool;
+import com.sun.jna.win32.W32APIOptions;
+
+@SuppressWarnings("unused")
+public interface WinspoolEx extends Winspool {
+    WinspoolEx INSTANCE = Native.load("Winspool.drv", WinspoolEx.class, W32APIOptions.DEFAULT_OPTIONS);
+
+    int JOB_CONTROL_NONE = 0x00000000; // Perform no additional action.
+    int JOB_CONTROL_PAUSE = 0x00000001; // Pause the print job.
+    int JOB_CONTROL_RESUME = 0x00000002; // Resume a paused print job.
+    int JOB_CONTROL_CANCEL = 0x00000003; // Delete a print job.
+    int JOB_CONTROL_RESTART = 0x00000004; // Restart a print job.
+    int JOB_CONTROL_DELETE = 0x00000005; // Delete a print job.
+    int JOB_CONTROL_SENT_TO_PRINTER = 0x00000006; // Used by port monitors to signal that a print job has been sent to the printer. This value SHOULD NOT be used remotely.
+    int JOB_CONTROL_LAST_PAGE_EJECTED = 0x00000007; // Used by language monitors to signal that the last page of a print job has been ejected from the printer. This value SHOULD NOT be used remotely.
+    int JOB_CONTROL_RETAIN = 0x00000008; // Keep the print job in the print queue after it prints.
+    int JOB_CONTROL_RELEASE = 0x00000009; // Release the print job, undoing the effect of a JOB_CONTROL_RETAIN action.
+
+    boolean SetJob(Pointer hPrinter, int JobId, int Level, Pointer pJob, int Command);
+}
--- a/src/qz/communication/WinspoolEx.java
+++ b/src/qz/communication/WinspoolEx.java
@@ -6,6 +6,9 @@ import com.sun.jna.platform.win32.WinNT;
 import com.sun.jna.platform.win32.Winspool;
 import com.sun.jna.win32.W32APIOptions;
 
+/**
+ * TODO: Remove when JNA 5.14.0+ is bundled
+ */
 @SuppressWarnings("unused")
 public interface WinspoolEx extends Winspool {
     WinspoolEx INSTANCE = Native.load("Winspool.drv", WinspoolEx.class, W32APIOptions.DEFAULT_OPTIONS);
--- a/src/qz/communication/WinspoolEx.java
+++ b/src/qz/communication/WinspoolEx.java
@@ -2,6 +2,7 @@ package qz.communication;
 
 import com.sun.jna.Native;
 import com.sun.jna.Pointer;
+import com.sun.jna.platform.win32.WinNT;
 import com.sun.jna.platform.win32.Winspool;
 import com.sun.jna.win32.W32APIOptions;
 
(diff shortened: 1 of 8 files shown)


Vzor-:
I am developing across 2 machines so there will be plenty of WIP commits.

src/qz/utils/PrintingUtilities.java
```diff
@@ -219,4 +229,37 @@ public static void processPrintRequest(Session session, String UID, JSONObject p
...
+                    boolean result = WinspoolEx.INSTANCE.SetJob(phPrinter.getValue(), job.JobId, 0, null, WinspoolEx.JOB_CONTROL_DELETE);
+                    if (result) {
+                        deletedCount++;
+                    } else {
+                        log.warn("Job deletion error for job#{}, error code:{}", job.JobId, Kernel32.INSTANCE.GetLastError());
+                    }
+                }
+                log.info("Deleting {} jobs", deletedCount);
+            } else {
+                ArrayList<Integer> jobIds = CupsUtils.listJobs(printer.getPrinterId());
```
Vzor-, line 253: Does this need macNameFix like https://github.com/qzind/tray/blob/2605e2a8b7c03159f173a6489154a106e30c122e/src/qz/printer/status/StatusMonitor.java#L94?

src/qz/communication/WinspoolEx.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+import com.sun.jna.platform.win32.WinNT;
+import com.sun.jna.platform.win32.Winspool;
+import com.sun.jna.win32.W32APIOptions;
+
+@SuppressWarnings("unused")
+public interface WinspoolEx extends Winspool {
+    WinspoolEx INSTANCE = Native.load("Winspool.drv", WinspoolEx.class, W32APIOptions.DEFAULT_OPTIONS);
```
Vzor-, line 11: I am shocked this worked, we may want to make sure this is a good way to do this. I found no documentation about extending an interface from JNA.

src/qz/communication/WinspoolEx.java
```diff
@@ -0,0 +1,25 @@
...
+
+    int JOB_CONTROL_NONE = 0x00000000; // Perform no additional action.
+    int JOB_CONTROL_PAUSE = 0x00000001; // Pause the print job.
+    int JOB_CONTROL_RESUME = 0x00000002; // Resume a paused print job.
+    int JOB_CONTROL_CANCEL = 0x00000003; // Delete a print job.
+    int JOB_CONTROL_RESTART = 0x00000004; // Restart a print job.
+    int JOB_CONTROL_DELETE = 0x00000005; // Delete a print job.
+    int JOB_CONTROL_SENT_TO_PRINTER = 0x00000006; // Used by port monitors to signal that a print job has been sent to the printer. This value SHOULD NOT be used remotely.
+    int JOB_CONTROL_LAST_PAGE_EJECTED = 0x00000007; // Used by language monitors to signal that the last page of a print job has been ejected from the printer. This value SHOULD NOT be used remotely.
+    int JOB_CONTROL_RETAIN = 0x00000008; // Keep the print job in the print queue after it prints.
```
Vzor-, line 21: Side note: This could be useful in the future if we want to support job-data statuses for printers that aren't preconfigured to retain jobs.

src/qz/communication/WinspoolEx.java
tresf: Whoops, I see `SetJob` is being added to the signature, so I take back that last part. 😅

src/qz/utils/PrintingUtilities.java
```diff
@@ -219,4 +229,37 @@ public static void processPrintRequest(Session session, String UID, JSONObject p
         }
     }
 
+    public static void cancelJobs(Session session, String UID, JSONObject params) {
+        try {
+            NativePrinter printer = PrintServiceMatcher.matchPrinter(params.getString("printerName"));
+            if (SystemUtilities.isWindows()) {
+                WinNT.HANDLEByReference phPrinter = new WinNT.HANDLEByReference();
+                WinspoolEx.INSTANCE.OpenPrinter(printer.getName(), phPrinter, null);
```
tresf, line 237: I believe this should be guarded, quoting:

> **Return value**
> *  If the function succeeds, the return value is a `nonzero` value.
> * If the function fails, the return value is `zero`.

src/qz/utils/PrintingUtilities.java
tresf: No, AFAIR.

It should be documented better, but `getPrinterId` is the "ugly" CUPS id of the printer (e.g. "Zebra_LP2844"), so this is a safe operation, not unlike our raw printing API here:

https://github.com/qzind/tray/blob/2605e2a8b7c03159f173a6489154a106e30c122e/src/qz/printer/status/CupsUtils.java#L57

The reason the helper isn't needed is because `NativePrinter` is "namefix" aware, as it's constructed with two values:
* name: What Java calls this printer
* printerId: What the OS calls this printer

In Windows and Linux, these values are always identical, but they're different on Mac, so we're unknowingly appling the name fix by using the `NativePrinter` object.

On the other hand, `StatusMonitor` is accepting raw `String` values as input, so they must be sanitized for Mac.

src/qz/communication/WinspoolEx.java
tresf: > I would expect this to work if the method signatures are similar enough however if there's any methods in Winspool that either aren't in WinspoolEx or take slightly different parameters in WinspoolEx, I would expect calling them to cause a crash.

Ok, this statement is off the mark.  Since you chose [`WinspoolEx`](https://stackoverflow.com/a/4039059/3196753) as a name, I assumed this actually existed.  What you're doing here is perfectly sane.  Pragmatically, we would open a PR to contribute these back to JNA contib.

Vzor-: Yea i guess if they add `SetJob` to JNA this will break. It may make sense to pr JNA to add it, even though that will make that issue more likely. For now, I don't know of any other way to hook in `SetJob`

tresf: Yes, but it would be a compilation error, and only if the signature was different, and only after we updated the dependency, so our CI would catch it. https://www.geeksforgeeks.org/the-override-annotation-in-java/.

Vzor-:
I think thats the last of it, ready for review.

Vzor-:
setJob will be supported in the next version of JNA. https://github.com/java-native-access/jna/pull/1556

tresf:
Merged.  We'll do a clean-up commit once JNA 5.14.0+ is released and bundled.

PR #1199, opened 2023-10-21
Cleanup old log format

### LINKED ISSUE #1196: Cleanup old ".1", ".2" log files
In QZ Tray 2.2.2 and older log files would be named `debug.log.1`, `debug.log.2`, etc.

Since #1160, QZ Tray generates logs in the format: `debug.1.log`, `debug.2.log`, etc.

Is there an easy way to cleanup the old ones on rotation?

tresf:
Perform cleanup of old log file rotations:
* `debug.log.1` 
* `debug.log.2` 
* `debug.log.3` 
* `debug.log.4` 
* `debug.log.5` 

In favor of new format introduced in #1160:
* `debug.1.log` 
* `debug.2.log` 
* `debug.3.log` 
* `debug.4.log` 
* `debug.5.log` 

... but not without ensuring a replacement adjacent log exists first, to avoid any debugging data loss.

Closes #1196

### DIFF
--- a/src/qz/installer/Installer.java
+++ b/src/qz/installer/Installer.java
@@ -174,31 +174,24 @@ public abstract class Installer {
         // Convert old < 2.2.3 log file format
         Path logLocation = USER_DIR;
         int oldIndex = 0;
-        int newIndex = 0;
-        while(true) {
+        File oldFile;
+        do {
             // Old: debug.log.1
-            File oldFile = logLocation.resolve("debug.log." + oldIndex++).toFile();
+            oldFile = logLocation.resolve("debug.log." + ++oldIndex).toFile();
             if(oldFile.exists()) {
-                // Rename to new format
-                inner:
-                while(true) {
-                    // New: debug.1.log
-                    File newFile = logLocation.resolve("debug." + ++newIndex + ".log").toFile();
-                    if (!newFile.exists()) {
-                        oldFile.renameTo(newFile);
-                        log.info("Migrated log file {} to new location {}", oldFile, newFile);
-                        break inner;
-                    }
-                }
-                continue;
+                // New: debug.1.log
+                File newFile = getNextLogFile(logLocation, 0);
+                oldFile.renameTo(newFile);
+                log.info("Migrated log file {} to new location {}", oldFile, newFile);
             }
-            // Loop at least as high as our rollover strategy in case there are human made gaps
-            if(oldIndex <= rolloverCount) continue;
-            break;
-
-        }
+        } while(oldFile.exists() || oldIndex <= rolloverCount);
 
         return this;
+    }
+
+    private File getNextLogFile(Path logLocation, int index) {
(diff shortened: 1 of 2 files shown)


akberenz:
I'm not sure this quite preserves logs how you indicated it should?
We don't convert any of the old format to the new and the rollover increments the number so `.log.5` will be the oldest file, but if `.5.log` hasn't been rolled into yet with the this code, that oldest file will still be around even though newer, lower numbered old logs will have been removed.

If anything, we should get a count of both the old format and the new format then delete the old format files, highest number first, down to the rollover count.

tresf:
> that oldest file will still be around even though newer, lower numbered old logs will have been removed.

Yes, but it's a self-healing wound, once 5.log is created, log.5 will be deleted the next time it opens, no?

> If anything, we should get a count of both the old format and the new format then delete the old format files, highest number first, down to the rollover count.

I don't want to purge good logs unless they're outdated.

tresf:
Oh, I see what you're saying... It deleted oldest files last instead of first.... Hmm ok, I'll have to rework this then...

src/qz/installer/Installer.java
```diff
@@ -170,6 +170,31 @@ public Installer removeLibs() {
...
     }
 
+    public Installer cleanupLegacyLogs(int rolloverCount) {
+        // Convert old < 2.2.3 log file format
+        Path logLocation = USER_DIR;
+        int nextFile = 0;
+        for(int i = 1; i <= rolloverCount; i++) {
+            // Old: debug.log.1
+            File oldFile = logLocation.resolve("debug.log." + i).toFile();
+            if(oldFile.exists()) {
```
akberenz, line 180: why do we specify a rolloverCount instead of just looping until the first non-existent .log.n file?

tresf: The original intention was to reduce the scope.  At that time I didn't know that the log engine was smart enough to cleanup similarly named files, so I only wanted to touch what we created.  Knowing what we know now, this can safely increment as long as there's files. I'll  do that now.

tresf: Should be addressed via 5585d84.

src/qz/installer/Installer.java
```diff
@@ -170,6 +170,37 @@ public Installer removeLibs() {
...
+                        oldFile.renameTo(newFile);
+                        log.info("Migrated log file {} to new location {}", oldFile, newFile);
+                        break inner;
+                    }
+                }
+                continue;
+            }
+            // Loop at least as high as our rollover strategy in case there are human made gaps
+            if(oldIndex <= rolloverCount) continue;
+            break;
```
akberenz, line 197: this is a bit confusing to read, instead can we use a do..while that loops if oldFile exists or index is less than rollover?
```
do { .. } while(oldFile.exists() || oldIndex <= rolloverCount);
```

also, pretty sure break/continue only works on the innermost loop, so the "inner" label in unnecessary

PR #1222, opened 2023-11-26
Add support for OpenJDK 21

### LINKED ISSUE #1218: Upgrading to OpenJDK 21
QZ Tray 2.3.0 will ship with OpenJDK 21 LTS (or OpenJDK 17 if 21 is too new)

High level tasks:

* [ ] Fix illegal reflections
* [ ] Bump JavaFX
* [ ] jlink / Packaging differences
* [ ] Bump CIs
* [ ] Company Branded portal

tresf:
Partially addresses/related #1218


tresf:
It appears that JavaFX 21 is not released for Linux x86_64.  We've filed a bug report with Gluon for this here: `SUPQZ-19` (private)

tresf:
> CI-builds are failing. It appears that JavaFX 21 "monocle" is not released for Linux x86_64. We've filed a bug report with Gluon for this here: `SUPQZ-19` (private)

Gluon is going to try to upstream the monocle features that we need into JavaFX proper, which would theoretically fix this.   Since this is going to take some time, we should probably revert back to JavaFX 19 so that this PR is in a passing state and update JavaFX as a separate initiative.

tresf:
This PR should be split up into two tasks:

1. Update build system to support JDK21
2. Switch build system to default to JDK21

The reasoning is that there is some uncertainty for #1225 in regards to exactly which JDK we should be targeting and the more PRs that depend on JDK21+ patches, the harder these will be to manage long-term.

PR #1225, opened 2023-11-30
Move away from "Monocle" in favor of "Headless"

tresf:
Switches to the new experimental `glass.platform="Headless"` property, removing the special monocle JavaFX dependency.

* This is built atop the `jdk21` branch, but the only relevant commits should be those on or after https://github.com/qzind/tray/commit/d87055beb13b35f0d8241800d577e9801a8a81bc.
* At time of writing this, this flag does not yet exist in OpenJFX as it's still undergoing internal R&D

Fixes https://github.com/qzind/tray/issues/1253.


.idea/misc.xml
```diff
@@ -8,5 +8,5 @@
       <item index="3" class="java.lang.String" itemvalue="org.eclipse.jetty.websocket.api.annotations.OnWebSocketMessage" />
     </list>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="11" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" project-jdk-name="temurin-21" project-jdk-type="JavaSDK" />
```
tresf, line 11: Hmm... I'm not sure why I hard-coded `temurin` here... Should be fixed before merge.

PR #1228, opened 2023-12-29
EscPos sample file

erikn69:
ESC/POS Sample file, why isn't there one?


tresf:
> ESC/POS Sample file, why isn't there one?

Because these types of files are not easily hand-editable due to the escape characters, making the `.txt` file extension very misleading.

> ExampleMart Ltd.

This doesn't really match the rest of our examples.  Where was this lifted from?

erikn69:
>Because these types of files are not easily hand-editable due to the escape characters, making the .txt file extension very misleading.

I suspected that, but I didn't have an example file on hand to test a Virtual ESC/POS printer.

> This doesn't really match the rest of our examples. Where was this lifted from?

[receipt-print-hq/escpos-tools/receipt-with-logo.bin](https://github.com/receipt-print-hq/escpos-tools/blob/master/receipt-with-logo.bin), 

[receipt-print-hq/escpos-tools/esc2html.md](https://github.com/receipt-print-hq/escpos-tools/blob/master/doc/esc2html.md) I  built a service to capture the raw print data and show a preview in html by executing the script, it is for testing

tresf:
Thanks.  Since these file are MIT licensed, we can legally bundle them with our demo.

If you notice, the project uses the `.bin` file extension, which I think is a bit more accurate as to the file's contents.

With regards to bundling an example file moving forward, we're not opposed to this, however we would likely prefer a small, curated example, likely one without an embedded image that uses the most common (and most compatible) ESC/POS commands.

We also prefer to add some branding to the receipt to indicate if it's was from our software.

This PR is better than nothing, so I think it's a good idea to leave it open until one's added that supercedes it.

erikn69:
I'll edit it

tresf (approved):

tresf:
Thanks!!

PR #1233, opened 2024-02-13
Sign message example with Node and Typescript

tomashlavac:
Hello,
I've added example of message signing with Node and Typescript.
Hope it helps.

### DIFF
--- a/assets/signing/sign-message-node.ts
+++ b/assets/signing/sign-message-node.ts
@@ -1,0 +1,63 @@
+/*
+ * Node.js typescript signing example
+ * Echoes the signed message and exits
+ * Request: /signQz?requestToSign=dataToSign
+ * Imports:
+ * import crypto from 'crypto'
+ * import {Request, Response} from 'express'
+ * import fs from 'fs'
+ */
+
+// #########################################################
+// #             WARNING   WARNING   WARNING               #
+// #########################################################
+// #                                                       #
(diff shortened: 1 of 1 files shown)


tomashlavac:
Hi @tresf ,
please can you review this ?
Thanks

tresf:
@tomashlavac thanks kindly for this.  I'm not really sure how to test this PR.  I'm also not very familiar with TypeScript, which makes it hard to read this example and provide any feedback necessary for merge.

For example:

```ts
req: Request<unknown, unknown, unknown, IQzSignHandlerRequest>,
```

I would expect this to simply be `string`.  If `express` requires all of this, it is very annoying to me 🍻 

```ts
res.status(200).send(signature)
```

I would expect 200 to be default status for HTTP requests.

```ts
const sign = crypto.createSign('SHA512')
```

For consistency, this should warn users:

```ts
// Use 'SHA1' for QZ Tray 2.0 and older
```

... and last, I believe that this example would be best to kept in sync with our existing node example here: https://github.com/qzind/tray/blob/master/assets/signing/sign-message.node.js.  This means we should ideally have two identical files, one that uses typescript and another that uses `.ts`.

We are not JavaScript developers, so we're welcome to feedback on our existing examples, but consistency is important for maintenance.

tresf:
Closing due to lack of response from OP.

PR #1247, opened 2024-04-03
Added 'default' ColorType

### LINKED ISSUE #1179: Add colorType: 'default' for pixel prints
A question came in today:

> [...] why is QZ Tray using it's own colour settings & not adhering to the windows print queue?

This is a very good question.  I've done a code review, and I believe `colorType: null` will allow QZ Tray to honor the printer's default preferences, but this is not available in the demo (sample.html) nor is it documented in the API.  We should have a more intuitive value to honor the printer's default `colorType`.

Vzor-:
Fixes #1179 
Null was intended to act as a 'default' option, but does not. Since it was not a documented feature, we will add a 'default' option rather than change the existing behavior of null.

Printers have a default color setting on windows, it can be "Black & White" or "Color". This is a table showing the test results of a given setting, with a given `colorType` in QZ.

Master:

| `colorType` | Printer: Black & White | Printer: Color |
|---|---|---|
| Grayscale | Black & White | Black & White |
| Color | Color | Color |
| Null | Color | Color |

This PR:

| `colorType` | Printer: Black & White | Printer: Color |
|---|---|---|
| Grayscale | Black & White | Black & White |
| Color | Color | Color |
| Null | Color | Color |
| Default | Black & White | Color |

Printing from HTML seems to have some issues when the printer is set to a color of 'Black & White', and QZ requests color. HTML also seems to have sticky color settings, requiring a restart after changing. I was able to verify that we are at least **attempting** to set it to the correct value in these scenarios.

### DIFF
--- a/src/qz/printer/action/PrintHTML.java
+++ b/src/qz/printer/action/PrintHTML.java
@@ -192,7 +192,7 @@ public class PrintHTML extends PrintImage implements PrintProcessor {
             settings.setJobName(pxlOpts.getJobName(Constants.HTML_PRINT));
             settings.setPrintQuality(PrintQuality.HIGH);
 
-            if (pxlOpts.getColorType() != null) {
+            if (pxlOpts.getColorType().getAsChromaticity() != null) {
                 settings.setPrintColor(getColor(pxlOpts));
             }
             if (pxlOpts.getDuplex() == Sides.DUPLEX || pxlOpts.getDuplex() == Sides.TWO_SIDED_LONG_EDGE) {
--- a/src/qz/printer/action/PrintHTML.java
+++ b/src/qz/printer/action/PrintHTML.java
@@ -191,6 +191,7 @@ public class PrintHTML extends PrintImage implements PrintProcessor {
             settings.setJobName(pxlOpts.getJobName(Constants.HTML_PRINT));
             settings.setPrintQuality(PrintQuality.HIGH);
 
+            // If colortype is default, leave printColor blank. The system's printer settings will be used instead.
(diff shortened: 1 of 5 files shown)


src/qz/printer/action/PrintHTML.java
```diff
@@ -192,7 +192,7 @@ public void print(PrintOutput output, PrintOptions options) throws PrinterExcept
             settings.setJobName(pxlOpts.getJobName(Constants.HTML_PRINT));
             settings.setPrintQuality(PrintQuality.HIGH);
 
-            if (pxlOpts.getColorType() != null) {
```
Vzor-, line 195: From what I could discern, pxlOpts.getColorType could never return null, so I removed it.

tresf: It's rarely a good idea to remove a null guard.  I can see that the workflow sets an initial value on class construction, then uses a try/catch to set the value, so I agree with this sentiment, but in my tenure of programming it's been very rare that I've removed a NPE guard.

src/qz/printer/action/PrintHTML.java
```diff
@@ -192,7 +192,7 @@ public void print(PrintOutput output, PrintOptions options) throws PrinterExcept
...
+            if (pxlOpts.getColorType().getAsChromaticity() != null) {
```
tresf, line 195: Why wouldn't this be `pxlOpts.getColorType() != DEFAULT`?  That is more readable in my opinion.  It also should have a comment explaining that it will fallback to the default value of the printer, since this behavior doesn't seem to be documented anywhere (Implied by name, but never explicitly stated).

tresf (approved):

PR #1254, opened 2024-04-16
Add "legacy" ESCPOS striping support

tresf:
Prior to QZ Tray 2.0.12, ESCPOS images were sent using a linespacing/newline technique.  Although newer ESCPOS image methods are much more compatible with a wide range of printers, some customers have requested to maintain the old behavior.

This introduces [`dotDensity: "double-legacy"` and `dotDensity: "single-legacy"`](https://qz.io/docs/raw#escpos), which will restore the old behavior.

https://github.com/tresf/tray/releases/download/v2.2.4/qz-tray-2.2.4-DOTDENSITY-x86_64.exe

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -853,6 +853,17 @@ var qz = (function() {
                     }
                 }
 
+                if(_qz.tools.versionCompare(2, 2, 4) < 0) {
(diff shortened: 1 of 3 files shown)


tresf:
Customer states that tests pass.  Merging.

PR #1261, opened 2024-04-27
Added PrintService Retry

Vzor-:
fixes 1259

### DIFF
--- a/src/qz/utils/PrintingUtilities.java
+++ b/src/qz/utils/PrintingUtilities.java
@@ -23,6 +23,7 @@ import qz.ws.PrintSocketClient;
 
 import javax.print.PrintException;
 import java.awt.print.PrinterAbortException;
+import java.awt.print.PrinterException;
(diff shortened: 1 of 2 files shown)


src/qz/utils/PrintingUtilities.java
```diff
@@ -209,7 +210,15 @@ public static void processPrintRequest(Session session, String UID, JSONObject p
             }
 
             processor.parseData(params.getJSONArray("data"), options);
-            processor.print(output, options);
+
+            try {
```
Vzor-, line 214: I don't like the nested try, but I couldn't find a way to query if a PrintService was still valid

tresf: Agreed.  I was able to avoid the nested try, but I haven't tested it.  The code is here: https://github.com/qzind/tray/compare/master...tresf:tray:ps-avail

PR #1263, opened 2024-05-03
Add provisioning support for custom websocket ports

### LINKED ISSUE #1071: Ability to specify custom websocket ports
Currently QZ Tray has its ports hard-coded (e.g. 8181), but certain use-cases (such as embedding it into another application) may prefer this to be changed to allow it to co-exist with QZ Tray proper.

https://github.com/qzind/tray/blob/0e151fb888b6b13f7e4301cc32153ea44107f560/src/qz/common/Constants.java#L108-L110

Note: JavaScript [already allows this](https://qz.io/api/qz.websocket#.connect):
https://github.com/qzind/tray/blob/b6849b20b770e29b618732c317ad4032008a90b5/js/qz-tray.js#L1085-L1088

tresf:
Adds support for specifying custom websocket ports in `qz-tray.properties` as well as `provisioning.json`

`provisioning.json`
```json
[
    {
        "description": "Custom secure websocket ports",
        "type": "property",
        "data": "websocket.secure.ports=9191,9292,9393,9494"
    },
    {
        "description": "Custom insecure websocket ports",
        "type": "property",
        "data": "websocket.insecure.ports=9192,9293,9394,9495"
    }
]
```

```bash
ant -Dprovision.file=provision.json nsis
```

Testing:

1. Install the software, verify everything works on the custom port
   * Visit: https://localhost:9191)
3. Check the firewall rules are installed for the custom port:
   * Start, Run:`wf.msc`, Inbound Rules, "QZ Tray"

Closes #1071


src/qz/installer/provision/ProvisionInstaller.java
```diff
@@ -129,6 +136,31 @@ private boolean invokeStep(Step step) throws Exception {
         return invoker.invoke();
     }
 
+    /**
+     * Loops through steps searching for a property which sets a custom websocket ports
+     */
+    public void setCustomPorts(Installer installer) {
+        for(Step step : steps) {
+            if(step.getType() == Type.PROPERTY) {
+               for(AbstractMap.SimpleEntry<String,String> pair : PropertyInvoker.parsePropertyPairs(step)) {
```
Vzor-, line 145: Map.Entry may be easier to read here

Vzor-: Better yet, instead of doing an ArrayList of entries, a Hashmap<String, String> would probably be better. Then for this part of the code, you do a `map.get(ArgValue.WEBSOCKET_SECURE_PORTS.getMatch())` instead of the for loop.

src/qz/ws/PrintSocketServer.java
```diff
@@ -223,12 +225,37 @@ public static void main(String ... args) {
...
+            fallback = true;
+        }
+        if(fallback) {
+            log.warn("Falling back to default WebSocket ports: ({}), ({})", secure, insecure);
+            secure = Constants.DEFAULT_WSS_PORTS;
+            insecure = Constants.DEFAULT_WS_PORTS;
+        }
+
+        securePorts = Collections.unmodifiableList(Arrays.asList(secure));
+        insecurePorts = Collections.unmodifiableList(Arrays.asList(insecure));
```
Vzor-, line 250: If we look back at the path this array takes, we can see back in `getIntegerArray(ArgValue argValue, Properties ... propsArray)` it is constructed as a list, then converted to an array. Could we just keep it as a List?

src/qz/utils/PrefsSearch.java
```diff
@@ -90,6 +92,26 @@ public static int getInt(ArgValue argValue, Properties ... propsArray) {
         return getInt(argValue, true, propsArray);
     }
 
+    public static Integer[] getIntegerArray(ArgValue argValue, Properties ... propsArray) {
+        List<Integer> parsed = parseIntegerArray(getString(argValue, propsArray));
+        return parsed.toArray(new Integer[parsed.size()]);
+    }
+
+    public static List<Integer> parseIntegerArray(String commaSeparated) {
+        List<Integer> parsed = new ArrayList<>();
```
Vzor-, line 101: Since we know the length this list needs to be, it may be slightly more efficient to pass that size to the array initializer. You would need jungle things around a bit though. If there isn't a clean way to do it, it doesn't really matter. It just prevents the list from over-allocating space.

src/qz/installer/provision/ProvisionInstaller.java
tresf: Agreed, but we should warn if duplicates are found.  Working on it...

tresf: Done via 37b2838702c4bfd73a1f3e90e16ec5bf16794885

src/qz/ws/PrintSocketServer.java
tresf: Well, we still have to construct it due to the Constants being primitive, but done with the larger refactor that was just pushed.

src/qz/utils/PrefsSearch.java
tresf: I don't think we know, the list size (generically) can be any number greater than 1 and less than infinity.  Ports are usually in 4s, but the code doesn't mandate that either.  It's rare I use initial size in code as it has a tradeoff of speed versus versatility.  Happy to change my mind if its warranted though.

Vzor-: I was thinking more along these lines
```java
    public static List<Integer> parseIntegerArray(String commaSeparated) {
        List<Integer> parsed = null;
        try {
            if (commaSeparated != null && !commaSeparated.isEmpty()) {
                String[] split = commaSeparated.split(",");
                parsed = new ArrayList<>(split.length);
                for(String item : split) {
                    parsed.add(Integer.parseInt(item));
                }
            }
        } catch(NumberFormatException nfe) {
            log.warn("Failed parsing {} as a valid integer array", commaSeparated, nfe);
        }
        if (parsed == null) parsed = new ArrayList<>();
        return parsed;
    }
```
But yea, its a trivial optimization and could be omitted.

tresf: Right, but if we ever add sanitization in the future, it can still end up being slightly bigger than we need, which the default size of 20 is already doing for most use-cases.  But pragmatically, the purpose of a linked list is to have dynamic sizing, so unless there's a lot resizing happening, the initial size of 20 is generally sane.  If it were a memory-strict language, I'd be more strict, but there's so much GC happening in Java, I've generally left the slightly wasteful default value of 20 to do it's thing. 🤷‍♂️

Vzor- (approved):

PR #1266, opened 2024-05-14
Improve Linux HiDPI Support

tresf:
image

image

image

image


tresf:
Tested on Fedora 34 200%

image

Issue #1284, opened 2024-07-25
To open "QZ Tray", you need to install Rosetta.  Do you want to install it now?


tresf:
QZ Tray for Apple Silicon installs and starts fine, but when closing and reopening it, it says the following:

<img width="510" alt="image" src="https://github.com/user-attachments/assets/5432312f-0f1f-4778-b70b-54e2c88e711c">

Similar issue:
* https://apple.stackexchange.com/q/452070/147537
* https://apple.stackexchange.com/q/457171/147537

Workaround:

1. Open Terminal
2. "/Applications/QZ Tray.app/Contents/MacOS/QZ Tray"

(or reboot)

The launch script:

https://github.com/qzind/tray/blob/5c78e2e6ef0651c2f7f7bad8d1ab018338fd87ff/ant/unix/unix-launcher.sh.in#L1-L136

PR #1286, opened 2024-08-05
Fix Rossetta Prompt Bug on Apple Silicon

Vzor-:
Closes https://github.com/qzind/tray/issues/1284

This is the better of two fixes for this issue I found. The other is to include a arm64 binary in the Content/MacOS folder, or a symlink thereof. This solution only fixes the issue if this is the first time qz has ran on the system, or you manually edit "QZ Tray.app/Content/info.plist" and change `<key>CFBundleIdentifier</key><string>io.qz.qz-tray</string>` to anything else, for example `<key>CFBundleIdentifier</key><string>io.qz.qz-tray-arm64</string>`.

We need to find a way around this bundleID issue. Something is being cached, and I am unsure of how to clear it.

### DIFF
--- a/ant/apple/apple-bundle.plist.in
+++ b/ant/apple/apple-bundle.plist.in
@@ -20,5 +20,9 @@
                 <key>CFBundleURLSchemes</key>
                 <array><string>${vendor.name}</string></array>
             </dict>
-        </array>
+        </array>        
+    <key>LSArchitecturePriority</key>
+    <array>
+        <string>${apple.target.arch}</string>
+    </array>
 </dict></plist>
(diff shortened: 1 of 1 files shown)


tresf:
> change <key>CFBundleIdentifier</key><string>io.qz.qz-tray</string> to anything else, for example <key>CFBundleIdentifier</key><string>io.qz.qz-tray-arm64</string>.
> 
> We need to find a way around this bundleID issue. Something is being cached, and I am unsure of how to clear it.

This also seems to prevent it from launching automatically after install on my system until this identifier is changed, I'm not sure if that's a hint as to where the problem is.

tresf:
Question posted:

* Stack exchange: https://apple.stackexchange.com/questions/474475/reset-rosetta-flag-for-desktop-app
* Apple developer forums: https://developer.apple.com/forums/thread/761342

PR #1288, opened 2024-08-28
Replaced deprecated URLWhitelist

### LINKED ISSUE #1277: Chrome: qz:launch shows pop-up
At some point, our code to suppress the `qz:launch` pop-ups stopped working.

Initial glance would blame the deprecation of `URLWhitelist`, but changing this to `URLAllowlist` doesn't seem to fix the symptom.
* ⚠️ https://chromeenterprise.google/policies/#URLWhitelist
* ✅ https://chromeenterprise.google/policies/#URLAllowlist

With regards to whether or not the "Always allow" checkbox is shown, there's a policy for that as well, which we don't currently configure:
* https://chromeenterprise.google/policies/#ExternalProtocolDialogShowAlwaysOpenCheckbox

image

Vzor-:
Partially addresses/related to #1277

### DIFF
--- a/src/qz/installer/LinuxInstaller.java
+++ b/src/qz/installer/LinuxInstaller.java
@@ -27,7 +27,7 @@ public class LinuxInstaller extends Installer {
     public static final String APP_LAUNCHER = APP_DIR + SHORTCUT_NAME;
     public static final String UDEV_RULES = "/lib/udev/rules.d/99-udev-override.rules";
     public static final String[] CHROME_POLICY_DIRS = {"/etc/chromium/policies/managed", "/etc/opt/chrome/policies/managed" };
-    public static final String CHROME_POLICY = "{ \"URLWhitelist\": [\"" + DATA_DIR + "://*\"] }";
+    public static final String CHROME_POLICY = "{ \"URLAllowlist\": [\"" + DATA_DIR + "://*\"] }";
 
     private String destination = "/opt/" + PROPS_FILE;
     private String sudoer;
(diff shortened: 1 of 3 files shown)


tresf:
> Fixes #1277

What about Windows, Linux?

tresf:
I tested this on Windows 11 ARM64 and the popup still shows.

image

tresf (changes requested):
Changes are needed; seems to still be broken on Windows.

Vzor-:
This is a **working** policy on MacOS
Screenshot 2024-08-29 at 2 04 01 PM
this is a **not working** policy on Windows. 
Screenshot 2024-08-29 at 2 03 14 PM
They seem similar, but there are some minor discrepancies. I'll try resolving these to find the issue.

tresf:
@Vzor- based on our conversations, I believe you found an upstream bug here but this bug report was never updated to reflect it.  Furthemore, I do not believe that an upstream bug report was ever filed to accompany this downstream bug.  Can you please provide your findings so that we can take action on this?

Vzor-:
Certainly. I was in the middle of testing a possible solution when I had to jump to a more pressing issue. This will be the next issue I start.

tresf:
> Certainly. I was in the middle of testing a possible solution when I had to jump to a more pressing issue. This will be the next issue I start.

It's been a month and the upstream bug/upstream bug report still isn't mentioned here or in #1277.  For sake of completeness, I'm going to change the wording of this PR from "fixes" to "related", merge and paste your screenshot over there.  Please provide the details that you gathered researching this problem to #1277 so that we can determine next actions.

tresf (approved):

Issue #1289, opened 2024-09-03
Site Manager: Show timezone for cert expiration


tresf:
Date calculations are in UTC, but this is not obvious.

Dates should show in UTC by default and if clicked can switch to the local timezone.

<img width="488" alt="image" src="https://github.com/user-attachments/assets/5b7af2bf-003c-4050-a72c-fc06e410d869">

PR #1292, opened 2024-09-09
Add Timezone to Cert

Vzor-:
Closes https://github.com/qzind/tray/issues/1289

~~WIP~~

Copy support soon to come in a separate PR. 
Old:
Screenshot 2024-09-08 at 11 37 34 PM

New:
Screenshot 2024-09-08 at 11 34 38 PM
Screenshot 2024-09-08 at 11 34 47 PM


Vzor-:
It is not at all obvious that the date entry is intractable, but I could not find a clean way to make it apparent. Due to how tables are implemented, we can't even do the hand-cursor to show it is interactable without a bunch of hit-detection code.

tresf:
When switching from EST to UTC, the selected row loses focus.  This should be fixed.

Also, the selection affects rows not clicked, which makes sense, but is out of context.

tresf:
> When switching from EST to UTC, the selected row loses focus. This should be fixed.
> 
> Also, the selection affects rows not clicked, which makes sense, but is out of context.

Both of these are addressed in the latest code changes, thanks.  After a code review, I'm concerned about the use of `description` in the `Certificate.java` class.  I think the enum is an improvement over hard-coded string values, but I think these descriptions are better suited for a UI-related class.

src/qz/ui/component/CertificateTable.java
```diff
@@ -126,7 +170,22 @@ public Component getTableCellRendererComponent(JTable table, Object value, boole
 
             // First Column
             if (value instanceof CertificateField) {
-                label = stylizeLabel(STATUS_NORMAL, label, isSelected);
```
tresf, line 129: @Vzor- do you mind explaining these changes?  At a glance I would not expect the first column logic to change as a result of this PR.  Perhaps it was due to some unreverted changes, but I would expect this logic to be largely untouched.

tresf:
@Vzor- thanks for the simplification.  I've made some minor refactorings to simplify the timezone flag by stashing it in the enum field.  This will have the side-effect of being remembered between reloading of certs into the table.

Issue #1295, opened 2024-09-30
Image prints across several pages with scaleContent disabled


tresf:
Printing this image: 8.5x11:

https://github.com/user-attachments/assets/3cffd93e-c1d8-40a3-a7bc-f48b92e9f288

* ✅ (PASSS) Works when `scaleContent: true`
* ⛔ (FAIL) Splits across many pages when `scaleContent: false`.

I believe this bug was introduced by https://github.com/qzind/tray/commit/8024139b24c94006706d785163b2f7d24eeabe13 ~5 years ago and never noticed in QA testing.

PR #1296, opened 2024-10-02
Adjust image splitting to account for page dpi

akberenz:
For #1295


tresf:
Tests pass.  Will check with customer to confirm.

tresf (approved):

tresf:
Customer confirms that the patch fixes the issue.  Merging.

PR #1299, opened 2024-10-21
Added ant dependency for intelij

### LINKED ISSUE #1298: Intelij Ant Plugin Missing
As of Intelij 2023.3, the Ant plugin is no longer bundled and needs to be downloaded through the marketplace. https://blog.jetbrains.com/idea/2023/12/intellij-idea-2023-3/#other This should be reflected in the wiki, or maybe we can somehow include it as a project dependency so intelij will automatically download it.

Vzor-:
Fixes #1298

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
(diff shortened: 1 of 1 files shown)


Vzor-:
Screenshot_20241021_160327-1

Vzor-:
After installing the plugin, I had to close and re-open the project before the download-javafx target worked.

Issue #1300, opened 2024-10-29
Logs spammed with ClosedChannelException


tresf:
A client has logs files with these messages spammed hundreds of times.

I believe this is a sign that the websocket has closed, but the listener for USB has not.

```
2024-10-25T13:04:50,416 [ERROR] Could not send message
java.io.IOException: java.nio.channels.ClosedChannelException
	at org.eclipse.jetty.util.FutureCallback.block(FutureCallback.java:163) ~[qz-tray.jar:?]
	at org.eclipse.jetty.util.FutureCallback.block(FutureCallback.java:139) ~[qz-tray.jar:?]
	at org.eclipse.jetty.websocket.common.JettyWebSocketRemoteEndpoint.sendBlocking(JettyWebSocketRemoteEndpoint.java:191) ~[qz-tray.jar:?]
	at org.eclipse.jetty.websocket.common.JettyWebSocketRemoteEndpoint.sendString(JettyWebSocketRemoteEndpoint.java:52) ~[qz-tray.jar:?]
	at qz.ws.PrintSocketClient.send(PrintSocketClient.java:803) [qz-tray.jar:?]
	at qz.ws.PrintSocketClient.sendStream(PrintSocketClient.java:788) [qz-tray.jar:?]
	at qz.utils.UsbUtilities$1.run(UsbUtilities.java:209) [qz-tray.jar:?]
Caused by: java.nio.channels.ClosedChannelException
	at org.eclipse.jetty.websocket.core.internal.WebSocketSessionState.onOutgoingFrame(WebSocketSessionState.java:184) ~[qz-tray.jar:?]
	at org.eclipse.jetty.websocket.core.internal.WebSocketCoreSession.sendFrame(WebSocketCoreSession.java:514) ~[qz-tray.jar:?]
	at org.eclipse.jetty.websocket.common.JettyWebSocketRemoteEndpoint.sendBlocking(JettyWebSocketRemoteEndpoint.java:190) ~[qz-tray.jar:?]
	... 4 more
```

tresf:
Closed via #1304

PR #1304, opened 2024-11-02
Better handling of closed stream errors

akberenz:
Addresses the issue in #1300 where streams will continue to attempt sending data to a closed stream until explicitly closed.

### DIFF
--- a/src/qz/communication/UsbIO.java
+++ b/src/qz/communication/UsbIO.java
@@ -1,21 +1,25 @@
 package qz.communication;
 
-import qz.utils.UsbUtilities;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import qz.ws.SocketConnection;
 
 import javax.usb.*;
 import javax.usb.util.UsbUtil;
 
 public class UsbIO implements DeviceIO {
-
+    private static final Logger log = LogManager.getLogger(UsbIO.class);
     private UsbDevice device;
     private UsbInterface iface;
 
     private boolean streaming;
 
+    private DeviceOptions dOpts;
+    private SocketConnection websocket;
 
-    public UsbIO(DeviceOptions dOpts) throws DeviceException {
-        UsbDevice device = UsbUtilities.findDevice(dOpts.getVendorId().shortValue(), dOpts.getProductId().shortValue());
-
+    public UsbIO(DeviceOptions dOpts, SocketConnection websocket) throws DeviceException {
+        this.dOpts = dOpts;
+        this.websocket = websocket;
         if (device == null) {
             throw new DeviceException("USB device could not be found");
         }
@@ -129,16 +133,18 @@ public class UsbIO implements DeviceIO {
         }
     }
 
-    public void close() throws DeviceException {
+    public void close() {
+        streaming = false;
+        // Remove orphaned listeners
+        websocket.removeDevice(dOpts);
         if (iface.isClaimed()) {
             try {
                 iface.release();
             }
             catch(UsbException e) {
-                throw new DeviceException(e);
+                log.error("Unable to close USB device", e);
             }
         }
-        streaming = false;
     }
 
 }
--- a/src/qz/communication/UsbIO.java
+++ b/src/qz/communication/UsbIO.java
@@ -133,6 +133,7 @@ public class UsbIO implements DeviceIO {
         }
     }
 
+    @Override
     public void close() {
         streaming = false;
         // Remove orphaned listeners
--- a/src/qz/communication/UsbIO.java
+++ b/src/qz/communication/UsbIO.java
@@ -2,6 +2,7 @@ package qz.communication;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import qz.utils.UsbUtilities;
 import qz.ws.SocketConnection;
 
 import javax.usb.*;
@@ -20,6 +21,7 @@ public class UsbIO implements DeviceIO {
     public UsbIO(DeviceOptions dOpts, SocketConnection websocket) throws DeviceException {
         this.dOpts = dOpts;
         this.websocket = websocket;
+        UsbDevice device = UsbUtilities.findDevice(dOpts.getVendorId().shortValue(), dOpts.getProductId().shortValue());
         if (device == null) {
             throw new DeviceException("USB device could not be found");
         }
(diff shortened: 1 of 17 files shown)


src/qz/communication/UsbIO.java
```diff
@@ -1,21 +1,25 @@
...
     private UsbDevice device;
     private UsbInterface iface;
 
     private boolean streaming;
 
+    private DeviceOptions dOpts;
+    private SocketConnection websocket;
 
-    public UsbIO(DeviceOptions dOpts) throws DeviceException {
-        UsbDevice device = UsbUtilities.findDevice(dOpts.getVendorId().shortValue(), dOpts.getProductId().shortValue());
```
akberenz, line 17: It looks like this line was removed when the local param assignments were added, but it should have been kept in alongside those.

src/qz/printer/status/StatusMonitor.java
```diff
@@ -204,9 +207,10 @@ private synchronized static boolean dispatchStatusEvent() {
             // And find every client that subscribed to all printers
             listeningConnections.addAll(clientPrinterConnections.get(ALL_PRINTERS));
         }
-        for (SocketConnection connection : listeningConnections) {
-            statusSessions.get(connection).statusChanged(status);
-        }
+
+        for(SocketConnection connection : listeningConnections)
+            statusSessions.get(connection).statusChanged(status, () -> listeningConnections.remove(connection));
```
akberenz, line 212: `listeningConnections` is a local hash built each call, so shouldn't this be removing the connection from the source list under the `clientPrinterConnections` instead?

And minor nitpick, but can we add the brackets back to this loop?

src/qz/communication/UsbIO.java
tresf: Whoa, good catch!  Fixing. :)

tresf: Fixed via 04402bc.

src/qz/printer/status/StatusMonitor.java
tresf: > And minor nitpick, but can we add the brackets back to this loop?

Whoops.... I thought I was helping the diff by removing them but I clearly misread the original code.

With regards to the variable scope, I think it's identical to your commit 1183887363bbe92b42adddc64b40ed8b22ca2766 so I kept the same scope.  I'll study the code to find out what it's actually doing....

tresf: I think this can be resolved, please mark it if so.  I'll also ping @Vzor- for a second set of eyes.

Vzor-: If the intention is to stop listening to the specific printer, `clientPrinterConnections` is where that should be added. https://github.com/qzind/tray/commit/326f2960b6eb6c78deeac5966cfeedaedad73faa will not have the intended effect, it will simply throw a NPE during the next status. I will write a quick example of what it should look like.

tresf: > If the intention is to stop listening to the specific printer, clientPrinterConnections is where that should be done. https://github.com/qzind/tray/commit/326f2960b6eb6c78deeac5966cfeedaedad73faa will not have the intended effect, it will simply throw a NPE during the next status. I will write a quick example of what it should look like.

It's not the intention.  The intention is to stop attempting to send data to that websocket.

tresf: Done via bde79b5.

src/qz/utils/UsbUtilities.java
```diff
@@ -206,22 +206,22 @@ public void run() {
...
                                 try { Thread.sleep(interval); } catch(Exception ignore) {}
                             }
                         }
-                        catch(WebSocketException e) {
-                            usb.setStreaming(false);
-                            log.error("USB stream error", e);
-                        }
-                        catch(DeviceException e) {
+                        catch(WebSocketException | DeviceException e) {
+                            // Calling usb.close() can cause a hard-crash on macOS
```
tresf, line 215: @Vzor- I thought we had gracefully handled this in our upstream PR, but somehow I'm able to crash the app here if I call `usb.close()`.  https://github.com/gary-rowe/hid4java/pull/153/files

I'm leaving this comment in the source code for now as a future warning.

Crashing it's pretty easy, just close the tab while the device is streaming.

tresf:
Testing this... sadly, the `ClosedChannelException` is being wrapped in an `IOException` which is causing the explicit catches to not fire.  Investigating... 

```
[ERROR] 2024-12-03T13:12:37,022 @ qz.ws.PrintSocketClient:834
	Could not send message
java.io.IOException: java.nio.channels.ClosedChannelException
	at org.eclipse.jetty.util.FutureCallback.block(FutureCallback.java:163) ~[jetty-util-10.0.22.jar:10.0.22]
	at org.eclipse.jetty.util.FutureCallback.block(FutureCallback.java:139) ~[jetty-util-10.0.22.jar:10.0.22]
	at org.eclipse.jetty.websocket.common.JettyWebSocketRemoteEndpoint.sendBlocking(JettyWebSocketRemoteEndpoint.java:191) ~[websocket-jetty-common-10.0.22.jar:10.0.22]
	at org.eclipse.jetty.websocket.common.JettyWebSocketRemoteEndpoint.sendString(JettyWebSocketRemoteEndpoint.java:52) ~[websocket-jetty-common-10.0.22.jar:10.0.22]
	at qz.ws.PrintSocketClient.send(PrintSocketClient.java:828) ~[production/:?]
	at qz.ws.PrintSocketClient.sendError(PrintSocketClient.java:768) ~[production/:?]
	at qz.ws.PrintSocketClient.processMessage(PrintSocketClient.java:375) ~[production/:?]
	at qz.ws.PrintSocketClient.lambda$onMessage$0(PrintSocketClient.java:171) ~[production/:?]
	at java.base/java.lang.Thread.run(Thread.java:829) [?:?]
Caused by: java.nio.channels.ClosedChannelException
	at org.eclipse.jetty.websocket.core.internal.WebSocketSessionState.onOutgoingFrame(WebSocketSessionState.java:184) ~[websocket-core-common-10.0.22.jar:10.0.22]
	at org.eclipse.jetty.websocket.core.internal.WebSocketCoreSession.sendFrame(WebSocketCoreSession.java:514) ~[websocket-core-common-10.0.22.jar:10.0.22]
	at org.eclipse.jetty.websocket.common.JettyWebSocketRemoteEndpoint.sendBlocking(JettyWebSocketRemoteEndpoint.java:190) ~[websocket-jetty-common-10.0.22.jar:10.0.22]
	... 6 more
```

tresf:
> Testing this... sadly, the `ClosedChannelException` is being wrapped in an `IOException` which is causing the explicit catches to not fire. Investigating...

I think this is fixed via f0e3754.

PR #1309, opened 2024-12-13
Fix qz:steal on macOS

### LINKED ISSUE #850: qz:steal is not supported on MacOS
More information here:
https://stackoverflow.com/questions/27435606

Quoting the accepted answer:
> Note that your URI handler will be called in the currently running instance of you app or will first create a new instance of your app then get called. The OpenURIEvent will contain the entire URI message you send.

This means we can capture the call to `qz:steal` via:

```java
// app.setOpenURIHandler(this);

@Override
public void openURI(final com.apple.eawt.AppEvent.OpenURIEvent oue) {
    // perform action
}
```

... however this event is proprietary to Apple, the code won't compile on Windows or Linux, requiring some [elaborate proxying](https://github.com/qzind/tray/blob/802d1f8f2d47b7c36d9f6c1187a117d0170f6567/src/com/apple/OSXAdapter.java).

Fortunately, Java 9 offers an API to handle this gracefully, we do this for some of our other proprietary Apple APIs, but getting both of these tested requires Custom URL registration between trial and error attempts (i.e. a full `ant pkgbuild` and watching the logs between each change)

Another caveat is that this technique causes QZ Tray to start asynchronously of it's startup handling, possibly causing a race between `SingleInstanceChecker` and `PrintSocketServer`, so safe, synchronized techniques must be used.

tresf:
Fixes #850 

Initial attempt is not working, nothing fires... Possible name conflict with `com.apple`?

### DIFF
--- a/src/qz/common/TrayManager.java
+++ b/src/qz/common/TrayManager.java
@@ -349,6 +349,7 @@ public class TrayManager {
         componentList.add(aboutDialog);
 
         if (SystemUtilities.isMac()) {
+            MacUtilities.registerUriHandler();
             MacUtilities.registerAboutDialog(aboutDialog);
             MacUtilities.registerQuitHandler(this);
         }
--- a/src/qz/utils/MacUtilities.java
+++ b/src/qz/utils/MacUtilities.java
@@ -26,6 +26,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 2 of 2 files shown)


tresf:
Possibly related:
https://stackoverflow.com/questions/20074610
https://stackoverflow.com/questions/76522890
https://stackoverflow.com/questions/74744794
https://stackoverflow.com/questions/27435606
https://stackoverflow.com/questions/51167654

tresf:
Opened a question with our JDK support provider, QZ-24

tresf:
This issue is not reproducible in a minimal project, so we'll have to dive deeper as to the cause... 

[OpenUriTest.zip](https://github.com/user-attachments/files/20233941/OpenUriTest.zip)

tresf:
Closing since this issue is being handled upstream.

PR #1380, opened 2025-11-12
Provision ingestion: Re-use identical resources instead of copying

tresf:


tresf:
@lukas-w and I discussed this and he brought up a great point about using hardlinks instead of copying files.

Testing this locally is nice:
* Hard-linking doesn't take up additional file space
* Hard-linking allows two identical files to be compared without using costly byte-for-byte comparisons

However, I have some greater concerns for this approach... 
* When developing  through trial-and-error locally things such as scripts, it may not be immediately obvious to a developer why `dist/provisioning` contains versions of scripts that haven't been ingested yet.
* Deleting the original file will orphan the hardlink, which may cause confusion

That said, these risks are mostly limited to local dev instances; automated builds can benefit from this; hardlinks should be considered in the future if disk usage or long file compares becomes a cause of concern.

tresf:
Testing:
* ✅ Identical 14 GB file - PASS, saves only once to `dist/provisioning`
* ✅ Different files with identical names - PASS, saves with appended name to `dist/file.exe` and `dist/file-1.exe` respectively.

tresf:
> we have some control around where the files are copied to, so we can likely write our own avoidance. I'll see what I can whip up.

PR opened here: https://github.com/qzind/tray/pull/1381

Will test on Linux with similarly named files to see how well it works.

PR #1425, opened 2026-03-12
Add support for BellSoft API downloads

tresf:
Adds support for BellSoft API authentication which can download a JDK that's been build specifically for a BellSoft customer (such as QZ Industries, LLC).

```properties
# Fetch a copy of the JDK from BellSoft's support portal
jlink.api.enabled=false
jlink.api.url=https://api.bell-sw.com/v1/liberica/releases
# jlink.api.token=
# jlink.api.exact=false <-- false: use latest found version; true: match jlink.java.version exactly
```

TODO:
* 🚫 REDACTED: ~~Determine how to provide these values to `ant` from our Company Branded portal~~<br>**Edit:**: We'll just keep this as a local-only tool.


src/qz/build/JLink.java
```diff
@@ -141,19 +154,130 @@ private static boolean needsDownload(Version want, Version installed) {
...
+    /**
+     * Grabs a <code>URL,Version</code> pair from an API call
+     * TODO: Make this more vendor-agnostic
+     */
+    private Pair<URL,Version> getApiUrl(Arch arch, Platform platform, HashMap<String,String> headers) throws IOException {
+        String apiUrl = System.getProperty("jlink.api.url");
+        boolean exact = Boolean.parseBoolean(System.getProperty("jlink.api.exact")); // defaults to false
+
+        Version bestVersion = Version.of(0);
+        URL bestUrl = null;
```
Vzor-, line 185: If i'm reading this right, `besturl` can be null if there was no appurl included, but later on we call `apiPair.getKey().toString();` I think this will lead to a NPE

src/qz/build/Fetcher.java
```diff
@@ -50,55 +61,87 @@ public static void main(String ... args) throws IOException {
...
+        HttpRequest.Builder requestBuilder;
+        try {
+            requestBuilder = HttpRequest.newBuilder().uri(url.toURI()).GET();
+            if(headers != null) {
+                headers.forEach(requestBuilder::header);
+            }
+            HttpRequest request = requestBuilder.build();
+
+            // stream response to file
+            HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofFile(
```
Vzor-, line 125: I think this should have some basic sanitation. I think if we 404, it will write that error to the disk.

src/qz/build/JLink.java
```diff
@@ -141,19 +154,130 @@ private static boolean needsDownload(Version want, Version installed) {
...
+                        } else {
+                            if (versionFound != null) {
+                                Version foundVersion = JavaVersion.parse(versionFound);
+                                if (foundVersion.isHigherThan(bestVersion)) {
+                                    bestVersion = foundVersion;
+                                    bestUrl = new URL(downloadUrl);
+                                } else if (foundVersion.equals(bestVersion)) {
+                                    // handle '+nnn'
+                                    if (foundVersion.buildMetadata().isPresent() && bestVersion.buildMetadata().isPresent()) {
+                                        if (foundVersion.buildMetadata().get().compareTo(bestVersion.buildMetadata().get()) > 0) {
```
Vzor-: I think `compareTo` is wrong here, it will sort the strings alphabetically, so 9 < 10.

src/qz/build/JLink.java
```diff
@@ -125,13 +134,17 @@ public static void main(String ... args) throws IOException {
     /**
      * Handle incompatibilities between JDKs, download a fresh one if needed
      */
-    private static boolean needsDownload(Version want, Version installed) {
+    private boolean needsDownload(Version want, Version installed) {
+        if(apiEnabled) {
+            // always assume api downloads differ from the local version
+            return true;
```
Vzor-, line 140: Does this mean a new jdk is downloaded every build?

src/qz/build/JLink.java
tresf: Yeah if there are no downloads at all, it'll throw an NPE.   Can probably explicitly throw an IOException in that case with some type of usable message.

src/qz/build/JLink.java
tresf: It's build metadata -- a string value according to semver -- compareTo is the proper way to compare two strings.

I suppose we could try to write our own crude parser for common scenarios...

At time of writing this, `25.0.3+11` is what BellSoft is offering....  `25.0.3+9` is what Adoptium is offering, so this certainly has a potential to grab the non-latest build.  I suppose if the first build metadata slot is a singular value that can be parsed as an integer, we can attempt to compare it to others as an integer as well.

src/qz/build/Fetcher.java
tresf: Yeah, I tested 404s and it does download the file, then pukes on parsing it as a tarball, which was perfectly readable in the logs.  Advice welcome, but this is an internal-use tool, it's us reading these errors, not customers.  I found the existing behavior to be palatable.

src/qz/build/JLink.java
tresf: Not necessarily, `Fetcher` checks for existence to prevent downloading one that's already there.

```java
log.info("Resource '{}' from [{}] has already been downloaded and extracted.  Using: [{}]", resourceName, url, extracted);
```

`needsDownload` was initially intended to help detect if we're to download the host jdk AND the target jdk.

The reason stems from Java's promise to be able to use jlink from all 11.x versions to package all 11.x versions... so we could technically skip the download in edge-cases, such as building for a foreign platform when Java 11 was already being used to compile.  Since running the same major version as our project uses is becoming increasingly uncommon (most of my systems are on JDK25 now), this function can probably just be removed and we assume that we need an exact copy of the JDK for the host and the target every time we build.  Furthermore, OpenJDK broke this promise \*\***cough JDK-8240734**\*\* once already as can be seen by the workarounds in the source code. 😞

src/qz/build/JLink.java
tresf: Guarded via 13916da.

src/qz/build/JLink.java
tresf: As it turns out, `Version.compareTo` handles build metadata, so this should be possible without a bunch of ugly parsing logic.

tresf: Switched to `Version.compareTo` via 4ee5b2f.

Vzor- (approved):
Looks good

PR #1438, opened 2026-04-13
Add detailed printer crawling

### LINKED ISSUE #830: Printer crawling should be more verbose
Today @lite1979 and I spent a considerable amount of time on a call getting IntelliJ setup just to see which printers were hanging QZ Tray.

The printer crawling can be expensive and I believe logging at least the printer name will help allude to the problem.  In the case of this computer, there were two offending printer mappings causing QZ Tray to deadlock.

tresf:
Closes #830


Vzor- (approved):
Looks great!

PR #1441, opened 2026-04-15
Add LocalNetworkAccess detection

### LINKED ISSUE #1382: Detect permission state of Local Network Access setting
Upstream question:
* https://github.com/WICG/local-network-access/issues/69

We need a way to programmatically detect the presence of the Local Network Access (LNA) setting to better inform users when it's preventing a connection from succeeding.

```js
navigator.permissions.query({ name: 'local-network-access'})
    .then(result => console.log("LNA permission state: ", result))
    .catch(err => console.error(err));
```

Firefox 147 Nightly: 

```
TypeError: 'local-network-access' (value of 'name' member of PermissionDescriptor) is not a valid value for enumeration PermissionName.
    <anonymous> debugger eval code:1
```

Safari 18.6:

```
TypeError: Type error
```

TODO:
* Find a way to detect this reliably and add to `qz-tray.js`.

tresf:
Allow qz-tray.js to notify in a usable error when LocalNetworkAccess restrictions prevent a connection.

Detailed write-up here: https://qz.io/docs/lna

Closes #1382


sample.html
```diff
@@ -8,6 +8,9 @@
 <!-- Required scripts -->
 <script type="text/javascript" src="js/qz-tray.js"></script>
 
+<!-- LNA Detection -->
+<script type="text/javascript" src="js/sample/lna.umd.cjs"></script>
```
akberenz: I don't see this file added in the PR, did it get missed?

lukas-w: I was just hesitant to commit a bundled/minified artifact, especially while it's under active development.

So in short: Yes :grin:

js/qz-tray.js
```diff
@@ -157,67 +174,39 @@ var qz = (function() {
...
-                    catch(err) {
-                        _qz.log.error(err);
-                        deeper();
-                        return;
-                    }
-
-                    if (_qz.websocket.connection != null) {
-                        _qz.websocket.connection.established = false;
+                    var promise = _qz.websocket.setup.connectToAddress(address);
+                    _qz.websocket.connection = null;
```
akberenz: is this assignment to null here needed? as it looks like this only gets set now if the promise resolves successfully

lukas-w: I'm not entirely sure. I was trying to mirror the previous logic where `_qz.websocket.connection.established` would be set to `false` at the start of the connection attempt. Thoughts welcome.

lukas-w: Looking at this more, I think the new behavior of only setting `_qz.websocket.connection` after the connection is opened breaks some assumptions in other parts of the code. Specifically, `qz.websocket.connect` uses `_qz.websocket.connection` to detect an ongoing connection attempt, which my changes break.

I'll see if it's feasible to bring back the original `_qz.websocket.connection`, though that may require changes in the lna library. It may be easier to add something like a `connectionPromise` state instead.

akberenz: not opposed to some sort of "connecting state" variable being added in for that. We don't have anything like that already since we could just query that websocket.connection value directly.

lukas-w: I ended up restoring the previous behavior via e2ec587276445e9c0014d2667f954d1c036db1f9 so that `_qz.websocket.connection` is available during the connection attempt. Less changes overall and this new solution can actually keep lna's API simpler.

js/qz-tray.js
```diff
@@ -77,6 +77,54 @@ var qz = (function() {
...
+                            resolve(ws);
+                        };
+                        onError = function(e) {
+                            _qz.websocket.connection = null;
+                            reject(e);
+                        };
+                        ws.addEventListener("open", onOpen);
+                        ws.addEventListener("error", onError);
+                        // Older Safari versions may trigger close event instead of error event.
+                        ws.addEventListener("close", onError);
```
tresf, line 122: Neat!  Curious... what's preventing other browsers from calling `onError()` when close is called?  Is it because this is only part of the connection promise, so the close can essentially be assumed to be an error in all cases?

I really like the event listener usage.  I'll test to make sure this syntax is OK with IE.

lukas-w:
I added an optional `require('lna')` via 610b2e68950925389431a14877f0e9476ed6cbed but I'm not 100% sure it's kosher and will work with most bundlers. Only tested with Vite 7 which uses Rollup AFAIK. Would love a second set of eyes from someone who's more familiar with the various JS module systems.

js/qz-tray.js
tresf: > I'll test to make sure this syntax is OK with IE.

Branch was behind, so had to FF it to get over some January patches...  Hope it doesn't mess you up @lukas-w.    Is deployment to https://qzind.github.io/lna/ automatic?

tresf:
> I added an optional `require('lna')` via [610b2e6](https://github.com/qzind/tray/commit/610b2e68950925389431a14877f0e9476ed6cbed) but I'm not 100% sure it's kosher and will work with most bundlers. Only tested with Vite 7. Would love a second set of eyes from someone who's more familiar with the various JS module systems.

Hmm... I had to do something similar here... https://github.com/qzind/tray/blob/b0d004d9cbf8eec7da0ebab628036a9d4f35f2f1/js/qz-tray.js#L785  Slightly different detection techniques used, should probably match or be consolidated?

js/qz-tray.js
lukas-w: > what's preventing other browsers from calling `onError()` when close is called?

It's also called in other browsers, though not after the connection was opened. In the `.finally` function below, the event listeners are removed after the promise is resolved, so immediately after any event out of the three `open` / `error` / `close` is fired. The rationale (as described in the commit message of 41997e8ea67dd15ec5acfb4f95d01a948a8260bc) of removing the check for Safari is that a `close` event before an `open` event can - though it's in violation of the spec - probably always be considered a connection failure.

I figured this is a safe simplification while potentially catching some other offenders in case Safari is not the only one.

> I really like the event listener usage. I'll test to make sure this syntax is OK with IE.

Should be fine in IE9+ according to https://caniuse.com/addeventlistener.

lukas-w:
> Slightly different detection techniques used, should probably match or be consolidated?

You mean `typeof exports === 'object'` vs `typeof require === 'function'`? I believe either is fine, but I'll have to admit that I don't really know enough about this. The only thing I can find that's close to a standard way of doing this is the UMD patterns which use `typeof exports === 'object' && typeof exports.nodeName !== 'string'` (e.g. in https://github.com/umdjs/umd/blob/master/templates/commonjsStrict.js). ~~I have no idea what the extra `exports.nodeName` check is for.~~

_EDIT:_ `exports.nodeName` prevents misdetection when there's an HTML element with id `exports`, since elements with ids are globally available by their ID which is crazy and which I did absolutely not know.  https://github.com/umdjs/umd/issues/35

lukas-w:
Off-topic, but that line you quoted looks buggy. The result of `resolve(loc)` isn't used:

https://github.com/qzind/tray/blob/b0d004d9cbf8eec7da0ebab628036a9d4f35f2f1/js/qz-tray.js#L783-L787

js/qz-tray.js
lukas-w: > Is deployment to https://qzind.github.io/lna/ automatic?

It's automatic, but from https://github.com/qzind/lna/tree/tray-demo. I'll update it.

tresf:
> I have no idea what the extra `exports.nodeName` check is for.

IE-ism apparently lol... `<span id="exports"/>` `typeof exports` `Object` 🤢

lukas-w:
> IE-ism apparently lol

This actually appears to be defined by the HTML5 spec :eyes:  https://html.spec.whatwg.org/#named-access-on-the-window-object

tresf:
> > IE-ism apparently lol
> 
> This actually appears to be defined by the HTML5 spec 👀 https://html.spec.whatwg.org/#named-access-on-the-window-object

Well, that is a relief.  This cursed behavior is not a bug, but a feature! :D

tresf:
I feel like we can adopt this as a `_qz.tools` candidate for both scenarios... let me know if you want me to tackle this.  I'll assume you will for now and wait for instructions otherwise :D

js/qz-tray.js
tresf: > > Is deployment to https://qzind.github.io/lna/ automatic?
> 
> It's automatic, but from https://github.com/qzind/lna/tree/tray-demo. I'll update it.

Ok, the sample.html copy was outdated as well.  Updated.  Testing is passing in IE11 now.

lukas-w:
> I feel like we can adopt this as a `_qz.tools` candidate for both scenarios

I don't think there's much that can be extracted / consolidated here. A helper function to load dependencies would mean calling `require` with dynamic arguments, which breaks bundlers like Webpack that use static analysis to find `require` calls and identify the dependencies to include.

Also rethinking this, I believe having different checks before calling `require` can actually make sense here since they serve different use-cases:

- `require('path')` is only for node.js environments, so the `typeof exports === 'object'` is probably suitable here
- `require('lna')` in contrast shouldn't run in node.js at all -- the library is only useful in browser environments. So the call is mostly there for bundlers to pick up the requirement.

I don't know how technologies like AMD or in-browser module loaders like require.js play into this or if they're even relevant anymore. @akberenz do you happen to have any insight on whether 610b2e68950925389431a14877f0e9476ed6cbed is an adequate way of including `lna` as optional dependency?

tresf:
> I don't think there's much that can be extracted / consolidated here. A helper function to load dependencies would mean calling require with dynamic arguments, which breaks bundlers like Webpack that use static analysis to find require calls and identify the dependencies to include.

No, I meant the module detection specifically.

> require('lna') in contrast shouldn't run in node.js at all 

Good point although I'm still slightly concerned that there are certain scenarios (AOT compilers) where this position may not be accurate.  This is sort of the predicament we got into with `path()` [here](https://qz.io/docs/api-overrides#angularjs-quickstart).  The various web bundlers for JS are quite voodoo to me because I don't know when or where the JS is to be interpreted and compiled.  In the case of `path()` it's a Node built-in (and people can use the `path` dependency if needed), but now that we're introducing LNA, the code path is a bit more murky, so I have to lean on you for advice here.

> require('path') is only for node.js environments, so the typeof exports === 'object' is probably suitable here

Right, but only due to the `window`|`document` guard, but to your point, they're different scenarios we're guarding for, so (I'm pretty confident that) you're right to say we need to keep them separate.   We can re-utilit-ize them in the future when we have a second use-case for either/each.

I just see this happening again in the future, but we can cross that road when we get to it.  I agree, the use-cases are likely too dissimilar to attempt to consolidate.

akberenz:
> @akberenz do you happen to have any insight on whether https://github.com/qzind/tray/commit/610b2e68950925389431a14877f0e9476ed6cbed is an adequate way of including lna as optional dependency?

From what I can tell from things like webpack's docs, this should be acceptable.
Though we may want a comment around the `if(typeof require` bit to help denote your reasoning for using that instead of "exports" when we come across this code in the future.

tresf:
@lukas-w can you please bump the code at https://qzind.github.io/lna/ too so that I can test and approve?

lukas-w:
Sure, done. Sorry for the delay!

tresf (commented):
Approving via comment (since it's my own PR).

PR #1448, opened 2026-05-06
Fix HTML snapshot height calculation

### LINKED ISSUE #1447: Height "snapshot" unpredictable when printing HTML
HTML `{ rasterize: true }` content prints erratically (unusually tall) when the height is not specified.
> Example:
> 
> Notice that the document above is **much** taller than the content.

### Background
HTML printing takes a two-pass approach when the height of a document is not specified.
* First pass, to determine the "natural size"
   * Use JavaScript on the DOM to calculate the height
 * Second pass to set the height

The height is calculated using the following equation:
```java
    private static double findHeight() {
        String heightText = webView.getEngine().executeScript("Math.max(document.body.offsetHeight, document.body.scrollHeight)").toString();
        return Double.parseDouble(heightText);
    }
```

### Steps

The fastest way to reproduce is to run the `WebAppTest` class.
1. With IntelliJ IDEA, git, Java 25 and `ant` installed, clone the `ivy` branch which has the bleeding-edge version of JavaFX available:
    ```sh
    git clone -b ivy https://github.com/qzind/tray
    ```
2. Open IntelliJ, ensure Java 25 is ready and install the [`ant`](https://plugins.jetbrains.com/plugin/23025-ant) plugin.
3.  Once installed, run the `ivy-deps` task to resolve any missing dependencies -- including JavaFX -- from Maven Central.
4.  Locate `WebAppTest.java` in the `test` directory and run `main()` (this class will eventually be converted to TestNG but for now is a standalone test that is not automatically run with project unit tests)
5.  Change `rasterKnownHeightTests` from `1000` to `0` to skip unrelated tests.  We will be focusing on `rasterFittedHeightTests`.
6.  Run `main()` and look in `out/` directory for results.

> [!NOTE]
> The height of the documents should be resized to "fit" the content.  Do not rely on any pass/fail logic in this test and instead use manual inspection to see if sizing has occurred correctly.

Related:
* #547
* #1426

tresf:
Fixes inconsistent HTML height calculation when `rasterize: true` is used.
* Call `updatePeer` each time the `WebView` size changes
* Fix JavaScript height calculation
* Simplify scrollbar hiding code (remove CSS injection)

TODO:
 * [x] Fix vector fitted prints regression
 * [x] ~~Migrate `doUpdatePeer()` to call `NodeHelper.updatePeer(webView);` instead~~<br>**Edit:** Shelving this; causes issues with CI.

This is a PR against JavaFX 27-ea in the `ivy` branch (which will eventually become 2.3.0) but the patches should work independently of `ivy` (e.g. cherry-picked against 2.2.7 if needed.)

* Fixes #1447
* Possibly addresses #1426

### DIFF
--- a/src/qz/printer/action/html/WebApp.java
+++ b/src/qz/printer/action/html/WebApp.java
@@ -1,6 +1,7 @@
 package qz.printer.action.html;
 
 import com.github.zafarkhaja.semver.Version;
+import com.sun.javafx.scene.NodeHelper;
 import com.sun.javafx.tk.TKPulseListener;
 import com.sun.javafx.tk.Toolkit;
 import javafx.animation.AnimationTimer;
@@ -28,7 +29,6 @@ import qz.ws.PrintSocketServer;
 
 import java.awt.image.BufferedImage;
 import java.io.IOException;
-import java.lang.reflect.Method;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
@@ -57,7 +57,6 @@ public class WebApp extends Application {
     private static double pageWidth;
     private static double pageHeight;
     private static double pageZoom;
-    private static boolean raster;
     private static boolean headless;
 
     private static CountDownLatch startupLatch;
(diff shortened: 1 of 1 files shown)


tresf:
Well, we were in good shape, but it appears a blank page regression occurs now, but only when printing vector and the height is not known.

PR #1453, opened 2026-05-20
Add unit tests for META-INF/services

tresf:
Tests `META-INF/services` installations.

* [x] ICU4J (Arabic)
* [x] Vietnamese
* [x] Jbig2
* [x] Jpeg2000
* [x] TwelveMonkeys
* [x] BouncyCastle (in PDFBox, Security)<br>... more?

This "service" issue is best explained on the TwelveMonkey's README, so I'll quote it:

> The recommended way to use the plugins, is just to include the JARs as-is in your project, through a Maven dependency or similar. Re-packaging is not necessary to use the library, and not recommended.
> 
> However, if you like to create a "fat" JAR, or otherwise like to re-package the JARs for some reason, it's important to remember that automatic discovery of the plugins by ImageIO depends on the [Service Provider Interface (SPI)](https://docs.oracle.com/javase/tutorial/sound/SPI-intro.html) mechanism. In short, each JAR contains a special folder, named `META-INF/services` containing one or more files, typically `javax.imageio.spi.ImageReaderSpi` and `javax.imageio.spi.ImageWriterSpi`. These files exist with the same name in every JAR, so if you simply unpack everything to a single folder or create a JAR, files will be overwritten and behavior be unspecified (most likely you will end up with a single plugin being installed).
> 
> The solution is to make sure all files with the same name, are merged to a single file, containing all the SPI information of each type. If using the Maven Shade plugin, you should use the [ServicesResourceTransformer](https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html#ServicesResourceTransformer) to properly merge these files. You may also want to use the [ManifestResourceTransforme](https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html#ManifestResourceTransformer) to get the correct vendor name, version info etc. Other "fat" JAR bundlers will probably have similar mechanisms to merge entries with the same name.

QZ Tray uses `ant` (not `maven` plugins) so these files are merged using a manual effort here:

https://github.com/qzind/tray/blob/aaa76ba8bd949336e4e0a1f0db6c623272415c6d/build.xml#L301-L307

Sadly, if this steps fails, it's not immediately obvious, so @bratpeki is adding unit tests to detect this directly to the codebase.


tresf:
CI failure is due to a bug with TCVN encoding.  Nice find @bratpeki.  Please link the upstream bug report once filed. https://github.com/yohanes/vncharsets

Please feel free to guard this exception for now with a `// TODO` comment linking the upstream bug.

test/qz/integration/services/Jbig2Tests.java
```diff
@@ -0,0 +1,70 @@
...
+		ImageIO.scanForPlugins();
+
+		Path projectRoot = Path.of(System.getProperty("user.dir"));
+
+		Path assetsFolder = projectRoot
+			.resolve("test")
+			.resolve("qz")
+			.resolve("integration")
+			.resolve("services")
+			.resolve("assets");
```
tresf: `getResourceAsStream` with relative path please.

test/qz/integration/services/Jbig2Tests.java
```diff
@@ -0,0 +1,70 @@
...
+			);
+
+			try {
+
+				reader.setInput(iis, false, false);
+
+				BufferedImage decoded = reader.read(0);
+				Assert.assertNotNull(decoded, "Decoded image is null!");
+
+				File truthFile = assetsFolder.resolve("jbig2-cameraman.png").toFile();
```
tresf: Rather than having baselines here that test the pixel-perfect accuracy of the image handlers, I'm curious if there's a simpler way, such as to ensure that the decoding worked (e.g. content vs. blank pixels, exception thrown, etc).

I'm fine with baseline tests, but we have other areas of the code that I feel they would fit those tests better because those part of the code actually provide mechanisms for **updating the baselines** for instances when something out of our control changes (such as the PDF, HTML rendering -- or in this case, the image rendering).  Does that make sense?  These are integration tests and should only ensure that the services are properly registered with the JVM, not that they're pixel-accurate.

test/qz/integration/services/Jbig2Tests.java
```diff
@@ -15,53 +14,38 @@ public void testJbig2() throws Exception {
 
 		ImageIO.scanForPlugins();
 
-		Path projectRoot = Path.of(System.getProperty("user.dir"));
+		String assetPath = "qz/integration/services/assets/jbig2-cameraman.jb2";
+		String truthPath = "qz/integration/services/assets/jbig2-cameraman.png";
```
tresf: Should be able to use relative paths here...

tresf: Example: <https://github.com/qzind/tray/blob/9da43e0e6fb27c5994ec6e34c8b78e69b5b87a99/test/qz/ws/substitutions/SubstitutionsTests.java#L11>

test/qz/integration/services/ImageReaderSpiTests.java
```diff
@@ -23,10 +24,19 @@ public class ImageReaderSpiTests {
 	@DataProvider(name = "nameAndImage")
 	public Object[][] nameAndImage() {
 		return new Object[][] {
+				// JBIG2 - Provided by Apache PDFBox
 				{
-					// JBIG2 - Provided by Apache PDFBox
 					"JBIG2",
 					"assets/jbig2-cameraman.jb2"
+				},
+				// JPEG 2000 - Provided by ???
```
tresf: ~~TwelveMonkeys provides it.~~

tresf: I'm wrong jai-imageio provides it, which isn't even a real organization, but rather a few volunteers keeping these formats alive.  I tried to learn the history to make this comment useful but I can't find one, so for now, `Provided by jai-imageio` is just fine

test/qz/integration/services/ImageReaderSpiTests.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@DataProvider(name = "nameAndImage")
+	public Object[][] nameAndImage() {
+		return new Object[][] {
+				// JBIG2 - Provided by Apache PDFBox
+				{
+					"JBIG2",
+					"assets/jbig2-cameraman.jb2"
+				},
+				// JPEG 2000 - Provided by ???
```
bratpeki:
```suggestion
				// JPEG 2000 - Provided by jai-imageio
```

bratpeki:
Some tests failing, for some reason. Will look into it!

tresf:
> Some tests failing, for some reason. Will look into it!

It was just a typo.  I fixed it and pushed some other simplifications.

bratpeki:
Nice! Thanks for taking care of the `DataProvider` duplication, I wanted to take care of it as well

tresf:
Finally added TwelveMonkeys coverage.  Thoughts welcome here, had to refactor a bit. 😅

tresf:
Gluon mirrors appear to be down for JavaFX.  Will need to refire the CI when it's back online.

PR #1461, opened 2026-06-12
Introduce isConnected method for qz.websocket in order to check for an established connection.

### LINKED ISSUE #1458: qz-tray.js: isActive() is also true while connecting
Hi @tresf,

I'm wondering what the motivation behind the concept isActive() === true while the websocket is connecting is.
In my opinion a connection is only active when it is truly connected, connecting state might not result in a connection.

In async operations qz reports back that it is active, while in fact it is still trying to connect.

```
            isActive: function() {
                return !_qz.websocket.shutdown && _qz.websocket.connection != null
                        || qz.websocket.connection.readyState === _qz.tools.ws.CONNECTING);
            },
```

antoineveldhoven:
This additional method won't break any exisiting implementations.

Aims to provide a solution for #1458

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -901,6 +901,11 @@ var qz = (function() {
                 throw new Error("A connection to " + _qz.TITLE + " has not been established yet");
             },
 
+            isConnected: function() {
+                return !_qz.websocket.shutdown && _qz.websocket.connection != null
+                    && _qz.websocket.connection.established;
+            },
+
             uint8ArrayToHex: function(uint8) {
                 return Array.from(uint8)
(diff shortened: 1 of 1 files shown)


tresf:
@antoineveldhoven thanks, will discuss with @akberenz and decide what's best here.

tresf:
@antoineveldhoven thanks again for the attention to this issue as well as the offer to fix it.

With the help fo @akberenz we took a slightly simpler direction.  Closing this out as superceded by https://github.com/qzind/tray/pull/1462.  Please request a reopen at #1458 if you spot a problem.

PR #1462, opened 2026-06-14
isActive no longer returns true during a CONNECTING state

### LINKED ISSUE #1458: qz-tray.js: isActive() is also true while connecting

akberenz:
Closes #1458

### DIFF
--- a/js/qz-tray.js
+++ b/js/qz-tray.js
@@ -889,8 +889,7 @@ var qz = (function() {
 
             isActive: function() {
                 return !_qz.websocket.shutdown && _qz.websocket.connection != null
-                    && (_qz.websocket.connection.readyState === _qz.tools.ws.OPEN
-                        || _qz.websocket.connection.readyState === _qz.tools.ws.CONNECTING);
+                    && _qz.websocket.connection.readyState === _qz.tools.ws.OPEN;
             },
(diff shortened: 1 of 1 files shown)


tresf:
@akberenz Thanks.  I found this function is also used with `assertActive()`.  Do you anticipate any regressions as a result of this change?

tresf:
Quoting a PM from @akberenz
> im not sure if it would have any
> it seems to be used in two spots:
> `versionCompare` which needs the connection to be `CONNECTED` to work, so it'll improve that; and,
> `getConnectionInfo` to pull the details of the websocket url which im not sure if it would even work in a `CONNECTING` state either https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/url

So this looks good to merge.

PR #1468, opened 2026-06-23
Added Filtering for WebSocketTimeoutException

### LINKED ISSUE #691: java.util.concurrent.TimeoutException: Idle timeout expired: 300015/300000 ms
Quoting:

> Is there anything we can do to prevent this from happening?

image

> Do you think it's possibly a problem if they had two browser tabs open on the same page and they start competing using FileIO?

```
2020-07-16 08:32:31,128 [ERROR] Failed to read '/Users/jack/Downloads/io/MYFILE.IN' (does not exist)
2020-07-16 08:33:02,130 [DEBUG] Message: {"call":"file.read","promise":{},"params":{"path":"/Users/jack/Downloads/io/MYFILE.IN","sandbox":false,"shared":false},"timestamp":1594913582009,"uid":"zbv44m","position":{"x":768,"y":491.5}}
2020-07-16 08:33:02,134 [ERROR] Failed to read '/Users/jack/Downloads/io/MYFILE.IN' (does not exist)
2020-07-16 08:33:33,126 [DEBUG] Message: {"call":"file.read","promise":{},"params":{"path":"/Users/jack/Downloads/io/MYFILE.IN","sandbox":false,"shared":false},"timestamp":1594913612996,"uid":"utizud","position":{"x":768,"y":491.5}}
2020-07-16 08:33:33,130 [ERROR] Failed to read '/Users/jack/Downloads/io/MYFILE.IN' (does not exist)
2020-07-16 08:34:04,087 [DEBUG] Message: {"call":"file.read","promise":{},"params":{"path":"/Users/jack/Downloads/io/MYFILE.IN","sandbox":false,"shared":false},"timestamp":1594913643954,"uid":"bid1s0","position":{"x":768,"y":491.5}}
2020-07-16 08:34:04,092 [ERROR] Failed to read '/Users/jack/Downloads/io/MYFILE.IN' (does not exist)
2020-07-16 08:34:06,084 [DEBUG] Message: {"call":"file.read","promise":{},"params":{"path":"/Users/jack/Downloads/mp2/AUTOMAT.IN","sandbox":false,"shared":false},"timestamp":1594913645968,"uid":"9pg4cy","position":{"x":768,"y":491.5}}
2020-07-16 08:39:06,095 [ERROR] Connection error
org.eclipse.jetty.websocket.api.CloseException: java.util.concurrent.TimeoutException: Idle timeout expired: 300000/300000 ms
at org.eclipse.jetty.websocket.common.io.AbstractWebSocketConnection.onReadTimeout(AbstractWebSocketConnection.java:560)
at org.eclipse.jetty.io.AbstractConnection.onFillInterestedFailed(AbstractConnection.java:172)
at org.eclipse.jetty.websocket.common.io.AbstractWebSocketConnection.onFillInterestedFailed(AbstractWebSocketConnection.java:533)
at org.eclipse.jetty.io.AbstractConnection$ReadCallback.failed(AbstractConnection.java:317)
at org.eclipse.jetty.io.FillInterest.onFail(FillInterest.java:138)
at org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.onFillableFail(SslConnection.java:466)
at org.eclipse.jetty.io.ssl.SslConnection.onFillInterestedFailed(SslConnection.java:329)
at org.eclipse.jetty.io.ssl.SslConnection$2.failed(SslConnection.java:164)
at org.eclipse.jetty.io.FillInterest.onFail(FillInterest.java:138)
at org.eclipse.jetty.io.AbstractEndPoint.onIdleExpired(AbstractEndPoint.java:407)
at org.eclipse.jetty.io.IdleTimeout.checkIdleTimeout(IdleTimeout.java:171)
at org.eclipse.jetty.io.IdleTimeout.idleCheck(IdleTimeout.java:113)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.util.concurrent.TimeoutException: Idle timeout expired: 300000/300000 ms
... 8 more
```

Quoting @bberenz:

> Well that just sounds like the tab with the connection is in the background

But @bberenz we have a ping/heartbeat preventing the websocket from every closing as idle, no?  Why would this ping/heartbeat stop?  is there something in FileIO that locks the websocket from sending a ping back?

Vzor-:
Fixes #691 

I think jetty modified how the timeout exception was thrown, which broke our targeted catch. I wasn't able to find a mention of this in any migration guide though. They now use the much simpler WebSocketTimeoutException. I considered breaking the filter out into a method then adding a go and no-go test, but decided against it.

### DIFF
--- a/src/qz/ws/PrintSocketClient.java
+++ b/src/qz/ws/PrintSocketClient.java
@@ -10,8 +10,8 @@ import org.codehaus.jettison.json.JSONObject;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.websocket.api.Session;
 import org.eclipse.jetty.websocket.api.annotations.*;
-import org.eclipse.jetty.websocket.api.exceptions.CloseException;
 import org.eclipse.jetty.websocket.api.exceptions.WebSocketException;
+import org.eclipse.jetty.websocket.api.exceptions.WebSocketTimeoutException;
 import org.eclipse.jetty.websocket.server.JettyServerUpgradeRequest;
 import org.eclipse.jetty.websocket.server.JettyServerUpgradeResponse;
 import org.usb4java.LoaderException;
@@ -39,7 +39,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.concurrent.Semaphore;
-import java.util.concurrent.TimeoutException;
 
 
 @WebSocket
@@ -90,7 +89,7 @@ public class PrintSocketClient {
     public void onError(Session session, Throwable error) {
         if (error instanceof EOFException || error instanceof ClosedChannelException) { return; }
 
-        if (error instanceof CloseException && error.getCause() instanceof TimeoutException) {
+        if (error instanceof WebSocketTimeoutException) {
             log.error("Timeout error (Lost connection with client)", error);
             return;
         }


tresf:
Possible refactor that changed the exception type:
* https://github.com/jetty/jetty.project/pull/12359

tresf:
This has been a very difficult patch to confirm because of the way Jetty wraps exceptions internally with generic handlers.  We combed through the source and the best we can determine is that somewhere between Jetty 9.x and 10.x (specifically some of the early 10.x betas) Jetty greatly improved the way they report these `TimeoutExceptions`. 

We challenged a chatbot to help us find this code and it came up with the following, which was part of the Jetty 10.0-beta changes which seems to convert the cause back to the underlying exception type, removing the need to call `getCause()` as we have for so many years.

```java
    public static Throwable convertCause(Throwable cause)
    {
       // [...]

        if (cause instanceof CloseException)
            return new org.eclipse.jetty.websocket.api.exceptions.CloseException(((CloseException)cause).getStatusCode(), cause.getMessage(), cause);

        if (cause instanceof WebSocketTimeoutException)
            return new org.eclipse.jetty.websocket.api.exceptions.WebSocketTimeoutException(cause.getMessage(), cause);

        }

        // [...]

        return cause;
    }
}
```

The paranoid version of this patch will keep the old `getCause()` and add the new condition, but I don't think that paranoid programming is a good strategy here as it trades risk for technical debt.

If we don't handle this right, we'll be getting reports with QZ Tray 2.3.0 of increased notifications, so it's with great hesitation that I merge this... 

Note that @Vzor- did self-testing on this for Linux and Windows and in both cases, this patch suppressed the popups that are observed in https://github.com/qzind/tray/issues/691.

Issue #1393, opened 2025-12-10
Show more meaningful error when CUPS-BSD is not installed


tresf:
On some systems (such as Raspberry Pi OS) the package CUPS-BSD is needed for Java to print.  We should consider showing a more meaningful error when CUPS-BSD is not installed.

See also:
* https://github.com/qzind/tray/issues/1341

Currently at install time, we ask the user to install `certutil`, we may decide to do the same for `cups-bsd`.

https://github.com/qzind/tray/blob/5e01218f7cb4674f435eb9cf674b155839addcc1/src/qz/installer/certificate/LinuxCertificateInstaller.java#L70-L75

Issue #1341, opened 2025-04-23
Getting java.awt.print.PrinterIOException on RP5 and CUPS


captainjapeng:
I have QZ on a Raspberry Pi 5 on DietPI with CUPS and Gutenprint. I'm able to print to the printer using CUPS's print test page and lp command but whenever printing with QZ I'm getting the exact error from #1322

![Image](https://github.com/user-attachments/assets/34825516-0750-4cc4-b552-6721ddc221e3)

tresf:
Duplicate of #642. In short, Java printing requires `lpr`, which requires `cups-bsd` package.  Please DO NOT install the `lpr` standalone package.   See also: https://forums.raspberrypi.com/viewtopic.php?t=274302#p1662258

Issue #642, opened 2020-05-20
Error: java.io.IOException: error=1 running: '/usr/bin/lpr'


tresf:
Documenting this for others that come across it...

QZ Tray and Raspbian will throw this error:

```
Error: java.io.IOException: error=1 running: '/usr/bin/lpr' '-PPDF' '-J QZ Tray Pixel Print' '-h' '-o media=Letter' '/tmp/javaprint5623279197230565878' usage: lpr [-cdfghlmnpqrstv] [-#num] [-1234 font] [-C class] [-i [numcols]] [-J job] [-Pprinter] [-T title] [-U user] [-wnum] [name ...]
```

The solution is to install the package `cups-bsd`

PR #1480, opened 2026-07-24
Add cups-bsd error hinting for pixel printing in Linux

IdelsTak:
@tresf this PR addresses #1393. and in turn addresses #1341 too and references this old issue too: #642 
 
So here's what's changed, this PR:

- Adds an actionable Linux pixel-printing error when Java printing fails because the CUPS-compatible `lpr` command is missing
- Preserves the original `PrinterException` as the cause for debugging
- Keeps the change scoped to `PrintPixel#printCopies(...)`; raw printing and the `lp` fallback path are not changed
- Names `cups-bsd` as the package to install and explicitly warns against standalone `lpr`

I also added unit testing for all those new capabilities via the new class `PrintingUtilitiesTests`

Now, here's how to reproduce the manual smoke testing I used to test these features in a headless env:

### Smoke testing in a headless env

First, I created the necessary certs and keys using QZ tray GUI in my archlinux

Thus I ended up with 3 files
- qz-smoke-cert.txt
- qz-smoke-key.pem
- override.crt

I then copied these to my ubuntu server installation (in the vm) in a folder in its `$HOME` named `certs`

Then, in the `tray` folder  in the same server (I cloned `tray` in the server too) I ran:

1. to accept the cert/key pair 

```bash
java -DtrustedRootCert="$HOME/certs/override.crt" \
  -cp 'out/build/qz-tray:deps/ivy/*:deps/manual/*' \
  qz.App --allow "$HOME/certs/qz-smoke-cert.txt"
```
2. To start the headless qz tray

```bash
java -DtrustedRootCert="$HOME/certs/override.crt" \
  -Dwebsocket.secure.ports=8181 \
  -Dwebsocket.insecure.ports=8182 \
  -Dsecurity.wss.alloworigin=https://demo.qz.io \
  -cp 'out/build/qz-tray:deps/ivy/*:deps/manual/*' \
  qz.App --headless > /tmp/qz-tray-smoke.log 2>&1
  ```
Then in my archlinux, I forwarded the qz ports to the ubuntu server:

```bash
 ssh -N \
  -L 8182:127.0.0.1:8182 \
  -L 8181:127.0.0.1:8181 \
  <username>@<ubuntu-server ip>
  ```
Then I opened https://demo.qz.io/ in a browser in archlinux and before connecting, I opened dev tools and I overrode the page signing hooks so protected qz calls were signed with the matching private key using this
  
```js
const cert = `<contents of qz-smoke-cert.txt>`;
const privateKey = `<contents of qz-smoke-key.pem>`;

await new Promise((resolve, reject) => {
  const s = document.createElement("script");
  s.src = "https://cdnjs.cloudflare.com/ajax/libs/jsrsasign/11.1.0/jsrsasign-all-min.js";
  s.onload = resolve;
  s.onerror = reject;
  document.head.appendChild(s);
});

qz.security.setCertificatePromise(resolve => resolve(cert));
qz.security.setSignatureAlgorithm("SHA512");
qz.security.setSignaturePromise(toSign => (resolve, reject) => {
  try {
    const pk = KEYUTIL.getKey(privateKey);
    const sig = new KJUR.crypto.Signature({ alg: "SHA512withRSA" });
    sig.init(pk);
    sig.updateString(toSign);
    resolve(stob64(hextorstr(sig.sign())));
  } catch (e) {
    reject(e);
  }
});
```

With that done and connecting to qz from the `demo.qz.io` page and attempting to print an image for instance, I got the error

```text
Java printing on Linux requires the CUPS-compatible "lpr" command. Install the "cups-bsd" package, not the standalone "lpr" package, then retry.
```
In the ubuntu server (where qz was running) the qz log also contained the error message (that this PR introduces), while retaining the underlying print failure for debugging

However after installing `cups-bsd` on the ubuntu server:

```bash
sudo apt install cups-bsd
which lpr
# /usr/bin/lpr
```

and I restarted the headless qz, reconnected `demo.qz.io`, and retried Pixel Printing -> `Print Image`
> the missing-`lpr` / `cups-bsd` message no longer appeared.


IdelsTak:
So @tresf here's a showcase of the helpful errors shown when `cups-bsd` isn't installed:

https://github.com/user-attachments/assets/f6407a87-5f82-4c36-9ffa-a4c5be883e38


https://github.com/user-attachments/assets/710ae4f2-4795-48b4-81b1-d0e09fbd24ec

IdelsTak:
And here's the result when `cups-bsd` is found/installed


https://github.com/user-attachments/assets/b061efcf-2a8c-4adb-837d-6f36f0e88618


https://github.com/user-attachments/assets/ac12e0d2-c02f-4bd0-a8a3-55eff07b3c7c

test/qz/utils/PrintingUtilitiesTests.java
```diff
@@ -0,0 +1,107 @@
...
+        Assert.assertTrue(PrintingUtilities.hasLprSignature(ex));
+        Assert.assertTrue(PrintingUtilities.isLikelyCupsBsdMissing(ex, false));
+    }
+
+    @Test
+    public void nestedLprSignatureTests() {
+        PrinterException ex = new PrinterException("Failed to print");
+        // Missing lpr shape from Raspberry Pi thread
+        // linked by issue #1341
+        ex.initCause(new IOException("Cannot run program \"/usr/bin/lpr\": error=2, No such file or directory"));
```
tresf: Manipulating the system to actually generate this error is a better test, otherwise we're essentially just testing our own strings.

IdelsTak: Mmmh...I was covering this [issue](https://github.com/qzind/tray/issues/642) so that it never escapes detection ever again

tresf: Constructing this exception by hand is not "coverage" because these tests are very likely to ever fail.   Since the JVM has this `/usr/bin/lpr` as a hard-coded path, I believe we should consider finding this value via reflection.  We can test the binary at this path.   To manipulate the exception from the JVM, we may be able to override this array value before printing to a known-bad-location.

I think constructing a matching exception is not proper coverage because it will not catch changes to this hard-coded path logic in future JVMs.

[`src/java.desktop/unix/classes/sun/print/UnixPrintJob.java#L917`](https://github.com/openjdk/jdk11u-dev/blob/24a7cee41dbe81cb320e900712e780c3df08bbd0/src/java.desktop/unix/classes/sun/print/UnixPrintJob.java#L917)

```java
public class UnixPrintJob implements CancelablePrintJob {
    private String[] printExecCmd(String printer, String options,
                                 boolean noJobSheet,
                                 String jobTitle, int copies, String spoolFile) {
        if (PrintServiceLookupProvider.osname.equals("SunOS")) {
// ...
        } else {
            execCmd = new String[ncomps];
            execCmd[n++] = "/usr/bin/lpr";
// ...
```

IdelsTak: Mmmh...I agree these tests don't provide end-to-end coverage of the JVM print path.

My intent was narrower though. I aimed for regression tests for the exact exception shapes that the users reported in #642 and #1341 for example. So those historical scenarios will keep getting covered

For the reflection idea, I see the concern

But how far would we want to explore that idea in the tests? Because don't you think it may be somehow brittle?

tresf: > don't you think it may be somehow brittle?

Yes, but if this underlying JVM approach changes we should probably be double-checking our workaround and our unit tests anyway.

I would also be happy with an approach that overrides this `/usr/bin/lpr` value by injecting a different value in this element of the array.  This could theoretically allow us to create the exception (expected throw), suppress the exception (expected pass) and also provide the granularity we need in the future if -- for example -- the JVM is patched to allow a different value.

IdelsTak: Thanks @tresf, this makes sense

I updated the PR to separate the two kinds of coverage:

- The existing manually constructed exception tests are now framed as classifier regression coverage for the historical #642/#1341 error shapes.
- I added a Linux-only reflection test against `sun.print.UnixPrintJob#printExecCmd(...)` so the test now proves the current JVM still builds the Unix print command with `/usr/bin/lpr`.
- The testng task now includes `--add-opens java.desktop/sun.print=ALL-UNNAMED` so that the reflection assertion runs in CI properly without being blocked by Java module access

I thought about and did not add command-array mutation because that would have needed bytecode instrumentation

And anyway, after checking the [JDK source](https://github.com/openjdk/jdk/blob/e0cfc8c4ea52eaf5636483e5f779054f756c83ae/src/java.desktop/unix/classes/sun/print/UnixPrintJob.java#L826) it's clear that `/usr/bin/lpr` is assigned into a method-local array inside `printExecCmd(...)`

So overriding it during the real print flow would require a heavier/brittle instrumentation setup

The reflection test gives us the JVM-coupled proof we need now, and if JDK later changes that method or command path, this test should force us to re-check the workaround.

src/qz/utils/PrintingUtilities.java
```diff
@@ -32,6 +35,19 @@
...
+    private static final String[] LPR_DIRECT_SIGNATURES = {
+            "/usr/bin/lpr",
+            "lpr: command not found",
+            "usage: lpr",
+            "-j qz tray pixel print"
+    };
+    private static final String[] PRINTER_IO_SIGNATURES = {
+            // Wrapped Java print stack from #1341 https://github.com/qzind/tray/issues/1341
+            "java.awt.print.printerioexception",
+            "javax.print.printexception",
```
tresf: Hmm... since we match ANY signature, this feels like it could cause a false-positive.  I think getCause() on exact exception type is more succinct.  Furthemore, I do not see an advantage in using the exception by name instead of fetching it from the class instance.   I will type my other thoughts as a general comment as I believe this may be more complex than it needs to be.

tresf:
Thank you for the work on this.

I think this PR can be simplified to handle the exact reproducible use-cases that we're aware of, and work forward from there.  It also does not address the proactive install-time recommendation hinted in #1393 which would make most of this try/catch logic much less common.

The smaller the PR, the easier it is to review.   For example, in https://github.com/qzind/tray/pull/1415 we do a catch-and-release style exception handling, this may benefit from a similar coding style.

It may also be more pragmatic to draft a bug report to OpenJDK (we can file this with our JDK support vendor) so that this can be fixed upstream in a future OpenJDK version.

I think the first step should be a small, reproducible code snippet that can be run on a Raspberry Pi, rather than a full blown smoke testing.  Alternately, it may also be beneficial to offer instructions for modifying a base Ubuntu 24.04 install to "break" it in a similar fashion as Raspberry Pi so that the additional work of starting the software in headless is removed and the code can be tested and studied without the need for complex environment.

tresf:
Here's a (untested) exception wrapper that may simplify this imlementation:

```java
package qz.ws;

import qz.utils.SystemUtilities;

import javax.print.PrintException;
import java.awt.print.PrinterException;
import java.awt.print.PrinterIOException;
import java.io.IOException;

public class MissingLprException extends Exception {
    private static final String MESSAGE = "The 'lpr' command appears to be missing.  Please install the 'cups-bsd' package and then try again.";

    MissingLprException(Exception e, String message) {
        super(message);
        this.initCause(e);
    }

    public static Exception wrap(Exception e) {
        return new MissingLprException(e, MESSAGE);
    }

    public static boolean isLprSignature(Exception e) {
        if(!SystemUtilities.isLinux()) {
            return false;
        }
        return switch(e) {
            case PrinterException p -> isClassSignature(p);
            case IOException i -> isMessageSignature(i);
            default -> false;
        };
    }

    private static boolean isClassSignature(PrinterException e) {
        return hasCauseChain(e, PrintException.class, PrinterIOException.class);
    }

    @SafeVarargs
    private static boolean hasCauseChain(Throwable throwable, Class<? extends Throwable>... expectedCauses) {
        Throwable current = throwable;

        for (Class<? extends Throwable> expectedClass : expectedCauses) {
            if (current == null || current.getCause() == null) {
                return false;
            }
            current = current.getCause();
            if (!expectedClass.isInstance(current)) {
                return false;
            }
        }

        return true;
    }

    private static boolean isMessageSignature(IOException e) {
        String message = e.getMessage();
        return message.contains("error") &&
                message.contains("running") &&
                message.contains("lpr");
    }
}
```

Sample usage:

```java
                catch(Exception e) {
                    Exception exception = e;
                    if(MissingLprException.isLprSignature(exception)) {
                        exception = MissingLprException.wrap(exception);
                    }
                    log.error("Problem processing message", exception);
                    sendError(session, tUID, exception);
                }
```

src/qz/utils/PrintingUtilities.java
IdelsTak: Okay, so @tresf, I removed the broad _match any signature_ helper path

I've added a narrower implementation:

 - It checks the expected exception/cause chain for the reproduced Java print failure
 - and the shared `/usr/bin/lpr` availability check now lives in `PrintingUtilities` for both installer and runtime use.

IdelsTak:
About this:

I reworked this around the smaller path you suggested

The runtime handling is now narrower and follows the catch-and-release style at the websocket boundary

I also added the proactive install-time check in `LinuxInstaller`: Debian-family systems missing `/usr/bin/lpr` now get a recommendation-only `cups-bsd` prompt during interactive installs, while silent/headless installs stay log-only

I also added the smaller repro artifact in `docs/cups-bsd-java-print-repro.md` with `test/manual/JavaPrintSmoke.java`, so the behavior can be reproduced without running the full QZ Tray/browser flow

> Thank you for the work on this.
> 
> I think this PR can be simplified to handle the exact reproducible use-cases that we're aware of, and work forward from there. It also does not address the proactive install-time recommendation hinted in #1393 which would make most of this try/catch logic much less common.
> 
> The smaller the PR, the easier it is to review. For example, in #1415 we do a catch-and-release style exception handling, this may benefit from a similar coding style.
> 
> It may also be more pragmatic to draft a bug report to OpenJDK (we can file this with our JDK support vendor) so that this can be fixed upstream in a future OpenJDK version.
> 
> I think the first step should be a small, reproducible code snippet that can be run on a Raspberry Pi, rather than a full blown smoke testing. Alternately, it may also be beneficial to offer instructions for modifying a base Ubuntu 24.04 install to "break" it in a similar fashion as Raspberry Pi so that the additional work of starting the software in headless is removed and the code can be tested and studied without the need for complex environment.

IdelsTak:
About this:

I adopted the suggested direction

`MissingLprException` now lives in `qz.ws`, wraps at the socket error boundary, preserves the original exception as the cause, and uses the narrower cause-chain check instead of broad signature matching

> Here's a (untested) exception wrapper that may simplify this imlementation:
> 
> ```java
> package qz.ws;
> 
> import qz.utils.SystemUtilities;
> 
> import javax.print.PrintException;
> import java.awt.print.PrinterException;
> import java.awt.print.PrinterIOException;
> import java.io.IOException;
> 
> public class MissingLprException extends Exception {
>     private static final String MESSAGE = "The 'lpr' command appears to be missing.  Please install the 'cups-bsd' package and then try again.";
> 
>     MissingLprException(Exception e, String message) {
>         super(message);
>         this.initCause(e);
>     }
> 
>     public static Exception wrap(Exception e) {
>         return new MissingLprException(e, MESSAGE);
>     }
> 
>     public static boolean isLprSignature(Exception e) {
>         if(!SystemUtilities.isLinux()) {
>             return false;
>         }
>         return switch(e) {
>             case PrinterException p -> isClassSignature(p);
>             case IOException i -> isMessageSignature(i);
>             default -> false;
>         };
>     }
> 
>     private static boolean isClassSignature(PrinterException e) {
>         return hasCauseChain(e, PrintException.class, PrinterIOException.class);
>     }
> 
>     @SafeVarargs
>     private static boolean hasCauseChain(Throwable throwable, Class<? extends Throwable>... expectedCauses) {
>         Throwable current = throwable;
> 
>         for (Class<? extends Throwable> expectedClass : expectedCauses) {
>             if (current == null || current.getCause() == null) {
>                 return false;
>             }
>             current = current.getCause();
>             if (!expectedClass.isInstance(current)) {
>                 return false;
>             }
>         }
> 
>         return true;
>     }
> 
>     private static boolean isMessageSignature(IOException e) {
>         String message = e.getMessage();
>         return message.contains("error") &&
>                 message.contains("running") &&
>                 message.contains("lpr");
>     }
> }
> ```
> 
> Sample usage:
> 
> ```java
>                 catch(Exception e) {
>                     Exception exception = e;
>                     if(MissingLprException.isLprSignature(exception)) {
>                         exception = MissingLprException.wrap(exception);
>                     }
>                     log.error("Problem processing message", exception);
>                     sendError(session, tUID, exception);
>                 }
> ```

tresf:
@IdelsTak thanks for the work on this.  After discussing in PM and reviewing the latest OpenJDK 25 source code, it appears that this `/usr/bin/lpr` is a hard-coded dependency for printing with Java.  For this reason, I decided to break-out the `certutil` helper and handle this at install-time.  (very similar to your detection, but reducing code duplication)
* #1499

The downside to my simpler approach is we have no way to detect if OpenJDK changes this hard-coded path... but after reviewing the code complexity that this test requires (versus the benefit we get from it) I've decided that it is not worthwhile so as long as we can detect this dependency proactively.

In the future if/when OpenJDK removes this hard-coded `/usr/bin/lpr` we will need to revisit this strategy.

tresf:
Superseded by #1500

Issue #1366, opened 2025-08-20
Shared serial port access between tabs


tresf:
Currently, serial ports are exclusive to the tab that opened it.

Steps to reproduce:
1. Visit demo.qz.io
1. Click Serial tab
1. List serial ports
1. Claim a serial port
1. Open a new tab
1. Visit demo.qz.io
1. Click Serial tab
1. Claim the same serial port as Step 4.
   1. It will error saying it's busy
2. Try closing the port
   1. It will say it's not open

Since we currently offer a many-to-one relationship for other comms and listeners, I believe serial ports can be enhanced to support multiple tabs as well.

Assigning to @Vzor- because he's authored this many-to-one successfully in several other areas.

PR #1485, opened 2026-08-04
Harden serial port close lifecycle around pending reads

IdelsTak:
@tresf, this PR proposes to introduce these fixes:

- a narrow serial port adapter seam so `SerialIO` lifecycle behavior can be tested without physical hardware
- a QZ-owned serial lifecycle state around open, close, read, and write paths
- preventing reads/writes after close has been requested and makes close idempotent
- focused SerialIO tests for close failures, pending reads, timeout use, and reopen behavior

You should however note that I do not have access to the faulty USB-serial adapter described in #1431

So this PR validates the lifecycle behavior with simulated JSSC failures. Thus physical hardware validation is still needed from someone with the affected hardware

In other matters `serial.closePort` still follows the existing close/remove/success websocket convention

Thus native close failures are logged, but not returned as websocket errors because that would change API behavior

So, should failed JSSC close be surfaced to the browser?

### DIFF
--- a/src/qz/communication/JsscSerialPortAdapter.java
+++ b/src/qz/communication/JsscSerialPortAdapter.java
@@ -1,0 +1,54 @@
+package qz.communication;
+
+import jssc.SerialPort;
+import jssc.SerialPortEventListener;
+import jssc.SerialPortException;
+import jssc.SerialPortTimeoutException;
+
+class JsscSerialPortAdapter implements SerialPortAdapter {
+    private final SerialPort port;
+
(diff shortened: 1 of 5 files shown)


tresf:
Some proposed changes to simplify this PR:
* https://github.com/IdelsTak/tray/compare/issue-1431-serial-close-lifecycle...tresf:tray:serial-refactor?expand=1

Note that we will likely hold off on this PR (or my own fork of it) until after #1366 is addressed as to not duplicate efforts.

PR #1492, opened 2026-08-11
Add basic constraints to CA certs

tresf:
Restrict CA certs to the same domains as the SSL certs

### DIFF
--- a/ant/dependencies.properties
+++ b/ant/dependencies.properties
@@ -109,6 +109,10 @@ commons-pool2.org=org.apache.commons
 commons-pool2.name=commons-pool2
 commons-pool2.rev=2.13.1
 
+commons-validator.org=commons-validator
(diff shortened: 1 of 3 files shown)


tresf:
Testing is currently failing for non-default hosts (e.g. `certgen --host="foo.bar"`).

tresf:
> Testing is currently failing for non-default hosts (e.g. `certgen --host="foo.bar"`).

Just forgot my own CLI.  It's `--host "foo.bar"` (no `=`).  This looks good.  Tested with `https://qzind2.duckdns.org`.

tresf:
* ✅ PASS: DNS name
* ✅ PASS: IPv4 IP Address
* 🚫  FAIL: IPv6 IP Address (appears to be an existing bug with `isIp(...)`.

image

Issue #435, opened 2019-03-19
reverse-portrait support


klabarge:
It looks like `REVERSE_PORTRAIT` is available as a valid `orientation` for HTML printing

https://github.com/qzind/tray/blob/a478f9ef94656ee89d860b274990dd50dc719c41/src/qz/printer/action/PrintHTML.java#L368

however, I was unable to get this to work, despite it being valid in the logs

```
Page layout = Paper=Paper: Default size=612.0x792.0 POINT Orient=REVERSE_PORTRAIT leftMargin=0.0 rightMargin=0.0 topMargin=0.0 bottomMargin=0.0
```

If `REVERSE_PORTRAIT` isn't supported, can we fallback on `rotation: 180`? This is what we are advising clients to do that need it.

Lastly, `REVERSE_PORTRAIT` doesn't look to be available for PDF or Images, however, it does have an adverse effect on PDF printing. Notice the left portion of the sample PDF is cut off.

![image](https://user-images.githubusercontent.com/12505463/54646991-43c41b80-4a77-11e9-82c7-cb30d75bf7b9.png)

PR #1495, opened 2026-08-14
Add reverse-portrait fallback rotation

IdelsTak:
@tresf, this PR proposes to address the concerns raised by issue #435 

The changes:

 - preserve `REVERSE_PORTRAIT` intent for images and PDFs
 - apply the desired fallback behavior of: portrait page geometry plus a 180-degree content rotation
 
 I've also included a manual smoke testing class `ReversePortraitTest` that applies the said fixes resulting in an output like this
 
 
issue-435-pdf-reverse-portrait


src/qz/printer/action/PrintPDF.java
```diff
@@ -320,6 +321,19 @@ public void print(PrintOutput output, PrintOptions options) throws PrinterExcept
...
     }
 
+    static double getPageRotation(double rotation, PrintOptions.Orientation orientation) {
+        double pageRotation = rotation;
+        if (orientation == PrintOptions.Orientation.REVERSE_PORTRAIT) {
+            pageRotation += orientation.getDegreesRot();
+        }
+        return pageRotation;
+    }
+
```
tresf: Consider moving to `PrintPixel` or `PixelOptions`.  Also since this is additive to the existing rotation value, please name the function to better explain that.

src/qz/printer/action/PrintPDF.java
```diff
@@ -320,6 +321,19 @@ public void print(PrintOutput output, PrintOptions options) throws PrinterExcept
...
+    static boolean isLandscapeOrientation(PrintOptions.Orientation orientation) {
+        return orientation == PrintOptions.Orientation.LANDSCAPE
+                || orientation == PrintOptions.Orientation.REVERSE_LANDSCAPE;
+    }
+
```
tresf: Consider moving to `PrintPixel` or `PixelOptions`.

src/qz/printer/action/PrintImage.java
```diff
@@ -343,6 +342,17 @@ private BufferedImage rotate(BufferedImage image, double angle) {
...
+        double imageRotation = rotation;
+        if (orientation == PrintOptions.Orientation.REVERSE_PORTRAIT) {
+            imageRotation += orientation.getDegreesRot();
+        }
+        if (mac && orientation != null && orientation.getAsOrientRequested() == OrientationRequested.REVERSE_LANDSCAPE) {
+            imageRotation += 180;
+        }
+        return imageRotation;
+    }
+
```
tresf: Some notes:

* `mac` is calculated as a static field, can be moved to helper.
* Please consider a overarching helper such as `isOrientationSupported(Os os)` and `isOrientationSupported(PrintPixel px)` or simply `isOrientationSupported()` that uses an `instanceof` switch on `this`.  This way we can centrally manage these quirks in a parent class (such as `PrintPixel`) using static helpers rather than special quirk code in each implementing class.
* The `orientation != null` guard can apply to the entire function body.
* Please comment each quirk well so we're not confused down the road and we're re-reading these quirks or if the a future JDK patches these quirks.

tresf (changes requested):

src/qz/printer/action/PrintPDF.java
IdelsTak: Done. Moved `isLandscapeOrientation(...)` to `PrintPixel` and removed the duplicate helper from `PrintPDF`.

`PrintPDF` now uses the shared helper for the landscape-only imageable-area branch, so the geometry decision is centralized with the other pixel orientation logic

src/qz/printer/action/PrintPDF.java
IdelsTak: Done. Moved this into the shared `PrintPixel` helper as `getAdjustedRotation(...)`

The helper now makes the additive behavior explicit by starting with the configured rotation and applying orientation-specific adjustments in one place.

src/qz/printer/action/PrintImage.java
IdelsTak: Done. Centralized the orientation quirks in `PrintPixel`.

Changes made:
 - moved the macOS check into `isMacReverseLandscape(...)`
 - replaced `PrintImage.getImageRotation(...)` with shared `getAdjustedRotation(...)`
 - applied the null-orientation guard at the top of the shared rotation helper
 - added comments explaining each quirk, with `See:` links to the relevant QZ issues and JDK docs

I also kept `PDFWrapper` from needing broader helper visibility by passing it a precomputed `adjustForReverseLandscape` boolean

IdelsTak: @tresf, about this point that you suggested:

> Please consider a overarching helper such as `isOrientationSupported(Os os)` and `isOrientationSupported(PrintPixel px)` or simply `isOrientationSupported()` that uses an `instanceof` switch on `this`. This way we can centrally manage these quirks in a parent class (such as `PrintPixel`) using static helpers rather than special quirk code in each implementing class.

I used a centralized helper in `PrintPixel`:

```java
protected static double getAdjustedRotation(double rotation, PrintOptions.Orientation orientation) {
    if (orientation == null) {
        return rotation;
    }

    if (orientation == PrintOptions.Orientation.REVERSE_PORTRAIT) {
        rotation += orientation.getDegreesRot();
    }

    if (isMacReverseLandscape(orientation)) {
        rotation += 180;
    }

    return rotation;
}
```

That way any other required future quirks for other OSes can be plugged in into a full switch that has now started with a simple `if` check on mac

src/qz/printer/action/PrintPDF.java
```diff
@@ -281,22 +282,21 @@ public void print(PrintOutput output, PrintOptions options) throws PrinterExcept
...
+                } else if (isLandscapeOrientation(pxlOpts.getOrientation())) {
                     //flip imageable area dimensions when in landscape
                     Paper repap = page.getPaper();
                     repap.setImageableArea(repap.getImageableX(), repap.getImageableY(), repap.getImageableHeight(), repap.getImageableWidth());
                     page.setPaper(repap);
 
-                    //reverse fix for OSX
-                    if (SystemUtilities.isMac() && pxlOpts.getOrientation() == PrintOptions.Orientation.REVERSE_LANDSCAPE) {
+                    if (isMacReverseLandscape(pxlOpts.getOrientation())) {
                         pd.setRotation(pd.getRotation() + 180);
```
tresf: Hmm... I still think this can be consolidated... `+ 180` should live in `PrintPixel` by checking `this instanceof PrintPDF`.  If `pd.getRotation()` is needed for calculation, please add an override that takes the existing rotation value for calculation.

src/qz/printer/action/PrintPixel.java
```diff
@@ -188,6 +188,40 @@ protected static Map<RenderingHints.Key,Object> buildRenderingHints(Object dithe
         return rhMap;
     }
 
+    protected static boolean isLandscapeOrientation(PrintOptions.Orientation orientation) {
+        return orientation == PrintOptions.Orientation.LANDSCAPE
+                || orientation == PrintOptions.Orientation.REVERSE_LANDSCAPE;
+    }
+
+    protected static boolean isMacReverseLandscape(PrintOptions.Orientation orientation) {
```
tresf: I believe this helper can be in-lined once the calculation is consolidated.

src/qz/printer/action/PrintPixel.java
```diff
@@ -188,6 +188,40 @@ protected static Map<RenderingHints.Key,Object> buildRenderingHints(Object dithe
         return rhMap;
     }
 
+    protected static boolean isLandscapeOrientation(PrintOptions.Orientation orientation) {
```
tresf: Seems odd to make this protected.  Could even live in `PrintOptions` as a static helper.

src/qz/printer/action/PrintPixel.java
```diff
@@ -188,6 +188,40 @@ protected static Map<RenderingHints.Key,Object> buildRenderingHints(Object dithe
...
+
+    protected static boolean isMacReverseLandscape(PrintOptions.Orientation orientation) {
+        // macOS needs an extra reverse-landscape adjustment in these AWT print paths
+        // because its landscape origin differs from the Windows/PostScript convention
+        // See: https://github.com/qzind/tray/issues/197
+        //      https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/java/awt/print/PageFormat.html
+        return SystemUtilities.isMac() && orientation == PrintOptions.Orientation.REVERSE_LANDSCAPE;
+    }
+
+    protected static double getAdjustedRotation(double rotation, PrintOptions.Orientation orientation) {
```
tresf: If this is non-static, `this instanceof` can be used to use this for dynamic calculation per other comment.

src/qz/printer/action/PrintPDF.java
IdelsTak: Updated

The PDF reverse-landscape `+ 180` adjustment no longer lives directly in `PrintPDF`.

`PrintPDF` now calls the shared `PrintPixel` overload:

```java
getAdjustedRotation(0, pxlOpts.getOrientation(), pd.getRotation())
```

The overload accepts the existing PDF page rotation and the actual adjustment is handled inside `PrintPixel` through the centralized `switch(this)` path

src/qz/printer/action/PrintPixel.java
IdelsTak: Updated

After consolidating the rotation calculation, `isMacReverseLandscape(...)` is no longer a protected helper

It is now private to `PrintPixel`, so the macOS reverse-landscape quirk remains documented and centralized without becoming subclass API

src/qz/printer/action/PrintPixel.java
IdelsTak: Updated

Removed the protected `PrintPixel.isLandscapeOrientation(...)` helper

Landscape classification now lives on the enum as `PrintOptions.Orientation.isLandscape()`, and the PDF geometry branch plus tests now use that directly

src/qz/printer/action/PrintPixel.java
IdelsTak: Updated

`getAdjustedRotation(...)` is now instance-level and uses `switch(this)` for the print-path-specific calculation

The two-arg overload delegates to the three-arg overload, and the three-arg overload handles PDF existing page rotation before delegating to private PDF/image/default adjustment helpers

PR #1502, opened 2026-08-25
Better Gtk detection

### LINKED ISSUE #1501: Improve Gtk detection
When running on Ubuntu 24.04, the GTK detection doesn't appear to be working.

```
[DEBUG] 2026-08-25T14:10:45,949 @ qz.utils.GtkUtilities:42
	Finding preferred Gtk version...
[DEBUG] 2026-08-25T14:10:45,978 @ qz.utils.GtkUtilities:66
	Found Gtk0
[WARN] 2026-08-25T14:10:45,978 @ qz.utils.GtkUtilities:49
	Not a compatible Gtk version
[WARN] 2026-08-25T14:10:45,978 @ qz.utils.GtkUtilities:33
	An error occurred initializing the Gtk library
[INFO] 2026-08-25T14:10:45,979 @ qz.utils.GtkUtilities:15
	1.0
```

Starting with #1473, @IdelsTak authored a try/catch method for loading the Gtk libs directly which appears to greatly help this.  I'll cherry-pick the relevant portions to a new branch.

tresf:
Closes #1501


tresf:
Testing:
* ✅ PASS: Ubuntu 24.04 (GTK 3.24.41)
* ✅ PASS: Ubuntu MATE 16.04 (GTK 2.24.30)

src/qz/utils/GtkUtilities.java
```diff
@@ -1,124 +1,167 @@
...
-        int gtk_get_minor_version ();
+        int gtk_get_major_version();
+        int gtk_get_minor_version();
+        int gtk_get_micro_version();
 
         // Gtk 3.10-3.21
         int gdk_screen_get_monitor_scale_factor (Pointer screen, int monitor_num);
 
         // Gtk 3.22+
         Pointer gdk_display_get_primary_monitor (Pointer display);
```
IdelsTak: @tresf, could `getMonitor()` mirror `getScreen()` and reject a zero-valued display pointer before calling the GTK monitor APIs?

 `getScreen()` checks both `display != null` and `Pointer.nativeValue(display) != 0`, but `getMonitor()` only checks null before passing `display` to `gdk_display_get_primary_monitor(...)`
 
 I would suggest this check, as a result
 
 ```java
  default Pointer getMonitor() {
      Pointer display = gdk_display_get_default();
      if (display != null && Pointer.nativeValue(display) != 0) {
          // ...
      }
      log.warn("Unable to obtain primary monitor");
      return null;
  }

 ```

src/qz/utils/UnixUtilities.java
IdelsTak: Since `UnixUtilities.getScaleFactor()` no longer uses the Java-version `Toolkit` fallback, @tresf we can drop the now-unused `Version`, `Constants`, and `java.awt.*` imports

IdelsTak (commented):
On another note (not a PR blocker, though):

Java emitted a native-access warning for JNA

```
Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
Restricted methods will be blocked in a future release unless native access is enabled
```

then I checked with `-Djna.debug_load=true`...the warning appears before the GTK load path and is first triggered by JNA loading `libjnidispatch.so` during hostname lookup 

```
NetworkUtilities -> SystemUtilities.getHostName() -> UnixUtilities.getHostName()
```
so it's just a general native-access warning.  Not something caused specifically by this PR's changes

IdelsTak:
So @tresf I reviewed the changes locally

The direction looks good to me

- this PR removes the brittle `UNIXToolkit.getGtkVersion()` dependency
- probes GTK3 then GTK2 directly
- accepts only handles that pass `gtk_init_check(...)`
- falls back to `1.0` when GTK is unavailable or returns bad scale data

I also did a runtime smoke on Arch Linux / XFCE / X11: 

```text
[DEBUG] qz.utils.GtkUtilities:42
    Finding available GTK version...
[DEBUG] qz.utils.GtkUtilities:47
    Found GTK lib 'gtk-3'
[DEBUG] qz.utils.GtkUtilities:49
    Initialized GTK 3.24.52
```

 - QZ Tray started
 - server came up
 - then logs showed `Found GTK lib 'gtk-3'` then `Initialized GTK 3.24.52`

src/qz/utils/GtkUtilities.java
tresf: Thanks for catching this inconsistency.  I ran this through a couple LLMs and they seem to agree that if it's a `Pointer` type, simplifying it to `== null` is sufficient.

src/qz/utils/UnixUtilities.java
tresf: Good catch, thanks!

tresf:
> On another note (not a PR blocker, though):
> 
> Java emitted a native-access warning for JNA
> 
> ```
> Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
> Restricted methods will be blocked in a future release unless native access is enabled
> ```


Yeah it's been bugging me too...  I'll probably add that in another PR, but I would like to cover all illegal reflective operations when we do that... possibility migrate to module system.

src/qz/utils/GtkUtilities.java
IdelsTak: Yup, that's true @tresf...and on checking the actual source code its clear that jna converts address `0` to Java `null`, so your simplified `!= null` guards are sufficient
 
https://github.com/java-native-access/jna/blob/5.18.1/src/com/sun/jna/Function.java#L501

IdelsTak:
So @tresf I pulled the latest commits and ran the tests. The new GtkTests all pass locally and a runtime smoke on my Arch/XFCE/X11 machine now initializes GTK 4.22.4 cleanly

Way forward from my side: I think this pr is good to approve

Issue #1503, opened 2026-08-26
Linux: L&F defaults to Metal when Gtk is available


tresf:
Per #1473, there are scenarios where Java should attempt to load the Gtk theme even when the `UIManager.getSystemLookAndFeelClassName()` returns "Metal".

We'll cherry-pick this logic to a dedicated PR report so that the PR can be simplified.

https://github.com/qzind/tray/blob/728e940423aa140ea0e963ab4d01778d06f9018c/src/qz/utils/SystemUtilities.java#L388-L392

tresf:
Closed via #1504

PR #1504, opened 2026-08-27
Better Gtk detection (#1502)

tresf:
Improve Gtk detection via JNA

Reopened because after #1502 was merged, GTK4 seems to deadlock the Java process and this needs to be corrected before merged into master.

@IdelsTak thoughts welcome.

### DIFF
--- a/src/qz/utils/gtk/GtkType.java
+++ b/src/qz/utils/gtk/GtkType.java
@@ -1,7 +1,7 @@
 package qz.utils.gtk;
 
 enum GtkType {
-    GTK4("gtk-4", Gtk4.class),
+    // GTK4("gtk-4", Gtk4.class),
     GTK3("gtk-3", Gtk3.class),
     GTK2("gtk-x11-2.0", Gtk2.class);
 
--- a/src/qz/utils/gtk/GtkType.java
+++ b/src/qz/utils/gtk/GtkType.java
@@ -1,0 +1,15 @@
+package qz.utils.gtk;
+
+enum GtkType {
+    GTK4("gtk-4", Gtk4.class),
+    GTK3("gtk-3", Gtk3.class),
+    GTK2("gtk-x11-2.0", Gtk2.class);
+
+    final String lib;
+    final Class<? extends Gtk> type;
+
+    GtkType(String lib, Class<? extends Gtk> type) {
+        this.lib = lib;
+        this.type = type;
+    }
+}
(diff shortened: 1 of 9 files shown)


IdelsTak:
So, here are my detailed findings @tresf 

I tested this pr locally on XFCE/X11 with gtk2 2.24.33, gtk3 3.24.52, and gtk4 4.22.4 installed

Disabling gtk4 looks right to me for this PR. With GTK4 disabled, all target tests pass and the focused gtk test now selects gtk 3.24.52 as `gtk-3`, initializes it, and returns scale factor 1.0. In short, this branch avoids selecting the local gtk4

And that should be sufficient for this pr because QZ Tray only needs gtk here as a scale-factor source. And gtk3 already provides `gdk_display_get_primary_monitor(...)` / `gdk_monitor_get_scale_factor(...)` for the modern path, and GTK2 remains as the older DPI fallback

You see, gtk4 mainly adds another monitor-query path for this use case, not functionality that this PR needs

I also reproduced the deadlock behavior you saw:

1. I temporarily re-enabled gtk4
2. gtk4 alone can initialize in the focused scale-factor test
3. but the mixed path is the problem: 

- once gtk4 is initialized first, setting `GTKLookAndFeel` brings gtk3 into the same jvm
- Then the jvm emits gtk type-registration errors like `cannot register existing type 'GtkWidget'` / `GtkBuildable`, and the Java process deadlocks

So my recommendation:

Keep this pr narrow and merge it as the gtk3 -> gtk2 scale-factor recovery path with gtk4 disabled
The LaF/theming work should stay in #1503, built on top of this gtk3-only path
I'd also not remove gtk LaF work you'd started on #1503 because your changes worked well when layered over #1504 with gtk4 disabled.

src/qz/utils/gtk/GtkType.java
```diff
@@ -1,7 +1,7 @@
 package qz.utils.gtk;
 
 enum GtkType {
-    GTK4("gtk-4", Gtk4.class),
+    // GTK4("gtk-4", Gtk4.class),
```
IdelsTak, line 4: This disabled gtk4 entry now looks justified

gtk4 by itself initialized here, but when gtk4 was initialized first...setting `GTKLookAndFeel` later brought gtk3 into the same JVM resulting in a deadlock

tresf: Thanks for confirming.  We'll discuss our long terms plans of keeping this theme around in the future.

tresf: Private bug filed with Liberica OpenJDK 25 as `QZ-31` for the sizing issues.  Hopefully they confirm and file an upstream bug report for us to have on file.  If we can get the majority of sizing issues fixed in the JDK, it will allow us to remove our Gtk detection and avoid the deadlock at the source.

tresf:
Tested on Ubuntu 24.04, Fedora 40, seems OK.  Will need to revisit this in a future release when GTK4 becomes the norm.

PR #1509, opened 2026-08-28
Linux: Improve theme detection for Gnome, KDE

### LINKED ISSUE #1508: Linux: Improve Dark Theme Detection
When @IdelsTak was authoring #1473 he found seme pitfalls with dark theme detection.

I'm splitting this off into a dedicated issue to simplify the scope of that PR.

* Environment detection (KDE vs Gnome) 
* KDE dark theme detection via kdeglobals
* Better handling of "dark" variant theme names

tresf:
Closes #1508 

Test results:

Themes tested:
- ⬜  Breeze Classic |  ⬜  Breeze Light | ⬛ Breeze Dark
- ⬜ openSUSE | ⬛  openSUSE Dark
- ⬜ Adwaita | ⬛  Adwaita Dark
- ⬜ Graybird |  ⬛  Graybird Dark
- ⬜ High Contrast | ⬛  High Contrast Inverse
- ⬜ Numix

| Status | OS | Desktop | Comments |
|--------|----|----------|------------|
| ✅ PASS | Debian 13 | KDE 6.3.6 | |
| ✅ PASS | OpenSUSE TW | KDE 6.2.5 | |
| ✅ PASS | Ubuntu 24.04 | Gnome 46 | |
| ✅ PASS | Pop_OS! |  COSMIC 0.1.0 | |
| ✅ PASS | Xubuntu 24.04 |  XFCE 4.18 | |
| ✅ PASS | Manjaro |  XFCE | |
| ✅  PASS | Fedora 34 |  Gnome  40.9 | |
| ✅ PASS | FreeBSD 14.3 | XFCE 4.20 |  |


src/qz/utils/linux/Elevator.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.file.Path;
+import java.util.Arrays;
+
+public class Elevator {
+    private static final Logger log = LogManager.getLogger(Elevator.class);
+    private static final String[] KNOWN_ELEVATORS = {"pkexec", "gksu", "gksudo", "kdesudo" };
+
+    private static String foundElevator;
+
+    private static String findElevator() throws IOException {
```
IdelsTak, line 19: @tresf, Could this `throw` move behind a `foundElevator == null` check

As written, the first uncached call still throws even if `which pkexec` / `gksu` / etc succeeds

Suggested shape:

```java
private static String findElevator() throws IOException {
    if(foundElevator == null) {
        for(String elevator : KNOWN_ELEVATORS) {
            if (ShellUtilities.execute("which", elevator)) {
                foundElevator = elevator;
                break;
            }
        }
        if(foundElevator == null) {
            throw new IOException("Can't find an installed utility " + Arrays.toString(KNOWN_ELEVATORS) + " to elevate permissions.");
        }
    }
    return foundElevator;
}
```

test/qz/utils/linux/theme/KdeThemeManager.java
```diff
@@ -0,0 +1,33 @@
...
+            if(theme.trim().isEmpty()) {
+                throw new IOException("Failed to get theme name");
+            }
+        }
+        // Strip quotes
+        return theme.replace("\"", "").replace("'", "").trim();
+    }
+
+    @Override
+    public void setTheme(String themeName) throws IOException {
```
IdelsTak: Could the KDE helper skip when `plasma-apply-colorscheme` is missing?

The helper uses that command to mutate themes

On systems with `kreadconfig6` / `kreadconfig5` but no `plasma-apply-colorscheme`, this looks like a normal failure instead of an unsupported skip

test/qz/utils/linux/DeUtilsTests.java
```diff
@@ -0,0 +1,90 @@
...
+                    Assert.assertNotNull(DeThemeHelper.getDeThemeHelper(type));
+                }
+            }
+
+        }
+    }
+
+    @Test
+    public void testSaveCurrentTheme() throws IOException {
+        currentTheme = deThemeHelper.getTheme();
```
IdelsTak: Could these live theme mutation tests skip when `deThemeHelper == null`?

On my local XFCE/X11 session, the scale tests pass, but these theme tests fail with NPEs because XFCE has no helper

Image

The live theme mutation tests should skip cleanly when no helper exists

test/qz/utils/linux/ThemeUtilitiesTests.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    public void testSetLightTheme() throws IOException {
+        deThemeHelper.setTheme(false);
+        Assert.assertFalse(DeUtils.isDarkDesktop());
+        log.info("Light desktop confirmed");
+    }
+
+    @Test(priority = 99)
```
IdelsTak: Could the original theme restore move to teardown, maybe `@AfterClass(alwaysRun = true)`?

These tests mutate the real desktop theme

With restore as a later test, a failure or interrupted run can leave the theme changed

src/qz/utils/linux/CompositorType.java
```diff
@@ -0,0 +1,96 @@
...
+    CompositorType(String pattern, String ... scaleFactorCalls) {
+        this.pattern = pattern;
+        this.scaleFactorCalls = scaleFactorCalls;
+    }
+
+    static double getBestScaleFactor(boolean isSilent) {
+        double scaleFactor = 0;
+        for(CompositorType compositorType : CompositorType.getCompositors()) {
+            log.info("Trying to get scale factor using {}", compositorType);
+            if((scaleFactor = compositorType.getScaleFactor(isSilent)) != 0) {
```
IdelsTak: @tresf, there's a small scale-factor concern here

 - `0` currently means both "probe failed" and "desktop reported zero"
 - gnome/Mutter says `0` means automatic monitor scaling
 - MATE says `0` means auto-detect too

Could we separate "probe failed" from "reported automatic/default scale"?

src/qz/ui/ThemeUtilities.java
```diff
@@ -1,11 +1,77 @@
...
+                        }
+
+                        this.isDarkDesktop = isDarkDesktop;
+                        this.isDarkTaskbar = isDarkTaskbar;
+                        if(refreshAction != null) {
+                            this.refreshAction.run();
+                        }
+                    }
+                } catch (Throwable t) {
+                    log.warn("An error occurred polling the light/dark theme");
```
IdelsTak: Could this warning include at least the exception message?

The theme-change logs are useful, but this catch hides the reason polling failed

Low-noise option, for instance:

```java
log.warn("An error occurred polling the light/dark theme: {}", t.getMessage());
log.debug("Theme polling failure", t);
```

src/qz/utils/SystemUtilities.java
IdelsTak: So, @tresf, these comments still describe gtk detection / gtk dark-mode behavior

The implementation now uses FlatLaf on Linux and reads theme state from the portal path

src/qz/utils/SystemUtilities.java
```diff
@@ -477,23 +457,7 @@ public static boolean isWindowLocationValid(Rectangle window) {
      * @return Logical dpi scale as dpi/96
      */
     private static double getWindowScaleFactor(boolean forceRefresh) {
```
IdelsTak: `getWindowScaleFactor(...)` now always returns `1`, but the javadoc still says it detects logical  dpi scale

If Linux scaling now lives in early `sun.java2d.uiScale`, this comment should probably say that

src/qz/common/TrayManager.java
```diff
@@ -165,37 +168,7 @@ public TrayManager() {
...
-                                        ((JPopupMenu)c).pack();
-                                    }
-                                }
-                            });
-                        }
-                    }
-                    catch(InterruptedException ignore) {}
-                }
-            }).start();
+            new ThemeMonitor().startPolling(1000).onChange(this::refreshTheme);
```
IdelsTak, line 171: @tresf actually, `startPolling(1000)` runs before `onChange(this::refreshTheme)`

This is probably low risk, but the first immediate poll can run with no callback

src/qz/utils/linux/CompositorType.java
```diff
@@ -0,0 +1,96 @@
...
+    static CompositorType[] getCompositors() {
+        if(compositors == null) {
+            compositors = switch(LinuxUtilities.getDesktopEnvironment()) {
+                case CINNAMON -> setCompositors(MUFFIN);
+                case COSMIC -> setCompositors(COSMIC);
+                case HYPRLAND -> setCompositors(HYPRLAND);
+                case KDE -> setCompositors(KWIN6, KWIN5, MUTTER);
+                case MATE ->  setCompositors(MATE);
+                case XFCE -> setCompositors(XFCE);
+                default -> setCompositors(MUTTER); // always fallback to gsettings
```
IdelsTak: `GNOME` currently reaches `MUTTER` through the default branch

I think an explicit `case GNOME -> setCompositors(MUTTER);` would make that clearer

IdelsTak:
so, @tresf, I really like this PR's direction

Removing gtk from the Linux theme/scale path makes sense after the gtk3/gtk4 deadlock issue

For me, locally, a xfce/x11 smoke path works as:

 - xfce is detected
 - Scale resolves to 1.0 via xfconf-query
 - Portal theme reports no preference/light
 - Linux L&F resolves to FlatIntelliJLaf

I found a few things I think are worth fixing before merge, though:

 - The elevator lookup appears to always throw on first lookup even when a tool is found
  - The new live Linux tests fail on locally on my xfce setup because the theme mutation helper is `null` instead of skipped

Otherwise, I've added even more notes around

 - safer theme restoration
 - missing kde helper commands
 - `0` as auto-scale handling
 - stale gtk comments

tresf:
@IdelsTak did you forget to submit your review?  I only see your summary above.  I've addressed the bugs you've found, thank you.

IdelsTak (commented):
Oh, here are the comments/notes

IdelsTak:
> @IdelsTak did you forget to submit your review? I only see your summary above. I've addressed the bugs you've found, thank you.

Ah, thanks for catching that, I'd left them on "Pending"

src/qz/utils/linux/Elevator.java
tresf: Thanks.  Resolved in latest commit.

test/qz/utils/linux/DeUtilsTests.java
tresf: Thanks.  Resolved in latest commit.

src/qz/utils/linux/CompositorType.java
tresf: Will look into this.

src/qz/common/TrayManager.java
tresf: Yes, this was a readability choice.  I understand the risks.

src/qz/utils/linux/CompositorType.java
tresf: Agreed, but the IDE will complain if I do that saying duplicate branch..  Thoughts welcome.

tresf:
I'm re-testing and found that `gsettings get org.gnome.desktop.interface scaling-factor` doesn't work on Gnome as I thought it had, so more dbus calls are needed.

test/qz/utils/linux/theme/KdeThemeManager.java
tresf: I think this is completely fixed with the new Dispatcher handling that prefers the first successful command in most cases (gnome themes are an exception, but that's  limited to unit tests).

Please re-open if I'm wrong.

src/qz/utils/linux/CompositorType.java
tresf: I think this is completely fixed with the new Dispatcher handling that prefers the first successful command in most cases (gnome themes are an exception, but that's limited to unit tests).

Please re-open if I'm wrong.

src/qz/utils/SystemUtilities.java
tresf: Fixed, thanks!

src/qz/utils/linux/CompositorType.java
tresf: Switched to Enum(Supplier<Dispatcher>) with explicit Unknown dispatcher as fallback.  Should be all set.

src/qz/utils/SystemUtilities.java
tresf: Thanks for this catch.  Yeah, no custom logic is needed after this PR because we rely 100% on Java to do it for us.  Done.

src/qz/ui/ThemeUtilities.java
tresf: Done.

tresf:
Fedora 34 and FreeBSD 14.3 are passing now.  I think we're in really good shape.

~~Awaiting a second set of eyes by @IdelsTak~~

**Edit:** Need to move this along... Merging as-is.

PR #1473, opened 2026-07-10
Support for dark About dialog logos

IdelsTak:
@tresf, here's what has changed:

- Use `qz-logo-dark.png` for the About dialog logo on dark desktops when bundled
- Generate a cached inverted `qz-logo.png` fallback when `qz-logo-dark.png` is absent
- Refresh the About dialog logo when the dialog/theme refreshes
- Improve Linux dark-theme detection for GNOME/KDE theme names
- Select Swing GTK Look and Feel explicitly on Linux when GTK is available
- Keep the implementation on the current About path: `LOGO_ICON` / `qz-logo.png`
- Leave `BANNER_ICON` / `qz-banner.png` behavior unchanged

Closes https://github.com/qzind/tray/issues/1440

PR #1513, opened 2026-09-07
Baseline support for dark images

### LINKED ISSUE #1440: About banner: Better handling of dark images
From a customer:

>  it would [...] be great to have that feature available [to invert the logo on a dark background]. Our product design team isn't concerned about the potential for an "undesirable hue," so that wouldn't be a blocker for us

The customer is talking about a dark "about" banner logo on a dark background.  QZ Tray would need a way to provide either "dark" assets or color inversion. 

This is feature is predominantly for supporting dark banners on the Company Branded portal.

https://github.com/user-attachments/assets/582d45b8-1483-4a77-875d-6e1250414332

tresf:
Adds support for split light/dark mode images.

Before:

* `ui/resources/qz-logo.png`

After:

* `ui/resources/qz-logo.png`
* `ui/resources/qz-logo-dark.png` (optional)

Internally, this PR only addresses the About dialog which uses `qz-logo.png`.  This means that code changes are required to use dark logs elsewhere in the software, however this adds baseline support for dark images by adding a new `boolean isDark` parameter to all calls to IconCache's `getIcon(...)` or similar functions.

Inspired by #1473 
Supersedes #1473 
Closes #1440

### DIFF
--- a/src/qz/ui/AboutDialog.java
+++ b/src/qz/ui/AboutDialog.java
@@ -293,12 +293,19 @@ public class AboutDialog extends BasicDialog implements Themeable {
     }
 
     private void refreshHeader() {
-        headerBar.setBackground(SystemUtilities.isDarkDesktop() ?
-                                        Constants.TRUSTED_COLOR.darker().darker() : Constants.TRUSTED_COLOR_DARK);
-        headerBar.setVisible(Substitutions.areActive());
-        pack();
+        if(headerBar != null) {
+            headerBar.setBackground(SystemUtilities.isDarkDesktop()?
+                                            Constants.TRUSTED_COLOR.darker().darker():Constants.TRUSTED_COLOR_DARK);
+            headerBar.setVisible(Substitutions.areActive());
+            pack();
+        }
     }
 
+    private void refreshLogo() {
+        if(logo != null) {
+            logo.setIcon(getIcon(IconCache.Icon.LOGO_ICON, SystemUtilities.isDarkDesktop()));
+        }
+    }
 
     @Override
     public void setVisible(boolean visible) {
@@ -312,7 +319,7 @@ public class AboutDialog extends BasicDialog implements Themeable {
     @Override
     public void refresh() {
         refreshHeader();
-        logo.setIcon(getIcon(IconCache.Icon.LOGO_ICON, SystemUtilities.isDarkDesktop()));
+        refreshLogo();
         super.refresh();
     }
 }
--- a/src/qz/ui/AboutDialog.java
+++ b/src/qz/ui/AboutDialog.java
@@ -40,6 +40,7 @@ public class AboutDialog extends BasicDialog implements Themeable {
     private static final Logger log = LogManager.getLogger(AboutDialog.class);
     private final boolean limitedDisplay;
     private Server server;
+    private JLabel logo;
     private JLabel lblUpdate;
     private JButton updateButton;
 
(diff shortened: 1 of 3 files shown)


src/qz/ui/AboutDialog.java
```diff
@@ -311,6 +312,7 @@ public void setVisible(boolean visible) {
     @Override
     public void refresh() {
         refreshHeader();
+        logo.setIcon(getIcon(IconCache.Icon.LOGO_ICON, SystemUtilities.isDarkDesktop()));
```
IdelsTak: @tresf, could we have a `null` guard around `logo` here?

See, `aboutDialog` is added to `componentList` before `setServer()` initializes the About UI

So a theme refresh in that startup gap can call `refresh()` before `logo` exists

src/qz/ui/component/IconCache.java
```diff
@@ -129,11 +139,23 @@ public String getId(Dimension size) {
             return getId();
```
IdelsTak: Ideally, shouldn't this fallback preserve `isDark`?

Because, if a dark sized lookup has no exact size match...this currently falls back to the light id via `getId()`

So, @tresf, maybe `return getId(isDark)` here?

src/qz/ui/component/IconCache.java
```diff
@@ -253,16 +299,13 @@ public static Icon[] getTypes() {
      * @return The BufferedImage representing the data
      */
     private static BufferedImage getImageResource(String imagePath) {
```
IdelsTak, line 311: @tresf, could missing light resources still warn while missing dark variants stay quiet?

The dark files are optional, but, the base light files are still required

See, as it is, this shared loader makes both cases silent

src/qz/ui/AboutDialog.java
tresf: Good catch.  This is appears to be bad historical lifecycle for the About dialog.  Dialogs should initiate all of their UI components on creation.  To avoid scope creep and keep changes minimal.... I'll simply add the `null` check.  Thanks.

src/qz/ui/component/IconCache.java
tresf: ~~Did you comment on the correct line here?  I don't see the problem here... First, the method `getId(Dimension size)`  isn't currently used in-code (was kept for consistency compat) but all calls that do not provide `isDark` default to light, so this is identical to other helpers we have.  Please clarify.~~

tresf: Good catch.  Sorry about the previous comment, the preview in GitHub was weird.

src/qz/ui/component/IconCache.java
tresf: Done.

tresf:
All comments addressed.  Also added resizing support for dark/light images which can occur with the system tray or taskbar (although the system tray already has some support for this, so some of this may be reundant).

Moving forward, I think we can drastically reduce IconCache's size and complexity by internally levering an SVG library in the future.  I'll await more comments before merging.

IdelsTak:
> All comments addressed. Also added resizing support for dark/light images which can occur with the system tray or taskbar (although the system tray already has some support for this, so some of this may be reundant).
> 
> Moving forward, I think we can drastically reduce IconCache's size and complexity by internally levering an SVG library in the future. I'll await more comments before merging.

@tresf, all the 3 review concerns look addressed to me now

Also the added dark/light resizing support makes sense to me as a baseline, even if some tray paths already cover part of it

I'd now say that no further blockers are evident from what I see

Oh, also yeah, I agree the svg idea makes sense as a future cleanup

Anyway, for this PR, I think keeping it to the png light/dark baseline is the right scope

Again, also, that `IconCache` complexity you've brought up is worth revisiting later. But I wouldn’t block this PR on that

Issue #1456, opened 2026-06-02
Linux: Better System Tray support


tresf:
Improve the System Tray handling for Linux Desktops, possibly through a D-Bus bridge.

* The current implementation uses a JPopupMenu attached to a 1px x 1px JFrame that attempts to appear when the Task Launcher is clicked.

@IdelsTak will be creating a proof of concept using D-Bus.

Target desktops:
* [x] Gnome
* [x] KDE
* [x] XFCE
* [x] COSMIC
* [ ] Pantheon

Stretch goals:
* [x] Support "symbolic" icon types https://gitlab.gnome.org/GNOME/gtk/-/work_items/1762
* [x] Update notification system to be "native" as well
* [ ] Allow "symbolic" icons to change color based on QZ Tray status (red, yellow)

<img width="2282" height="1576" alt="Image" src="https://github.com/user-attachments/assets/7fc0a887-3939-4d50-9b8e-81ffc19e3e23" />

IdelsTak:
Added Budgie tray support:

https://github.com/user-attachments/assets/2667733f-0619-4614-93ae-12746130b0c6

IdelsTak:
Added Gnome tray support:

https://github.com/user-attachments/assets/c88b3f9c-45f4-455a-86a6-00408792ce77

IdelsTak:
Added KDE tray support:

https://github.com/user-attachments/assets/5c23b7f7-04c2-47c6-9a80-a73bae00d4cf

IdelsTak:
Added MATE tray support:

https://github.com/user-attachments/assets/bef2b157-a51c-44c6-b693-5fe06f50fdf0

IdelsTak:
Added Cinnamon tray support:

https://github.com/user-attachments/assets/44ab48f7-9eb7-45dd-8ab8-a628913fd764

IdelsTak:
Added XFCE tray support:

https://github.com/user-attachments/assets/d55362f3-cd36-4757-81c8-377f8c764b40

IdelsTak:
Added LXQt tray support:

https://github.com/user-attachments/assets/80eed27b-66bd-4ae3-9eaa-99733d524eb7

IdelsTak:
Added COSMIC tray support (not yet production ready, still buggy):

https://github.com/user-attachments/assets/8747dd7f-12a1-4411-8faf-9339b93dda92

IdelsTak:
[...]

PR #1515, opened 2026-09-13
Initial support for SVG images

tresf:
Leverages jsvg to load images from SVG instead of PNG.

* Uses weisj/jsvg render
* Scales proportionally to the provided height, or use the SVGs natural size if not provided
* Removes unused `qz-warning[_size].png` assets
* Adds mask support for `qz-severe[_size].png` (renamed to `tray-loading.svg`)
* Adds support for ANY asset to have a `-dark` version (e.g. `trust-issue.svg` vs `trust-issue-dark.svg` -- most are dormant for future provisioning)

Conversion chart:

| QZ Tray < = 2.3.0 | QZ Tray >= 2.3.1 |
|------------------|------------------|
|`qz-severe[_size].png`  | `tray-loading.svg` |
|`qz-warning[_size].png`  | -- |
|`qz-default[_size].png`  | `tray-ready.svg` |
| --  | `tray-loading-mask.svg` |
|`qz-mask[_size].png`  | `tray-ready-mask.svg` |
|`qz-logo.png`  | `logo.svg` |
| -- | `logo-dark.svg` |

TODO:
* [x] Refactor IconCache to be singleton and not rely on TrayManager<br>Done, but to reduce the number of files touched in this PR, I'm keeping the TrayManager instance around.
* [x] Cleanup IconCache code to better differentiate PNG, SVG
* [x] Add sane fallback before for IconCache when a resource is missing
* [x] Cleanup dark/light/sizing logic to be easier to follow
* [x] Test coverage

### DIFF
--- a/src/qz/ui/component/IconCache.java
+++ b/src/qz/ui/component/IconCache.java
@@ -10,6 +10,9 @@
 
 package qz.ui.component;
 
+import com.github.weisj.jsvg.SVGDocument;
+import com.github.weisj.jsvg.parser.SVGLoader;
+import com.github.weisj.jsvg.view.FloatSize;
 import com.github.zafarkhaja.semver.Version;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -22,6 +25,7 @@ import java.awt.*;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.*;
 import java.util.List;
 
@@ -75,9 +79,10 @@ public class IconCache {
         QUESTION_ICON("qz-question.png"),
 
         // Banner
-        LOGO_ICON("qz-logo.png"),
+        LOGO_ICON(260, "qz-logo.svg"),
         BANNER_ICON("qz-banner.png");
 
+        private Integer preferredHeight;
         private boolean padded = false;
         private String[] fileNames;
         private String[] fileNamesDark;
@@ -87,12 +92,17 @@ public class IconCache {
          *
          * @param fileNames path(s) to image
          */
-        Icon(String ... fileNames) {
+        Icon(Integer preferredHeight,  String ... fileNames) {
+            this.preferredHeight = preferredHeight;
             this.fileNames = fileNames;
             this.fileNamesDark = new String[fileNames.length];
             for(int i = 0; i < fileNames.length; i++) {
                 this.fileNamesDark[i] = String.format("%s" + DARK_SUFFIX + ".%s", getBaseName(fileNames[i]), getExtension(fileNames[i]));
             }
+        }
+
+        Icon(String ... fileNames) {
+            this(null, fileNames);
         }
 
         /**
@@ -157,6 +167,8 @@ public class IconCache {
 
         public String[] getIds() { return getIds(false); }
 
+        public Integer getPreferredHeight() { return preferredHeight; }
+
         private void addId(String id, boolean isDark) {
             if(isDark) {
                 fileNamesDark = Arrays.copyOf(fileNamesDark, fileNamesDark.length + 1);
@@ -197,8 +209,15 @@ public class IconCache {
             for(int j = 0; j < lightIds.length; j++) {
                 String lightId = lightIds[j];
                 String darkId = darkIds[j];
-                BufferedImage lightBi = getImageResource(RESOURCES_DIR + lightId);
-                BufferedImage darkBi = getImageResource(RESOURCES_DIR + darkId);
+                BufferedImage lightBi;
+                BufferedImage darkBi;
+                if(lightId.contains(".svg")) {
+                    lightBi = getImageResourceFromSvg(i.getPreferredHeight(), RESOURCES_DIR + lightId);
+                    darkBi = getImageResourceFromSvg(i.getPreferredHeight(), RESOURCES_DIR + lightId);
+                } else {
+                    lightBi = getImageResource(RESOURCES_DIR + lightId);
+                    darkBi = getImageResource(RESOURCES_DIR + darkId);
+                }
                 // Cache light icons
                 imageIcons.put(lightId, new ImageIcon(lightBi));
                 images.put(lightId, lightBi);
@@ -309,8 +328,8 @@ public class IconCache {
      * @return The BufferedImage representing the data
      */
     private static BufferedImage getImageResource(String imagePath) {
-       try(InputStream is = IconCache.class.getResourceAsStream(imagePath.replace("#", ""))) {
-            if (is != null) {
+        try(InputStream is = IconCache.class.getResourceAsStream(imagePath.replace("#", ""))) {
+           if (is != null) {
                 return ImageIO.read(is);
             } else {
                 if(!getBaseName(imagePath).endsWith(DARK_SUFFIX)) {
@@ -320,6 +339,41 @@ public class IconCache {
         }
         catch(IOException e) {
             log.error("Cannot load {}", imagePath, e);
+        }
+        return null;
+    }
+
+    public static BufferedImage getImageResourceFromSvg(Integer preferredHeight, String svgPath) {
+        URL url = IconCache.class.getResource(svgPath);
+        if (url != null) {
+            SVGLoader loader = new SVGLoader();
+            SVGDocument svgDocument = loader.load(url);
+            if(svgDocument != null) {
+                FloatSize svgSize = svgDocument.size();
+                float w = svgSize.width;
+                float h = svgSize.height;
+
+                // scale proportionally
+                if(preferredHeight != null) {
+                    w = svgSize.width * (preferredHeight / h);
+                    h = preferredHeight;
+                }
+
+                BufferedImage image = new BufferedImage((int)w, (int)h, BufferedImage.TYPE_INT_ARGB);
+                Graphics2D g = image.createGraphics();
+                g.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
+                g.setRenderingHint(RenderingHints.KEY_STROKE_CONTROL, RenderingHints.VALUE_STROKE_PURE);
+
+                double scaleX = w / svgSize.width;
+                double scaleY = h / svgSize.height;
+                g.scale(scaleX, scaleY);
+                svgDocument.render(null, g);
+                g.dispose();
+                return image;
+            }
+        }
+        if(!getBaseName(svgPath).endsWith(DARK_SUFFIX)) {
+            log.warn("SVG '{}' could not be loaded", svgPath);
         }
         return null;
     }
--- a/ant/dependencies.properties
+++ b/ant/dependencies.properties
@@ -52,6 +52,10 @@ jssc.org=io.github.java-native
 jssc.name=jssc
 jssc.rev=2.10.3
 
+jsvg.org=com.github.weisj
+jsvg.name=jsvg
+jsvg.rev=2.2.0
+
 libusb4java.org=org.usb4java
 libusb4java.name=libusb4java
 libusb4java.rev=1.2.0
--- a/ant/ivy.xml
+++ b/ant/ivy.xml
@@ -88,6 +88,7 @@
         <dependency name="imageio-core" org="com.twelvemonkeys.imageio" conf="*->default" rev="${twelvemonkeys.rev}"/>
         <dependency name="imageio-jpeg" org="com.twelvemonkeys.imageio" conf="*->default" rev="${twelvemonkeys.rev}"/>
         <dependency name="imageio-metadata" org="com.twelvemonkeys.imageio" conf="*->default" rev="${twelvemonkeys.rev}"/>
+        <dependency name="jsvg" org="${jsvg.org}" conf="*->default" rev="${jsvg.rev}"/>
 
         <dependency name="jai-imageio-core" org="${jai.org}" conf="*->default" rev="${jai.rev}"/>
         <dependency name="jai-imageio-jpeg2000" org="${jai.org}" conf="*->default" rev="${jai.rev}"/>
--- a/src/qz/ui/resources/qz-logo.svg
+++ b/src/qz/ui/resources/qz-logo.svg
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+<svg id="svg4208" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http://www.w3.org/2000/svg" height="128" width="128" version="1.1" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" viewBox="0 0 128.00001 128.00001">
+ <metadata id="metadata4213">
+  <rdf:RDF>
+   <cc:Work rdf:about="">
+    <dc:format>image/svg+xml</dc:format>
+    <dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/>
+    <dc:title/>
+   </cc:Work>
+  </rdf:RDF>
+ </metadata>
+ <g id="layer1" transform="translate(-125.25 -253.66)">
+  <path id="XMLID_48_" d="m253.07 310.83v2.3662l-0.78875 3.9437-5.5211 32.338-65.465 30.761v-41.014c0-5.5212-2.3662-11.042-5.5212-14.197l-0.78873-0.78874-49.69-46.535 73.352-22.085 48.113 40.226 0.78875 0.78875c3.1549 3.1549 5.5211 7.8874 5.5211 14.197z" fill="#44aa53"/>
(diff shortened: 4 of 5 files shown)


tresf:
Assets have been switched to SVG but this seems to have caused a quality regression... Will need to investigate...

image

tresf:
Much better...  (Still need to fix artifacting)

tresf:
Artifacting fixed

image


image

tresf:
Added template-favorable "loading" icon, replaced the red "severe" icon.

https://github.com/user-attachments/assets/4e474c24-bfb1-4d71-a5fa-e76a19479816

src/qz/utils/ImageUtilities.java
```diff
@@ -0,0 +1,135 @@
...
+     * @param path The file path of the SVG to load
+     * @param size The desired size to scale the SVG to, or the natural SVG size if <code>null</code>
+     * @param relativeClass Class used for looking up relative resources
+     * @return The BufferedImage representing the data
+     */
+    public static BufferedImage imageFromSvgResource(Path path, Integer size, Object relativeClass) {
+        URL url = relativeClass.getClass().getResource(path.toString());
+        if (url != null) {
+            SVGLoader loader = new SVGLoader();
+            SVGDocument svgDocument = loader.load(url);
```
IdelsTak, line 87: @tresf, could this svg path catch parser/render failures...and then return `null` like the png loader path does?

You see, a malformed or future svg would currently escape during `IconCache` construction

And since png/svg fallback is chosen before decode, a bad existing SVG also would not retry an old png fallback name

I find that:
- `imageFromSvgResource(...)` handles missing resource urls and null svgs
- yet it does not catch `SVGLoader.load(...)` or `svgDocument.render(...)` failures
- then `IconCache.buildImageCache()` calls this while constructing the cache. So one bad svg can break broader icon initialization
- and that's noting that the png/svg fallback decision happens during resource discovery...not after decode/render failure

src/qz/utils/ImageUtilities.java
```diff
@@ -0,0 +1,135 @@
...
+                double w = svgSize.getWidth();
+                double h = svgSize.getHeight();
+
+                // scale proportionally
+                if(size != null) {
+                    w = svgSize.getWidth() * (size / h);
+                    h = size;
+                }
+
+                BufferedImage image = new BufferedImage((int)w, (int)h, BufferedImage.TYPE_INT_ARGB);
```
IdelsTak, line 99: Here @tresf, could we round or ceil the svg raster dimensions before creating the `BufferedImage`?

This is because some source svgs are non-square/fractional...and flooring the computed width/height could clip edge pixels at tray sizes

For instance, `tray-ready-mask.svg` has a `height="15.604px"`

Although in my local visual inspection looked this one looked ok, but exported alpha bounds touch the canvas edges for tray/mask icons

So, I'd say that rounding/ceiling is a small defensive improvement for future svgs with fractional natural sizes

src/qz/ui/tray/TrayType.java
```diff
@@ -50,9 +52,10 @@ private static Image blankImage() {
...
     public boolean getTaskbar() { return taskbar != null; }
 
-    public void setIcon(IconCache.Icon icon) {
+    public void setIcon(IconCache.Icon requestedIcon) {
+        IconCache.Icon icon = ThemeUtilities.getTrayIcon(requestedIcon);
         if (isTray()) {
-            tray.setImage(iconCache.getImage(icon, tray.getSize()));
+            tray.setImage(iconCache.getImage(icon, tray.getSize(), SystemUtilities.isDarkTaskbar(false)));
         } else {
             taskbar.setIconImages(iconCache.getImages(icon));
```
IdelsTak, line 63: In this instance, @tresf, is it intentional that the taskbar fallback icon list always uses the light/default images?

That real tray path from the other `if` branch passes `SystemUtilities.isDarkTaskbar(false)`...yet the taskbar fallback still calls `getImages(icon)` with no dark flag

Nonetheless, this is about intended dark-mode behavior...not a breakage per se

IdelsTak (commented):
Just a small cleanup here @tresf:

https://github.com/qzind/tray/blob/968d53b3e88ec3c2676deb13b6631a92526103e3/src/org/jdesktop/swinghelper/tray/JXTrayIcon.java#L171

that method no longer exists after the icon-cache refactor

IdelsTak (commented):
I also noticed something @tresf:

https://github.com/qzind/tray/blob/968d53b3e88ec3c2676deb13b6631a92526103e3/src/qz/ui/AboutDialog.java#L306

I've been wondering: the About dialog is wired to refresh its logo on theme changes...but this current PR's resources only include `logo.png`

As a result, dark mode falls back to the same logo image

Well, that seems fine for baseline svg support, but it means that this PR does not yet provide a visibly different dark About logo

src/qz/ui/component/IconCache.java
```diff
@@ -10,454 +10,318 @@
...
-                imageIcons.put(lightId, new ImageIcon(lightBi));
-                images.put(lightId, lightBi);
-                // Cache dark icons if present
-                imageIcons.put(darkId, new ImageIcon(darkBi != null ? darkBi : lightBi));
-                images.put(darkId, darkBi != null ? darkBi : lightBi);
-            }
-        }
-        // Stash scaled 2x, 3x versions if missing
-        int maxScale = 3;
+    Map<String, BufferedImage> buildImageCache() {
```
IdelsTak, line 224: Something to consider @tresf: would you be open to a small headless `IconCache`/`ImageUtilities` test here or as a follow-up?

The risky parts look testable without a desktop: 
- cache construction
- svg render dimensions
- dark fallback
- sized lookup
- and, icon-to-base64 output

See, this pr replaces many pngs and centralizes image loading...so, a small test would protect the new path

src/qz/common/TrayManager.java
```diff
@@ -557,12 +549,7 @@ public void displayInfoMessage(String text) {
      * Thread safe method for setting the default icon
      */
     public void setDefaultIcon() {
-        // Workaround for JDK-8252015
-        if(SystemUtilities.isMac() && Constants.MASK_TRAY_SUPPORTED && !MacUtilities.jdkSupportsTemplateIcon()) {
-            setIcon(DEFAULT_ICON, () -> MacUtilities.toggleTemplateIcon(tray.tray()));
-        } else {
-            setIcon(DEFAULT_ICON);
-        }
+        setIcon(DEFAULT_ICON);
```
IdelsTak, line 552: Well, @tresf, since this path no longer does the old `MacUtilities.toggleTemplateIcon(...)` workaround after setting the default icon...is that no longer needed with the new mask svg routing?

(Personally, I can't verify that from my arch/xfce setup locally)

src/qz/ui/component/IconCache.java
```diff
@@ -10,454 +10,318 @@
...
-                return ImageIO.read(is);
-            } else {
-                if(!getBaseName(imagePath).endsWith(DARK_SUFFIX)) {
-                    throw new IOException("InputStream is null");
-                }
-            }
-        }
-        catch(IOException e) {
-            log.error("Cannot load {}", imagePath, e);
+    public static IconCache getInstance() {
```
IdelsTak: Small observation here @tresf, this intended lazy singleton first-use is not synchronized

Nonetheless, cache contents should be equivalent and no practical failure was observed

src/qz/utils/ImageUtilities.java
tresf: > one bad svg can break broader icon initialization

This is intended.  We want to fail hard if we're missing a required resource.  Fallback behavior is for Company Branded (whitelabel) customers so they can still build the software.... We'll need to add some special logic in our Company Branded scripts to remove our SVGs when PNGs are provided.  (I had initially considered looking for PNGs first, but this seemed like a forever performance penalty for a temporary compat code/feature).

src/qz/utils/ImageUtilities.java
tresf: > tray-ready-mask.svg has a height="15.604px"

Thanks for pointing this out.  This is a bug with the asset.   Fixed and pushed.

> could we round or ceil the svg raster dimensions before creating the BufferedImage?

Hmm... I'm not sure.  Let me look at the code again.

tresf: I think our logic is correct as it allows non-square images to keep their aspect ratio.  I assume height is more important than width when laying out components such as the About dialog, the OK button, etc.  Please reopen if you think this comment was resolved in error.

src/qz/ui/tray/TrayType.java
tresf: Correct, the idea of a mask or 'template' icon is uncommon and not an interface standard for a Task Bar, but it is documented for macOS and appears to be a defacto standard for most System Tray implementations including but not limited to Windows 10+, Gnome, KDE, etc.

We will revisit how to implement this once #1456 is resolved.

tresf:
> this PR does not yet provide a visibly different dark About logo

I forgot the most important SVG!  Whoops!  Now I've added both, thanks!  Didn't know what to do for our own "logo-dark.svg", opinions welcome.

src/qz/ui/component/IconCache.java
tresf: Good idea, fixed.

src/qz/ui/component/IconCache.java
tresf: I considered adding coverage.  Help welcome here if interested!

src/qz/common/TrayManager.java
tresf: I wrote a helper that intercepts it instead here.  It's not as obvious as the old helper but it effectively does the same thing.

https://github.com/qzind/tray/blob/bbdf70720e64ec2dc8d71b3b93639179bf84c866/src/qz/ui/ThemeUtilities.java#L126-L135


The helper is registered in `TrayType.setIcon()`.  I don't love the unintuitive replacement here, but I wanted to avoid touching too many files.  Thoughts welcome.

tresf:
I think I addressed all comments.  Let me know if I missed something!

src/qz/utils/ImageUtilities.java
IdelsTak: That makes sense @tresf, I’ll treat this one as resolved from my side

src/qz/utils/ImageUtilities.java
IdelsTak: I'm fine with keeping the sizing logic as-is if the assets are normalized

By visual inspection anyway, the original fractional-height concern looks addressed by the asset fix

src/qz/ui/tray/TrayType.java
IdelsTak: That sounds reasonable to defer to #1456

I'll treat the taskbar fallback behavior as intentional for this PR

src/qz/ui/component/IconCache.java
IdelsTak: Understood

I'll treat focused `IconCache`/`ImageUtilities` coverage as follow-up work...that's unless you want it folded into this PR before merge

src/qz/common/TrayManager.java
IdelsTak: Thanks @tresf, that path makes sense statically

I can't runtime-verify macOS from my side, but the replacement path is clear now

src/qz/ui/component/IconCache.java
IdelsTak: Confirmed, `getInstance()` is synchronized now

Resolved from my side

IdelsTak:
> > this PR does not yet provide a visibly different dark About logo
> 
> I forgot the most important SVG! Whoops! Now I've added both, thanks! Didn't know what to do for our own "logo-dark.svg", opinions welcome.

@tresf, I did a xfce/x11 smoke check...

On the visual side, `logo-dark.svg` is distinct

But the contrast change is subtle since the main face and white paper stay the same

The lower/inner green faces are lighter, so it reads a bit better on dark surfaces...but it's not a dramatic dark-mode variant

I'm fine with it for baseline SVG support...but if the goal is a clearly different dark logo, I'd probably push the dark variant a bit further

tresf:
Hmm.. in my opinion there's a slight degradation in the new tray icon.  I'll attempt to adjust before merging.

https://github.com/user-attachments/assets/a22d49f5-c8f3-4fc5-8d26-4eb2c688690d

tresf:
> I'm fine with it for baseline SVG support...but if the goal is a clearly different dark logo, I'd probably push the dark variant a bit further

I worked on a "dark" version for a while and then just deleted it. :D

This branch has been significantly updated with some test coverage.  I understand most of this is redundant with #1522, sorry for the duplicate effort.  These tests are not as easy to read as yours, but should be comprehensive and also cover mixed-images formats (SVG + PNG) which will occur for our company branded clients. The unit tests also exposed some incorrectly sized PNGs on master branch as well as a broken size information for the system tray icons and task bar icons (both were missing 16x16).   This reinforces your instinct to add coverage, otherwise we would have experienced these bugs after merge.

In order to test multiple image formats, I had to clone the resources directory and allow an override resources path in IconCache.  This allowed me to fix some incorrectly placed logic in IconCache.Icon... since enumerated types are inherently static. 

Thanks for the reviews... I think we're very close to merging.

src/qz/ui/component/IconCache.java
tresf: I've added coverage to this PR.

IdelsTak:
> > I'm fine with it for baseline SVG support...but if the goal is a clearly different dark logo, I'd probably push the dark variant a bit further
> 
> I worked on a "dark" version for a while and then just deleted it. :D
> 
> This branch has been significantly updated with some test coverage. I understand most of this is redundant with #1522, sorry for the duplicate effort. These tests are not as easy to read as yours, but should be comprehensive and also cover mixed-images formats (SVG + PNG) which will occur for our company branded clients. The unit tests also exposed some incorrectly sized PNGs on master branch as well as a broken size information for the system tray icons and task bar icons (both were missing 16x16). This reinforces your instinct to add coverage, otherwise we would have experienced these bugs after merge.
> 
> In order to test multiple image formats, I had to clone the resources directory and allow an override resources path in IconCache. This allowed me to fix some incorrectly placed logic in IconCache.Icon... since enumerated types are inherently static.
> 
> Thanks for the reviews... I think we're very close to merging.

@tresf, I pulled the latest [svg](https://github.com/qzind/tray/tree/svg) branch and ran the updated tests

And the new `IconCacheTests` pass locally

So I'll treat #1522 as superseded for now

tresf:
This is currently failing for Company Branded builds... 

```
Caused by: java.lang.RuntimeException: java.io.IOException: Could not find a mandatory resource under any of the following names 'tray-ready', 'qz-mask' [PNG, SVG]
```

... however `qz-mask.png` is there... This makes me concerned that our unit tests are flawed.

Extracted the JAR:
```
% ls -l qz/ui/resources/*mask*
-rw-r--r--  1 owner  staff  1348 Sep 20  2026 qz/ui/resources/qz-mask-20.png
-rw-r--r--  1 owner  staff  1448 Sep 20  2026 qz/ui/resources/qz-mask-24.png
-rw-r--r--  1 owner  staff  1694 Sep 20  2026 qz/ui/resources/qz-mask-32.png
-rw-r--r--  1 owner  staff  1910 Sep 20  2026 qz/ui/resources/qz-mask-40.png
-rw-r--r--  1 owner  staff  2159 Sep 20  2026 qz/ui/resources/qz-mask-48.png
-rw-r--r--@ 1 owner  staff  1233 Sep 20  2026 qz/ui/resources/qz-mask.png
```

tresf:
Unit tests are good, it's the integration tests that are missing... hmm... 

**Edit:** Added integration tests.  The calls to `getResourceAsStream` behave differently when run from a JAR.   Changed the path calculation to accommodate.

tresf:
> Unit tests are good, it's the integration tests that are missing... hmm...
> 
> **Edit:** Added integration tests. The calls to `getResourceAsStream` behave differently when run from a JAR. Changed the path calculation to accommodate.

... aaand behave differently on Windows too due to Path defaulting to backslashes.  Ugh.  Added helper.   Note that IconCache really should move out of Component back into the root of `ui/` but doing so would touch way more classes than I wanted to for this PR.

src/qz/ui/component/IconCache.java
```diff
@@ -22,442 +26,385 @@
...
-                images.put(id.replaceAll("mask", "default"), clone);
-                imageIcons.put(id.replaceAll("mask", "default"), new ImageIcon(clone));
+        Theme theme = Theme.get(isDark);
+        if(format == Icon.Format.SVG) {
+            Path resource = findFile(format, theme, i.names);
+            resource = resource == null ? findFile(format, Theme.LIGHT, i.names) : resource; // fallback
+            if(resource != null) {
+                Path svgPath = Files.createTempFile(String.format("%s-", id), Icon.Format.SVG.extension());
+                svgPath.toFile().deleteOnExit();
+                FileUtilities.configureAssetToFile(ThemeUtilities.class, resource.toString(), new HashMap<>(), svgPath.toFile());
```
IdelsTak: @tresf, a quick non-blocking thought here:

Most resource reads now go through the normalized `ImageUtilities` helper

This extraction path still passes `resource.toString()` into `configureAssetToFile(...)`

The CI is green on Windows, so I would not block on this, though

But using the same normalized resource-path helper here too may keep the lookup story easier to reason about

tresf: Ok, well, I cleaned this up now, but it is a major refactor. 🤦

PR #1520, opened 2026-09-16
Support legacy ANSI charsets via encoding: 'legacy'

### LINKED ISSUE #1519: QZ Tray 2.3.0:  Legacy code pages are no longer assumed
The following support issue came in after the release of QZ Tray 2.3.0:

> We have identified an issue after upgrading QZ Tray to version 2.3.0.  When printing ferry tickets through QZ Tray 2.3.0, Greek characters are not rendered correctly. The same print process works correctly with the immediately previous version, QZ Tray 2.2.6.  Based on our tests, this appears to be a regression introduced in version 2.3.0.

This is because in QZ Tray 2.2.6, we used Java 11 which defaulted to Cp1253 on a "Greek" desktop.

Starting with QZ Tray 2.3.0, we use Java 25 which defaults to UTF-8 on all desktops.

The impact is, customers using the RAW API with an `encoding: null` would automatically use the correct ANSI encoding.  This assumption breaks with Java 25 as this legacy behavior has been removed.

Ideally, we'll have an `encoding: legacy` for qz-tray.js@2.3.1+ to shoehorn this old  behavior in.

tresf:
* Windows; Adds a new encoding called `legacy` for QZ Tray 2.3.1+
* Compatible with older versions by removing this if provided
* Logs a warning if attempted on a non-Windows OS

```js
var config = qz.configs.create("Printer Name", { encoding: 'legacy' });
```

Backwards-compat patch to fix  #1519


src/qz/utils/WindowsUtilities.java
```diff
@@ -756,4 +763,54 @@ public static boolean isAdminOwned(Path path) {
...
+                int ret = Kernel32Ex.INSTANCE.GetLocaleInfoEx(localeName, lctype, buffer, 2);
+
+                if (ret > 0) {
+                    return buffer.getInt(0);
+                }
+            }
+        } catch(UnsatisfiedLinkError | NoClassDefFoundError ignore) {}
+
+        // Fallback to PowerShell
+        String output = ShellUtilities.executeRaw("(Get-WinSystemLocale).TextInfo.AnsiCodePage");
```
IdelsTak: @tresf, could this fallback call PowerShell explicitly?

Currently, it looks like it asks `Runtime.exec(...)` to run `(Get-WinSystemLocale).TextInfo.AnsiCodePage` as the program name

The existing PowerShell fallbacks in this file use `powershell.exe`, `-Command`, `...`

Something like this seems safer:

```java
String output = ShellUtilities.executeRaw(
        "powershell.exe",
        "-Command",
        "(Get-WinSystemLocale).TextInfo.AnsiCodePage"
);

File: js/qz-tray.js [hunk 2]
```diff
@@ -1030,6 +1030,11 @@ var qz = (function() {
                         delete config.spool;
                     }
                 }
+                if(_qz.tools.versionCompare(2, 3, 0, 4) < 0) {
+                    if(config.encoding === 'legacy') {
```
IdelsTak: @tresf, should this compatibility downgrade also cover object-form and case variants?

The java side accepts `legacy` case-insensitively...and `PrintOptions` now routes `encoding.from` / `encoding.to` through the same helper

And currently older QZ Tray versions only get `encoding = null` for exact string-form `encoding: 'legacy'`

src/qz/utils/SocketUtilities.java
```diff
@@ -34,7 +34,7 @@ public static void setupSocket(final Session session, String UID, SocketConnecti
             JSONObject options = params.getJSONObject("options");
 
             if (!options.isNull("encoding")) {
-                encoding = Charset.forName(options.getString("encoding"));
+                encoding = StringUtilities.getCharset(options.getString("encoding"));
```
IdelsTak, line 37: A small scope check here @tresf:

This part also enables `legacy` for socket encoding, not only RAW printing

That may be fine since sockets already expose an encoding option...but it means the feature is a bit broader than issue #1519

src/qz/utils/WindowsUtilities.java
```diff
@@ -756,4 +763,54 @@ public static boolean isAdminOwned(Path path) {
         }
         return false;
     }
+
+    static Charset getCharsetFromCodePage(int codePage) {
```
IdelsTak, line 767: One small coverage ask here @tresf:

This PR adds a compatibility path that is easy to regress...and I don't see focused tests around the charset helper yet

Could we add a tiny headless test for the mapping logic?

Something like:

```java
@Test
public void testLegacyCodePageMappings() {
    Assert.assertEquals(WindowsUtilities.getCharsetFromCodePage(1253).name(), "windows-1253");
    Assert.assertEquals(WindowsUtilities.getCharsetFromCodePage(65001), StandardCharsets.UTF_8);
    Assert.assertEquals(WindowsUtilities.getCharsetFromCodePage(-1), Charset.defaultCharset());
}
```
And maybe one small non-Windows helper test:

```java
@Test
public void testLegacyCharsetFallsBackOffWindows() {
    if (SystemUtilities.isWindows()) {
        throw new SkipException("Non-Windows fallback test");
    }

    Assert.assertEquals(StringUtilities.getCharset("legacy"), Charset.defaultCharset());
}
```

src/qz/printer/PrintOptions.java
```diff
@@ -57,14 +58,14 @@ public PrintOptions(JSONObject configOpts, PrintOutput output, PrintingUtilities
             if (encodings != null) {
                 // encoding may be a string or obj. Since optJSONObject didn't return null, it is an object
                 if (encodings.has("from")) {
-                    rawOptions.srcEncoding = Charset.forName(encodings.optString("from", Charset.defaultCharset().name()));
+                    rawOptions.srcEncoding = StringUtilities.getCharset(encodings.optString("from", Charset.defaultCharset().name()));
```
IdelsTak, line 61: Another possible follow-up test idea here too @tresf:

Since `PrintOptions` now accepts `legacy` through both string-form and object-form encoding, a small parser test would make the behavior clearer

For example:

```java
@Test
public void testLegacyEncodingStringForm() throws Exception {
    JSONObject options = new JSONObject().put("encoding", "legacy");
    PrintOptions printOptions = new PrintOptions(options, new PrintOutput((JSONObject)null), PrintingUtilities.Format.COMMAND);

    Assert.assertNotNull(printOptions.getRawOptions().getDestEncoding());
}

@Test
public void testLegacyEncodingObjectForm() throws Exception {
    JSONObject encoding = new JSONObject()
            .put("from", "legacy")
            .put("to", "legacy");
    JSONObject options = new JSONObject().put("encoding", encoding);
    PrintOptions printOptions = new PrintOptions(options, new PrintOutput((JSONObject)null), PrintingUtilities.Format.COMMAND);

    Assert.assertNotNull(printOptions.getRawOptions().getSrcEncoding());
    Assert.assertNotNull(printOptions.getRawOptions().getDestEncoding());
}
```

Not a blocker, per se...but it would cover the new parser surface directly

src/qz/utils/WindowsUtilities.java
tresf: Good catch, fixed.

js/qz-tray.js
tresf: Copied other techniques by changing to:
```js
if(config.encoding && config.encoding.toUpperCase() === 'LEGACY') {
    config.encoding = null;
}
```

src/qz/utils/SocketUtilities.java
tresf: Keen observation.  Yes, that code never used `Charset.defaultCharset();` however it would be inconsistent to allow a charset value in one place of the software but not others.  I've described this in the latest comments of #1519.   I've now added support for `legacy` to serial too.

src/qz/printer/PrintOptions.java
tresf: For windows-only, yes.  I would be happy with some unit-tests PR'd against this if you have time.

src/qz/utils/WindowsUtilities.java
tresf: ~~I completely agree.  If you don't mind, please create PR with coverage.~~

I'll write one, shoudl be quick.

tresf: added

src/qz/printer/PrintOptions.java
tresf: Nevermind, added.

test/qz/utils/WindowsUtilitiesTests.java
```diff
@@ -0,0 +1,45 @@
...
+    @Test
+    public void testGetCharsetFromCodePage() {
+        // Ensure proper fallback for bad charsets
+        Assert.assertEquals(getCharsetFromCodePage(-1), UTF_8); // -1 fallback to UTF8
+        Assert.assertEquals(getCharsetFromCodePage(Integer.MAX_VALUE), UTF_8); // missing fallback to UTF8
+        Assert.assertEquals(getCharsetFromCodePage(Integer.MAX_VALUE), UTF_8); // missing fallback to UTF8
+
+        // Ensure no fallback for good charsets
+        Assert.assertEquals(getCharsetFromCodePage(932), Charset.forName("windows-932")); // Windows Japanese
+        Assert.assertNotEquals(getCharsetFromCodePage(950), Charset.forName("windows-950")); // Windows Traditional Chinese
```
IdelsTak: @tresf, this looks like what is causing the Windows x86_64 CI failure

`getCharsetFromCodePage(950)` tries `windows-950` first, and java resolves that as `x-windows-950`

So this probably wants `assertEquals`, not `assertNotEquals`

I'd also consider asserting the selected charset directly for `1252` and `1253`, since that makes the search-order expectation clearer

test/qz/utils/WindowsUtilitiesTests.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.nio.charset.Charset;
+
+import static java.nio.charset.StandardCharsets.*;
+import static qz.utils.WindowsUtilities.*;
+
+public class WindowsUtilitiesTests {
+    private static final Logger log = LogManager.getLogger(WindowsUtilitiesTests.class);
+
+    @BeforeClass
```
IdelsTak, line 18: @tresf, maybe consider only the real tests that actually need Windows API...not all of them

See, `testGetSystemAnsiCodePage()` needs Windows...but `testGetCharsetFromCodePage()` is pure charset mapping logic...hence not Windows specific

And as a result, running the mapping test on Linux + macOS too would protect the search order and fallback behavior outside Windows CI

test/qz/utils/WindowsUtilitiesTests.java
tresf: Whoops, thanks, and I pasted it 3x.  Fixed.

test/qz/utils/WindowsUtilitiesTests.java
tresf: @IdelsTak normally I would agree, however this would require moving the helpers out of WindowsUtilities, yet these helpers are only ever needed for Windows.  In the future if we make a cross-platform approach at "guessing" this value on macOS and Linux, I will happily refactor them out and add coverage.

IdelsTak:
Good work here @tresf

PR #1522, opened 2026-09-17
Add focused svg icon tests

IdelsTak:
Adds a small headless test pass for the svg/`IconCache` work in #1515

I kept this intentionally narrow

Covered:
- representative `IconCache` resource loading
- logo/menu/dialog/tray/taskbar image sizes
- dark logo lookup
- normal dark fallback
- mask fallback
- taskbar multi-size images
- direct svg raster sizing through `ImageUtilities`
- png/base64 output used by web/about consumers

Left out on purpose:
- visual artifact checks
- live desktop tray rendering
- macOS template icon runtime behavior
- malformed svg fallback
- exhaustive every-icon inventory
- live theme-change refresh checks

Why:
- visual svg quality is better checked with runtime smoke tests
- tray rendering needs a real desktop tray host and varies by platform
- macOS template behavior cannot be proved well from the local arch/xfce machine
- malformed packaged svgs should fail hard based on @tresf current direction
- exhaustive icon inventory would baloon this tiny test pr unnecessarily
- live theme refresh involves ui/runtime state, not just cache and utility behavior


IdelsTak:
@tresf, review highly welcome

test/qz/utils/ImageUtilitiesTests.java
```diff
@@ -0,0 +1,38 @@
...
+    public void testSvgRasterSize() {
+        BufferedImage image = ImageUtilities.imageFromSvgResource(
+                Paths.get("../resources/logo.svg"),
+                260,
+                new IconCache()
+        );
+
+        Assert.assertNotNull(image, "SVG image should load");
+        Assert.assertEquals(image.getWidth(), 260, "SVG image width");
+        Assert.assertEquals(image.getHeight(), 260, "SVG image height");
```
tresf, line 24:
```suggestion
        // width will scale proportionally, height will remain the same
        Assert.assertEquals(image.getHeight(), 260, "SVG image height");
```

test/qz/ui/component/IconCacheTests.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static qz.ui.component.IconCache.Icon.*;
+
+public class IconCacheTests {
+    private IconCache iconCache;
+
+    @BeforeMethod
+    public void setUp() {
+        iconCache = new IconCache();
+    }
```
tresf, line 20:
```suggestion
    private IconCache iconCache = IconCache.getInstance();
```

test/qz/utils/ImageUtilitiesTests.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static qz.ui.component.IconCache.Icon.LOGO_ICON;
+
+public class ImageUtilitiesTests {
+    @Test
+    public void testSvgRasterSize() {
+        BufferedImage image = ImageUtilities.imageFromSvgResource(
+                Paths.get("../resources/logo.svg"),
+                260,
+                new IconCache()
```
tresf, line 19:
```suggestion
                IconCache.getInstance();
```

tresf:
Thanks. This is OK, but it uses hard-coded sizes which require changing our unit tests every time we make a cosmetic change.  This may make sense for well-known sizes but in general I feel complicates the tests.   Instead, I would recommend using a dataprovider to loop over all sizes.  Furthemore, the tests don't provide any PNG fallback tests, which are important to our company branded portal.

Perhaps put simpler, we'll notice RIGHT AWAY if we break our own artwork, so coverage there is not strictly necessary (although welcome).  Meanwhile, the hidden features of this PR -- such as logo.png fallback -- have NO COVERAGE at all.  Reminder, the unit tests can have their own resource directory if needed.  We do this for other unit tests.

tresf:
Fixing unit tests to use mixed-resource types (PNG + SVG) required quite a bit of refactoring.  I've pushed it to the other branch.

tresf:
Superceded by https://github.com/qzind/tray/pull/1515.

